PR #2, opened 2023-11-09
HYDRA-500 : Fix directional and default lights not working.

lanierd-adsk:


lanierd-adsk:
I will try to add a new unit test for this problem on tomorrow

lib/mayaHydra/hydraExtensions/adapters/directionalLightAdapter.cpp
```diff
@@ -53,11 +53,14 @@ class MayaHydraDirectionalLightAdapter : public MayaHydraLightAdapter
 
     void _CalculateLightParams(GlfSimpleLight& light) override
     {
+        //To simulate a directional light which has no actual position, but doesn't seem to be supported in hydra, we set a position very very far
+        //so it looks like a directional light.
         // Directional lights point toward -Z, but we need the opposite
         // for the position so the light acts as a directional light.
-        const auto direction = GfVec4f(0.0, 0.0, 1.0, 0.0) * GetTransform();
+        static const double farfarAway {1.0e15};//we use a point on the Z axis far far away
```
lilike-adsk: looks better to define a shared variable for it, as it's used elsewhere?

lanierd-adsk: Right, but both are not in the same project so I thought it would be simpler to do so.

lanierd-adsk: As discussed offline, I'll try to create a shared function that deal with that light direction computation.

lanierd-adsk:
@ppt-adsk @roopavr-adsk @debloip-adsk

lib/mayaHydra/hydraExtensions/adapters/directionalLightAdapter.cpp
lanierd-adsk: I created a shared function and added some tests.

lilike-adsk (dismissed):
looks great

lilike-adsk (approved):

PR #4, opened 2023-11-09
HYDRA-598 : Merge viewport information interface

lanierd-adsk:


lanierd-adsk:
FYI : @lilike-adsk @roopavr-adsk @ppt-adsk

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.cpp
```diff
@@ -0,0 +1,100 @@
...
+}
+    
+PXR_NAMESPACE_USING_DIRECTIVE
+
+namespace FVP_NS_DEF {
+
+static InformationInterfaceImp theInterface;
+
+//Set of information clients
+static InformationClientSet _viewportInformationClients;
```
debloip-adsk: Observation : I see this has been discussed before, but is using a set truly advantageous over a vector? From my understanding the callbacks are going to be invoked during user interaction with the DCC tools, and so will be executed much more often than client registration/unregistration, and in scenarios where performance is more relevant. So while using a set does makes registration O(1), it also changes the time complexity of callbacks iteration from being linear in the number of clients to being linear in the size of the set, which adds some amount of overhead. That being said, without concrete use cases that we can profile, it might be a tad too early to make a call on this?

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.h
```diff
@@ -0,0 +1,48 @@
...
+#include <flowViewport/API/fvpInformationInterface.h>
+#include <flowViewport/API/fvpInformationClient.h>
+
+namespace FVP_NS_DEF {
+
+///Is a singleton, use InformationInterfaceImp& _getInformationInterfaceImp(void) below to get an instance of that interface
+class InformationInterfaceImp :  public InformationInterface
+{
+public:
+    InformationInterfaceImp()             = default;
```
debloip-adsk: Broader comment about our codebase : we're starting to have multiple singletons, but many of them are implemented in different ways, so I think we should standardize how we implement them, with regards to : 
- Whether we make the "main" constructor private or not
- Whether we delete the copy constructor and assignment operator
- How we manage the instance object (e.g. anonymous namespace, as a static variable in the static accessor method, etc.)
- The naming of the accessor method

This would need discussion as a team so it's not for now, but I'm commenting here while it's on my mind

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.h
```diff
@@ -0,0 +1,48 @@
...
+#define FLOW_VIEWPORT_API_INTERFACESIMP_INFORMATION_INTERFACE_IMP_H
+
+//Local headers
+#include <flowViewport/api.h>
+#include <flowViewport/API/fvpInformationInterface.h>
+#include <flowViewport/API/fvpInformationClient.h>
+
+namespace FVP_NS_DEF {
+
+///Is a singleton, use InformationInterfaceImp& _getInformationInterfaceImp(void) below to get an instance of that interface
```
debloip-adsk: This comment seems outdated, looks like `InformationInterfaceImp& _getInformationInterfaceImp(void)` should be `InformationInterfaceImp::Get()`

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.cpp
```diff
@@ -0,0 +1,100 @@
...
+//
+
+//Local headers
+#include "fvpInformationInterfaceImp.h"
+#include "flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportDataManager.h"
+
+#include <mutex>
+
+namespace{
+    static std::mutex _viewportInformationClient_mutex;
```
debloip-adsk: Nit : doesn't need to be static since it is already in an anonymous namespace

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+//Set of information clients
+static InformationClientSet _viewportInformationClients;
+
+InformationInterface& InformationInterface::Get() 
+{ 
+    return theInterface;
+}
+
+InformationInterfaceImp& InformationInterfaceImp::Get(void)
```
debloip-adsk: Nit : I think void is unnecessary here

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.cpp
```diff
@@ -0,0 +1,100 @@
...
+{
+    return theInterface;
+}
+
+void InformationInterfaceImp::RegisterInformationClient(const InformationClient& client)
+{
+    std::lock_guard<std::mutex> lock(_viewportInformationClient_mutex);
+
+    InformationClient* clientNonConst = const_cast<InformationClient*>(&client);
```
debloip-adsk: If we are to const_cast the argument, should we not just directly take a non-const parameter?

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.cpp
```diff
@@ -0,0 +1,100 @@
...
+    if (foundResult == _viewportInformationClients.cend()){
+        _viewportInformationClients.insert(clientNonConst);
+    }
+}
+
+void InformationInterfaceImp::UnregisterInformationClient(const InformationClient& client)
+{
+    std::lock_guard<std::mutex> lock(_viewportInformationClient_mutex);
+
+    InformationClient* clientNonConst = const_cast<InformationClient*>(&client);
```
debloip-adsk: See my equivalent comment in RegisterInformationClient

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.cpp
debloip-adsk: Nit : I think these should be moved to the anonymous namespace

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportDataManager.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+//Hydra headers
+#include <pxr/imaging/hd/renderIndex.h>
+
+//STL Headers
+#include <mutex>
+
+namespace 
+{
+    static std::mutex _viewportInformationAndSceneIndicesPerViewportDataSet_mutex;
```
debloip-adsk: Nit : doesn't need to be static since it is already in an anonymous namespace

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportData.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+ViewportInformationAndSceneIndicesPerViewportData::ViewportInformationAndSceneIndicesPerViewportData(const CreationParameters& creationParams)
+    : _viewportInformation(creationParams._viewportInformation), _renderIndexProxy(creationParams._renderIndexProxy)
+{
+}
+
+ViewportInformationAndSceneIndicesPerViewportData::~ViewportInformationAndSceneIndicesPerViewportData()
+{
+    //The viewport info is own by this class
+    InformationInterfaceImp::Get().SceneIndexRemoved(_viewportInformation);
```
debloip-adsk: Nit : can we use the customer-facing interface here? It doesn't look like we need the actual implementation

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -748,14 +749,37 @@ void MtohRenderOverride::_InitHydraResources()
...
+    std::string cameraName;
+    int viewportWidth = 0;
+    int viewportHeight = 0;
+
+    //Get information from viewport to be used later
+    MString panelName;
+	if (drawContext.renderingDestination(panelName) == MFrameContext::k3dViewport) { 
+        M3dView view;
+	    if (M3dView::getM3dViewFromModelPanel(panelName, view)){
+            M3dView *viewPtr = const_cast<M3dView*>( &view );
```
debloip-adsk: Minor : this const_cast is unnecessary

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -387,18 +388,32 @@ MayaHydraSceneIndex::MayaHydraSceneIndex(
 }
 
 MayaHydraSceneIndex::~MayaHydraSceneIndex()
+{
+    //If you get a crash in a callback with a nullptr for _sceneIndex, 
+    // it may be due to the fact that the _sceneIndex pointer has been nulled as its ref count reached 0 but the destructor is still being called.
+    //You should call RemoveCallbacksAndDeleteAdapters(); before the destructor is called.
```
debloip-adsk, line 400: Not sure of what is being described here : if the _sceneIndex pointer's refcount reaches 0, isn't it expected/desired that the destructor be called? If this causes issues, wouldn't this point to a lifetime management issue?

lib/mayaHydra/hydraExtensions/mayaHydraLibInterfaceImp.cpp
```diff
@@ -46,4 +53,13 @@ const SceneIndicesVector& MayaHydraLibInterfaceImp::GetTerminalSceneIndices() co
     return _sceneIndices;
 }
 
+void MayaHydraLibInterfaceImp::SceneIndexRemoved(const HdSceneIndexBaseRefPtr& _sceneIndex)
```
debloip-adsk: 2 observations here : 
a) This seems to do the same work as UnregisterTerminalSceneIndex, so I feel like we should either have UnregisterTerminalSceneIndex call SceneIndexRemoved or vice-versa
b) Where/when is this called? It looks like the same type of callback as for InformationClients, but MayaHydraLibInterface isn't an InformationClient, and the method also doesn't seem to be called directly

lib/flowViewport/API/fvpInformationInterface.h
```diff
@@ -0,0 +1,93 @@
...
+
+        ///Struct used to store information about an Hydra viewport from the DCC
+        struct ViewportInformation
+        {
+            /// Constructor
+            ViewportInformation(const PXR_NS::HdSceneIndexBaseRefPtr& viewportSceneIndex, const std::string& cameraName, const size_t viewportWidth, const size_t viewportHeight, const std::string& rendererName)
+                : _viewportSceneIndex(viewportSceneIndex), _cameraName(cameraName), _viewportWidth(viewportWidth), _viewportHeight(viewportHeight), _rendererName(rendererName) {}
+            
+            ///_viewportSceneIndex is an Hydra viewport scene index
+            PXR_NS::HdSceneIndexBaseRefPtr _viewportSceneIndex;
```
debloip-adsk: Minor : can this pointer be const, since every other field of this struct is const?

lib/mayaHydra/hydraExtensions/mayaHydraLibInterface.h
```diff
@@ -17,15 +17,16 @@
...
 
+//Hydra headers
 #include <pxr/imaging/hd/sceneIndex.h>
 
-#include <vector>
 
 PXR_NAMESPACE_OPEN_SCOPE
 
-using SceneIndicesVector = std::vector<HdSceneIndexBasePtr>;
+using SceneIndicesVector                                        = std::vector<HdSceneIndexBasePtr>;//Be careful, these are not not Ref counted. Elements could become dangling
```
debloip-adsk, line 29: If this is causing issues, is there a downside to converting these to refcounted pointers? I made these non-owning originally simply because they technically don't own the scene indices, but I don't think it matters too much.

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportDataManager.h
```diff
@@ -0,0 +1,71 @@
...
+
+    ViewportInformationAndSceneIndicesPerViewportDataSet&        GetViewportInformationAndSceneIndicesPerViewportDataSet() {return _viewportInformationAndSceneIndicesPerViewportDataSet;}
+    const ViewportInformationAndSceneIndicesPerViewportDataSet&  GetViewportInformationAndSceneIndicesPerViewportDataSet() const {return _viewportInformationAndSceneIndicesPerViewportDataSet;}
+
+    ViewportInformationAndSceneIndicesPerViewportData*          GetViewportInformationAndSceneIndicesPerViewportDataFromViewportSceneIndex(const PXR_NS::HdSceneIndexBaseRefPtr& viewportSceneIndex);
+    const ViewportInformationAndSceneIndicesPerViewportData*    GetViewportInformationAndSceneIndicesPerViewportDataFromViewportSceneIndex(const PXR_NS::HdSceneIndexBaseRefPtr& viewportSceneIndex) const;
+
+    ViewportInformationAndSceneIndicesPerViewportData*          GetViewportInformationAndSceneIndicesPerViewportDataFromRenderIndexProxy(const RenderIndexProxy& renderIndexProxy);
+    const ViewportInformationAndSceneIndicesPerViewportData*    GetViewportInformationAndSceneIndicesPerViewportDataFromRenderIndexProxy(const RenderIndexProxy& renderIndexProxy) const;
```
debloip-adsk: Do we need the non-const versions of these methods? Could avoid some code duplication

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportData.h
```diff
@@ -0,0 +1,76 @@
...
+
+    PXR_NS::HdSceneIndexBaseRefPtr& GetLastFilteringSceneIndexOfTheChain() {return _lastFilteringSceneIndexOfTheChain ;}
+    const PXR_NS::HdSceneIndexBaseRefPtr& GetLastFilteringSceneIndexOfTheChain() const {return _lastFilteringSceneIndexOfTheChain;}
+
+    RenderIndexProxy& GetRenderIndexProxy() {return _renderIndexProxy;}
+    const RenderIndexProxy& GetRenderIndexProxy() const {return _renderIndexProxy;}
+
+    void                                    SetInputSceneIndex(const PXR_NS::HdSceneIndexBaseRefPtr& inputSceneIndex) {_inputSceneIndex = inputSceneIndex;}
+    PXR_NS::HdSceneIndexBaseRefPtr&         GetInputSceneIndex() {return _inputSceneIndex;}
+    const PXR_NS::HdSceneIndexBaseRefPtr&   GetInputSceneIndex() const {return _inputSceneIndex;}
```
debloip-adsk: Do we need the non-const versions of the GetXYZ methods? Could avoid some code duplication

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportData.cpp
```diff
@@ -0,0 +1,38 @@
...
+    delete &_viewportInformation;
```
debloip-adsk: (For conversation on this comment, see [this discussion](https://github.com/Autodesk/maya-hydra/pull/4#discussion_r1391790411))
This delete seems dangerous, since there is no guarantee that the ViewportInformation we receive has been created on the heap, nor that whoever creates it doesn't also delete it.

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportData.h
```diff
@@ -0,0 +1,76 @@
...
+#include "flowViewport/API/fvpInformationInterface.h"
+#include "flowViewport/sceneIndex/fvpRenderIndexProxy.h"
+
+namespace FVP_NS_DEF {
+
+/** Stores information and misc. scene indices data per viewport
+*   So, if there are "n" Hydra viewports in the DCC, we will have "n" instances of this class.
+*/
+
+class ViewportInformationAndSceneIndicesPerViewportData
```
debloip-adsk, line 31: Naming suggestion : this class name is a bit long and heavy to read, is it possible to shorten it? Maybe something like ViewportData? I realize this could be confusing when compared with ViewportInformation, but in that case maybe we could rename ViewportInformation to something like ViewportDescriptor? I don't really have a great solution in mind to be honest, just thinking out loud

test/lib/mayaUsd/render/mayaToHydra/cpp/testFlowViewportAPIViewportInformation.cpp
```diff
@@ -0,0 +1,140 @@
...
+    int _numSceneIndexAdded     = 0; 
+    int _numSceneIndexRemoved   = 0;
+};
+
+
+//Is global instance
+static InfoClientTest _infoClientTest;
+
+//Storm renderer name
+static const std::string _stormRendererName ("GL");
```
debloip-adsk: Nit : these could be moved to an anonymous namespace

plugin/mayaHydraSceneBrowserTest/mayaHydraSceneBrowserTestCmd.cpp
```diff
@@ -16,7 +16,7 @@
 #include "mayaHydraSceneBrowserTestCmd.h"
 
 #include <mayaHydraLib/hydraUtils.h>
-#include <mayaHydraLib/interfaceImp.h>
+#include <mayaHydraLib/mayaHydraLibInterfaceImp.h>
```
debloip-adsk: Nit : could use the non-imp version (introduced by me, sorry)

plugin/mayaHydraSceneBrowser/mayaHydraSceneBrowser.cpp
```diff
@@ -15,7 +15,7 @@
 #include "mayaHydraSceneBrowser.h"
 
 #include <mayaHydraLib/hydraUtils.h>
-#include <mayaHydraLib/interfaceImp.h>
+#include <mayaHydraLib/mayaHydraLibInterfaceImp.h>
```
debloip-adsk: Nit : could use the non-imp version (introduced by me, sorry)

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -773,22 +797,27 @@ void MtohRenderOverride::_InitHydraResources()
 
     _mayaHydraSceneProducer->Populate();
 
-    _selectionSceneIndex = Fvp::SelectionSceneIndex::New(_renderIndexProxy->GetMergingSceneIndex());
+    HdSceneIndexBaseRefPtr lastSceneIndexOfTheChain = _renderIndexProxy->GetMergingSceneIndex();
+
+    //Get the latest scene index of the custom filtering scene indices chain and apply the selection scene index
```
debloip-adsk: Question : this will be for another PR, correct? Since this PR doesn't have the filtering setup yet

lib/flowViewport/sceneIndex/fvpRenderIndexProxy.cpp
```diff
@@ -121,9 +125,26 @@ void RenderIndexProxy::RemoveSceneIndex(
...
 {
     return _mergingSceneIndex;
 }
 
+HdRenderIndex* RenderIndexProxy::GetRenderIndex() const 
+{ 
+    return _renderIndex;
 }
+
+void RenderIndexProxy::GetRendererDisplayName(std::string& outRendererName) const
```
debloip-adsk: Minor : the Pixar RenderIndex's GetRendererDisplayName returns a string rather than taking an out parameter, I feel like we should we follow the same signature for this method?

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportDataManager.cpp
```diff
@@ -0,0 +1,180 @@
...
+            if (renderIndexProxy){
+                //Following code is equivalent to calling FilteringSceneIndicesChainManager::GetManager().DestroyFilteringSceneIndicesChain(*renderIndexProxy);
+                //But we cannot do so because of the lock above.
+                auto renderIndex = renderIndexProxy->GetRenderIndex();
+                if (renderIndex && (*findResult)->GetLastFilteringSceneIndexOfTheChain()){
+                    renderIndex->RemoveSceneIndex((*findResult)->GetLastFilteringSceneIndexOfTheChain());//Remove the whole chain from the render index
+                }
+            }
+            
+            delete (*findResult); //delete the ViewportInformationAndSceneIndicesPerViewportData*
```
debloip-adsk: I'm worried about manually managing the ViewportInformationAndSceneIndicesPerViewportData : could we store the ViewportInformationAndSceneIndicesPerViewportData by value, or at least with smart pointers to reduce the risk of running into memory management issues?

debloip-adsk (changes requested):
Hi, thanks for doing this! I know I've put a lot of comments, but a lot of them are nits, minor things or questions to help me understand things a little better.

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.cpp
lanierd-adsk: @ppt-adsk : what do you think ? since you are the person who asked me to convert the std::vector to std::set in a previous code review.

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.h
lanierd-adsk: You're right.

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.h
lanierd-adsk: Thanks I fixed this and the constructor has been moved to private members of the class.

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.cpp
lanierd-adsk: It's an old habit I did that so you cannot send a parameter to that function or it produces an error, but I'll remove it.

plugin/mayaHydraSceneBrowserTest/mayaHydraSceneBrowserTestCmd.cpp
lanierd-adsk: Right, done.

plugin/mayaHydraSceneBrowser/mayaHydraSceneBrowser.cpp
lanierd-adsk: Right, done.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lanierd-adsk: That's right, I prepared the code for an upcoming PR, but I think we should have a proper function for this ordering, I am going to add it..

lib/flowViewport/sceneIndex/fvpRenderIndexProxy.cpp
lanierd-adsk: I usually prefer to pass the parameter to be filled as an out parameter instead of relying on the compiler to optimize the return value and avoid copies on the stack. But that's probably because I'm old school :-) 
But I guess for std::string they use std::move ? So I'd do it like you said.

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportDataManager.cpp
lanierd-adsk: I started to store them by value as you suggested but there is still the ViewportInformation pointer to manage (new/delete) which is inside this class so we would still have to manage some memory. 
So is this something you still want ?

lanierd-adsk: Also since this is data per viewport and is not shared by any other class I think it should be manageable.

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.cpp
ppt-adsk: Not sure I follow the argument.  Isn't the size of the size exactly equal to the number of clients?  If we're not indexing into the vector, simply iterating over all elements, then vector and set are equivalent, no?

debloip-adsk: Apologies, you're completely right, I think I got mixed up with some set implementations in other languages using hashsets. Ignore this.

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportDataManager.cpp
debloip-adsk: I haven't managed to make it work with with unique_ptrs (there is a copy I can't find somewhere, but the error message only indicates lines in the standard library headers), but it does work with shared_ptrs. I'm not sure if the overhead would be a problem or not, but either ways, I think as long as we stick to the one `new` and the one `delete` we currently have, this should be fine as it's isolated in the implementation and very symmetrical.

debloip-adsk: (Grouping this conversation with [this comment](https://github.com/Autodesk/maya-hydra/pull/4#discussion_r1391763392))

So after taking a closer look, the situation with `ViewportInformation` objects is quite similar, as like you said, those are also manually managed, with the exception that objects of that class are expected to be manually created outside of the Flow Viewport Toolkit, and their ownership then transferred to the Flow Viewport Toolkit. I feel like this should also be solvable with unique_ptrs.

That being said, if we can't get smart pointers to work, then I think we should at least explicitly document that the `AddViewportInformation` method will assume ownership of the `ViewportInformation` object it receives, so that users know not to have the object they pass in be deleted. And to make the method signature a bit more explicit, maybe we can also make it receive a raw pointer rather than a const reference?

lib/mayaHydra/hydraExtensions/mayaHydraSceneProducer.cpp
```diff
@@ -50,6 +53,11 @@ MayaHydraSceneProducer::MayaHydraSceneProducer(
         initData.producer = this;
         _sceneIndex = MayaHydraSceneIndex::New(initData, lightEnabled);
         TF_VERIFY(_sceneIndex, "Maya Hydra scene index not found, check mayaHydra plugin installation.");
+        
+        //Create an HydraViewportInformation instance
+        FVP_NS_DEF::InformationInterface::ViewportInformation* hydraViewportInformation = 
+            new FVP_NS_DEF::InformationInterface::ViewportInformation(_sceneIndex, initData.cameraName, initData.viewportWidth, initData.viewportHeight, initData.rendererName);
+        FVP_NS_DEF::ViewportInformationAndSceneIndicesPerViewportDataManager::Get().AddViewportInformation(*hydraViewportInformation, _renderIndexProxy);
```
debloip-adsk: Minor : I think these should be FVP_NS since we're in MayaHydra and thus a consumer of the Flow Viewport Toolkit? There are also a couple other instances (in mayaHydraSceneProducer.h/.cpp, and in testFlowViewportAPIViewportInformation.cpp)

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.cpp
```diff
@@ -0,0 +1,99 @@
...
+        if (viewportInfoClient){
+            viewportInfoClient->SceneIndexRemoved(_viewportInfo);
+        }
+    }
+}
+
+void InformationInterfaceImp::GetViewportsInformation(std::set<const ViewportInformation*>& outHydraviewportInformationSet)const
+{
+    outHydraviewportInformationSet.clear();
+    const ViewportInformationAndSceneIndicesPerViewportDataSet& viewportInformationAndSceneIndicesPerViewportDataVector =
```
debloip-adsk: This variable still has vector in its name but is now a set

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.cpp
lanierd-adsk: I prefer to keep a const reference as a parameter so we know the function doesn't change it. Only the storage in the std::set stores a non const, and I cannot change the std::set to "const ViewportInformation*" which I tried as I don't want SceneIndexAdded or SceneIndexRemoved to be const functions as a 3rd party using it as a callback could modify things in this function.

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportData.cpp
lanierd-adsk: I would prefer to use the internal representation, so InformationInterfaceImp instead of InformationInterface since this is internal code. 
For me the internal code should use the internal implementation and external code uses the external implementation.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lanierd-adsk: Right, I removed that variable which was useless.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
lanierd-adsk: This is a comment for potential future debugging. I am not sure if this is still valid since I made changes since then, but I would prefer to leave the comment to save time later if we get a crash in callbacks while this class has been deleted.

lib/mayaHydra/hydraExtensions/mayaHydraLibInterfaceImp.cpp
lanierd-adsk: You're right it's a duplicate of UnregisterTerminalSceneIndex, I am removing it.

lib/mayaHydra/hydraExtensions/mayaHydraLibInterface.h
lanierd-adsk: I think it's just a comment to be aware than when using SceneIndicesVector it's not ref counted but I don't think it can cause problems. But in one of the past version of this code, I was using some vector with custom class than were ref counted so we could have an ambiguity. I would left the comment as it's not always obvious from the code to differentiate between HdSceneIndexBasePtr and HdSceneIndexBaseRefPtr

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportDataManager.h
lanierd-adsk: I usually define fuinction as const and non const but you're right that we could define them when we need them, I am removing the non used versions.

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportData.h
lanierd-adsk: Right, I removed them.

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.cpp
lanierd-adsk: I replied above. Thanks.

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportData.cpp
lanierd-adsk: It is created by MayaHydraSceneProducer::MayaHydraSceneProducer in :
FVP_NS_DEF::InformationInterface::ViewportInformation* hydraViewportInformation = 
    new FVP_NS_DEF::InformationInterface::ViewportInformation(_sceneIndex, initData.cameraName, initData.viewportWidth, initData.viewportHeight, initData.rendererName);
 
So it's on the heap and nothing else deletes it.

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportData.h
lanierd-adsk: No, I prefer to keep the names like that, I prefer long heavy names that are explicit about what this class contains.

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportDataManager.cpp
lanierd-adsk: AddViewportInformation is not called by users, but by maya hydra only. But you're right, I changed the method signature and added more comments about the ownership.

lib/mayaHydra/hydraExtensions/mayaHydraSceneProducer.cpp
lanierd-adsk: Right, I am updating them.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -863,6 +867,36 @@ void MtohRenderOverride::ClearHydraResources()
...
+
+    HdSceneIndexBaseRefPtr lastSceneIndexOfTheChain = _renderIndexProxy->GetMergingSceneIndex();
+
+    _selection = std::make_shared<Fvp::Selection>();
+    _selectionSceneIndex = Fvp::SelectionSceneIndex::New(lastSceneIndexOfTheChain, _selection);
+    _selectionSceneIndex->SetDisplayName("Flow Viewport Selection Scene Index");
+    lastSceneIndexOfTheChain = _selectionSceneIndex;
+  
+    if (!_sceneIndexRegistry) {
+        _sceneIndexRegistry.reset(new MayaHydraSceneIndexRegistry(*_renderIndexProxy));
```
lanierd-adsk: @ppt-adsk : is that expected that _sceneIndexRegistry is not added to the render index proxy or isn't used as an entry of another scene index, it doesn't seem to be part of the chain ?

lanierd-adsk:
The preflight is passing on all platforms now.

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportDataManager.cpp
ppt-adsk: Explicit calls to delete are a maintenance burden and are not exception safe.  They violate this C++ core guideline:
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-newdelete
If we're still using an explicit call to delete (and it looks like we are), we should try to change the code to avoid it.

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportData.cpp
debloip-adsk: Ah yeah that makes sense indeed, and especially if we ever need to use an implemention-specific thing we would need to switch back to the implementation anyways.

lib/mayaHydra/hydraExtensions/mayaHydraSceneProducer.cpp
ppt-adsk: Just use "Fvp".  Since we control the compilation of the Flow Viewport library, using the FVP_NS macro to account for changing the Flow Viewport library namespace name is an over elaboration, since we won't do this.

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.cpp
ppt-adsk: I think this is where we should be careful with long identifier names: they should not reveal implementation details, which is violating encapsulation and Don't Repeat Yourself.

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.cpp
ppt-adsk: This feels very awkward in terms of ownership.  We shouldn't be storing a pointer to something that is conceptually passed by value, because the argument object may disappear if it goes out of scope.  I think this should be changed.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
ppt-adsk: I'm with @debloip-adsk on this, it seems that there is a lifescope management issue in there.

lib/mayaHydra/hydraExtensions/mayaHydraLibInterface.h
ppt-adsk: If it's a maintenance burden, then I would change to using HdSceneIndexBaseRefPtr.

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportDataManager.h
ppt-adsk: This is a side conversation, but the naming is very verbose, especially when coupled with accessing the singleton.  Would it really be:
`ViewportInformationAndSceneIndicesPerViewportDataManager::Get().GetViewportInformationAndSceneIndicesPerViewportDataFromRenderIndexProxy(aRenderIndexProxy);`
That's a bit much...

ppt-adsk: Also, why do we need the `ViewportInformationAndSceneIndicesPerViewportDataManager` to get data from the RenderIndexProxy?  Why can't the RenderIndexProxy provide this information itself?

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportData.cpp
ppt-adsk: Again, an explicit delete violates
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-newdelete

lib/mayaHydra/mayaPlugin/renderOverride.cpp
ppt-adsk: @lanierd-adsk why should the scene index registry be part of the chain?  It is not a scene index.

ppt-adsk (changes requested):

lib/mayaHydra/hydraExtensions/mayaHydraSceneProducer.cpp
lanierd-adsk: ok I am updating it.

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportDataManager.cpp
lanierd-adsk: ok, I will try to find a way to avoid delete.

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.cpp
lanierd-adsk: ok, I am removing the details implementation.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lanierd-adsk: ok, I thought it was, sorry about that. Fine then.

lanierd-adsk:
There should have only 1 remaining problem about the lifetime management which we are discussing offline.

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportData.h
debloip-adsk: I agree on having explicit names, I just usually like to keep names short, as long names can sometimes indicate that something has too much responsibility. But in this case I can't really find a name that conveys the meaning behind this, so I'm fine with leaving it like this

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.cpp
debloip-adsk: I think using Array in the name can still be confusing since arrays are also a specific container. What I tend to go with is using plurals, but since this variable/class name is already quite long, a lone "s" will be hard to spot and differentiate. Maybe we could name it something like allViewportInformationAndSceneIndicesPerViewportData? Ultimately I don't have that strong of an opinion on this.

lib/flowViewport/API/fvpInformationInterface.h
```diff
@@ -80,9 +89,9 @@ namespace FVP_NS_DEF
         /**
         *  @brief      Get the Hydra viewports information. 
         *
-        *  @param[out] outHydraViewportInformationSet is a set of ViewportInformation to have information about each Hydra viewport in use in the current DCC.
+        *  @param[out] outHydraViewportInformationArray is a set of ViewportInformation to have information about each Hydra viewport in use in the current DCC.
         */
-        virtual void GetViewportsInformation(std::set<const ViewportInformation*>& outHydraViewportInformationSet)const  = 0;
+        virtual void GetViewportsInformation(std::set<const ViewportInformation*>& outHydraViewportInformationArray)const  = 0;
```
ppt-adsk: Still uncomfortable as to why the set holds pointers to objects.

lib/flowViewport/API/fvpInformationInterface.h
```diff
@@ -60,6 +60,15 @@ namespace FVP_NS_DEF
 
             ///_rendererName is the Hydra viewport renderer name (example : "GL" for Storm or "Arnold" for the Arnold render delegate)
             const std::string _rendererName;
+
+            ///Comparison operator
+            bool operator ==(const ViewportInformation& other)const{
```
ppt-adsk, line 58: Where is this used?

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportData.h
```diff
@@ -41,18 +41,22 @@ class ViewportInformationAndSceneIndicesPerViewportData
...
+    ~ViewportInformationAndSceneIndicesPerViewportData() = default;
     
     const InformationInterface::ViewportInformation& GetViewportInformation()const { return _viewportInformation;}
     const PXR_NS::HdSceneIndexBaseRefPtr& GetLastFilteringSceneIndexOfTheChain() const {return _lastFilteringSceneIndexOfTheChain;}
     const RenderIndexProxy& GetRenderIndexProxy() const {return _renderIndexProxy;}
     void SetInputSceneIndex(const PXR_NS::HdSceneIndexBaseRefPtr& inputSceneIndex) {_inputSceneIndex = inputSceneIndex;}
     const PXR_NS::HdSceneIndexBaseRefPtr&   GetInputSceneIndex() const {return _inputSceneIndex;}
     
+    //Needed by std::set
+    bool operator < (const ViewportInformationAndSceneIndicesPerViewportData& other)const{
```
ppt-adsk, line 45: This seems wrong.  This would mean that a < b and b < a would both be true, which is logically inconsistent.  I'm not sure how a std::set would work with this.

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.cpp
lanierd-adsk: Right, I changed it  in allViewportInformationAndSceneIndicesPerViewportData.

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.h
lanierd-adsk: I did the changes you mentionned on that part and moved the constructor to private.

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportDataManager.cpp
lanierd-adsk: I am using a std::set with copies in it instead of arrays, the class to copy is not heavy and this set will have the size of the number of viewports in maya.

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.cpp
lanierd-adsk: I changed it to storing a copy of the class since the class is not heavy and the size of the std set is the number of hydra viewports in maya.

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportData.cpp
lanierd-adsk: I removed this delete by storing copies instead of pointers as the class is not heavy and the size of the std set will be the number of hydra viewports in maya

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportDataManager.h
lanierd-adsk: You both are right and this function is not used any longer, I removed it.

lib/mayaHydra/hydraExtensions/mayaHydraSceneProducer.cpp
```diff
@@ -54,11 +54,9 @@ MayaHydraSceneProducer::MayaHydraSceneProducer(
         _sceneIndex = MayaHydraSceneIndex::New(initData, lightEnabled);
         TF_VERIFY(_sceneIndex, "Maya Hydra scene index not found, check mayaHydra plugin installation.");
         
-        //Create an HydraViewportInformation instance
-        FVP_NS::InformationInterface::ViewportInformation* hydraViewportInformation = 
-            new FVP_NS::InformationInterface::ViewportInformation(_sceneIndex, initData.cameraName, initData.viewportWidth, initData.viewportHeight, initData.rendererName);
-        //This method takes the ownership of the hydraViewportInformation passed to it and will delete it
-        FVP_NS::ViewportInformationAndSceneIndicesPerViewportDataManager::Get().AddViewportInformation(hydraViewportInformation, _renderIndexProxy);
+        //Create an HydraViewportInformation 
+        const Fvp::InformationInterface::ViewportInformation hydraViewportInformation(_sceneIndex, initData.cameraName, initData.viewportWidth, initData.viewportHeight, initData.rendererName);
```
ppt-adsk: Unfortunate that we need to expand out MayaHydraDelegate::InitData to convert between two similar data structures.  Might want to eventually simply use Fvp::InformationInterface::ViewportInformation instead of MayaHydraDelegate::InitData, to reduce code complexity.

lib/flowViewport/API/interfacesImp/fvpInformationInterfaceImp.cpp
debloip-adsk: Looks like we're still storing pointers now though, no? Am I missing something?

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
debloip-adsk: Now logged as a separate future task, resolving this conversation.

lib/flowViewport/API/fvpInformationInterface.h
```diff
@@ -43,59 +43,53 @@ namespace FVP_NS_DEF
...
-                return  _viewportSceneIndex == other._viewportSceneIndex &&
+                return  _viewportId == other._viewportId &&
                         _cameraName == other._cameraName &&
-                        _viewportWidth == other._viewportWidth &&
-                        _viewportHeight == other._viewportHeight&&
                         _rendererName == other._rendererName;
             }
+
+            bool operator <(const ViewportInformation& other) const{ //to be used in std::set
+                return  this < &other;
```
ppt-adsk: This would mean that two equal ViewportInformation objects would be reported as non-equal, since set uses !(a < b) and !(b < a) to imply that a == b.
https://cplusplus.com/reference/set/set/
We should implement this correctly as something like:
```suggestion
auto a = {_viewportId, _cameraName, _rendererName};
auto b = {other._viewportId, other.cameraName, other._rendererName};
return std::lexicographical_compare(a.begin(), a.end(), b.begin(), b.end());
```

lib/flowViewport/API/fvpInformationInterface.h
```diff
@@ -43,59 +43,53 @@ namespace FVP_NS_DEF
...
+        ///Set of InformationInterface::ViewportInformation
+        typedef std::set<InformationInterface::ViewportInformation> ViewportInformationSet;
+
         /**
         *  @brief      Register a set of callbacks through an InformationClient instance
         *
         *  @param[in]  client is the InformationClient.
         */
-        virtual void RegisterInformationClient(const InformationClient& client) = 0;
+        virtual void RegisterInformationClient(InformationClient* client) = 0;
```
ppt-adsk: Since we're passing a plain pointer, what is the ownership model of InformationClient?  How can we ensure that the lifescope of the InformationInterface is always greater than the InformationClient?

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportData.cpp
```diff
@@ -22,10 +22,18 @@ PXR_NAMESPACE_USING_DIRECTIVE
...
 namespace FVP_NS_DEF {
 
-ViewportInformationAndSceneIndicesPerViewportData::ViewportInformationAndSceneIndicesPerViewportData(const CreationParameters& creationParams)
-    : _viewportInformation(creationParams._viewportInformation), _renderIndexProxy(creationParams._renderIndexProxy)
+ViewportInformationAndSceneIndicesPerViewportData::ViewportInformationAndSceneIndicesPerViewportData(const InformationInterface::ViewportInformation& viewportInformation)
+    : _viewportInformation(viewportInformation)
 {
 }
 
+void ViewportInformationAndSceneIndicesPerViewportData::SetRenderIndexProxy(RenderIndexProxy* renderIndexProxy)
```
ppt-adsk: Again, I'm concerned about plain pointers and lifescope here.

lib/flowViewport/sceneIndex/fvpRenderIndexProxyFwd.h
```diff
@@ -0,0 +1,27 @@
...
+//
+#ifndef FVP_RENDER_INDEX_PROXY_FWD_H
+#define FVP_RENDER_INDEX_PROXY_FWD_H
+
+#include "flowViewport/api.h"
+
+
+namespace FVP_NS_DEF {
+
+class RenderIndexProxy;
```
ppt-adsk, line 23: What is the added value of this header?  You're not defining any additional types.  A client of RenderIndexProxy can do this themselves.  Unless you're planning to add type definitions for RenderIndexProxy related types (e.g. a vector of them, or a pointer to them), I would remove this file.

lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportDataManager.cpp
```diff
@@ -40,110 +41,75 @@ ViewportInformationAndSceneIndicesPerViewportDataManager& ViewportInformationAnd
...
 //A new Hydra viewport was created
-void ViewportInformationAndSceneIndicesPerViewportDataManager::AddViewportInformation(const InformationInterface::ViewportInformation& viewportInfo, Fvp::RenderIndexProxy& renderIndexProxy)
+void ViewportInformationAndSceneIndicesPerViewportDataManager::AddViewportInformation(const InformationInterface::ViewportInformation& viewportInfo)
 {
     //Add it in our array if it is not already inside
     {
         std::lock_guard<std::mutex> lock(_viewportInformationAndSceneIndicesPerViewportDataSet_mutex);
 
-        auto findResult = std::find_if(_viewportInformationAndSceneIndicesPerViewportDataSet.cbegin(), _viewportInformationAndSceneIndicesPerViewportDataSet.cend(),
+        auto findResult = std::find_if(_viewportsInformationAndSceneIndicesPerViewportData.cbegin(), _viewportsInformationAndSceneIndicesPerViewportData.cend(),
```
ppt-adsk: Pretty sure that since your set is ordered using viewportInfo, that you can simply use
```suggestion
        auto findResult = (_viewportsInformationAndSceneIndicesPerViewportData.count(other._viewportsInformationAndSceneIndicesPerViewportData) ==1);
```

ppt-adsk (changes requested):
Thanks for the tests!

ppt-adsk (approved):

debloip-adsk (approved):

PR #7, opened 2023-11-09
HYDRA-612 : Support USD v23.11

debloip-adsk:
This PR adds build support for USD v23.11.


CMakeLists.txt
```diff
@@ -23,11 +23,11 @@ project(maya-usd)
 option(BUILD_TESTS "Build tests." ON)
 option(BUILD_STRICT_MODE "Enforce all warnings as errors." ON)
 option(BUILD_SHARED_LIBS "Build libraries as shared or static." ON)
-option(BUILD_WITH_PYTHON_3 "Build with python 3." OFF)
+option(BUILD_WITH_PYTHON_3 "Build with python 3." ON)
```
seando-adsk: I would suggest just completely removing this option. You don't support building with py2 at all. So all the places this option is checked can be cleaned up to remove the py2 part.

seando-adsk (dismissed):

CMakeLists.txt
debloip-adsk: Sounds good to me, removed BUILD_WITH_PYTHON_3 and its usages

seando-adsk (approved):

PR #8, opened 2023-11-10
Update test README to clarify how to use MayaUSD in tests

debloip-adsk:
Added a mention of needing to build MayaHydra with the MayaUSD location set, specified that the directory must point to where the MayaUSD .mod file resides and provided an example usage of the build.py `--mayausd-location` parameter.


lanierd-adsk (approved):
Thanks for the update !

PR #12, opened 2023-11-16
HYDRA-687 : fix a test related to Usd Ufe lights

lanierd-adsk:


lanierd-adsk:
FYI : @lilike-adsk @debloip-adsk @roopavr-adsk

test/lib/mayaUsd/render/mayaToHydra/cpp/testMayaUsdUfeItems.cpp
```diff
@@ -35,7 +35,7 @@ bool IsUfeLight(const HdSceneIndexBasePtr& sceneIndex, const SdfPath& primPath)
     }
     // The shape prim is used to display the light's wireframe and is the only prim we want to allow
     // for UFE lights
-    bool isShapePrim = primPath.GetElementString().find("ufeLightProxyShape") < std::string::npos;
+    bool isShapePrim = primPath.GetElementString().find("ufeLightProxy") < std::string::npos;
```
debloip-adsk, line 38: This seems wrong : if a prim comes in named ufeLightProxy, this method will now return false when it should return true. The original bug that this test looks for is that previously, MayaUSD added its lights both through UFE and through its UsdStageSceneIndex, which led to the light being present twice. The fix I did was to filter out the UFE light (ufeLightProxy), but not its wireframe display (ufeLightProxyShape). So conceptually, this method returns primPath == "ufeLightProxy" (I can't recall if there was a reason I didn't do this directly and instead used find calls). So if this test fails, to me this points to a different issue.

debloip-adsk: Follow-up : I've taken a look at this and it seems that even though we still filter the UFE lights correctly, the ufeLightProxyShape prims now have a parent ufeLightProxy prim which is essentially just an empty prim that has the shape prim as its child. But there is still no actual light prim being created, which is good. We just need to adjust the test accordingly. I've opened #14 to change the check to a more robust one than the one I previously did. As for the reason I used std::string::find, it's because the prim names contain id numbers (e.g. ufeLightProxy1).

PR #17, opened 2023-11-23
HYDRA-456 : Add test for USD stage layer muting

debloip-adsk:
This PR adds a test for HYDRA-456/HYDRA-436 where muting a USD stage layer would not be reflected in the scene index hierarchy.

### DIFF
--- a/test/lib/mayaUsd/render/mayaToHydra/cpp/testUsdStageLayerMuting.py
+++ b/test/lib/mayaUsd/render/mayaToHydra/cpp/testUsdStageLayerMuting.py
@@ -1,0 +1,73 @@
+# Copyright 2023 Autodesk
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+import maya.cmds as cmds
+
+import fixturesUtils
+import mtohUtils
+import unittest
+
+from testUtils import PluginLoaded
+
+class TestUsdStageLayerMuting(mtohUtils.MayaHydraBaseTestCase):
+    # MayaHydraBaseTestCase.setUpClass requirement.
+    _file = __file__
+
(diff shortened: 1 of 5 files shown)


debloip-adsk:
FYI @lanierd-adsk @roopavr-adsk @lilike-adsk

test/lib/mayaUsd/render/mayaToHydra/cpp/testUtils.h
```diff
@@ -144,6 +144,27 @@ class SceneIndexInspector
...
+
+    /**
+     * @brief Predicate to match a prim name.
+     *
+     * @param[in] sceneIndex The scene index in which the prim path to test is contained.
+     * @param[in] primPath The prim path to test.
+     *
+     * @return True if the argument prim path's name matches the predicate's prim name, false otherwise.
+     */
+    bool operator()(const HdSceneIndexBasePtr& sceneIndex, const SdfPath& primPath) {
```
ppt-adsk: sceneIndex argument is unused?

test/lib/mayaUsd/render/mayaToHydra/cpp/testUsdStageLayerMuting.py
```diff
@@ -0,0 +1,73 @@
...
+
+    CUBE_PRIM_PATH = "/USDCube"
+    SUB_LAYER_CUBE_SIZE = 8.0
+
+    def setSubLayerMuted(self, isMuted: bool):
+        import maya.cmds as cmds
+        cmds.mayaUsdLayerEditor(self.subLayer.identifier, edit=True, muteLayer=(isMuted, self.usdProxyShapeUfePathString))
+
+    def setupUsdStage(self):
+        import maya.cmds as cmds
```
ppt-adsk: Nit-pick: importing maya.cmds twice, may as well factor it out.

ppt-adsk (dismissed):

test/lib/mayaUsd/render/mayaToHydra/cpp/testUtils.h
debloip-adsk: Yes, as this class is to be used as a FindPrimPredicate, whose signature includes a SceneIndex. This was indeed not explained, so I've renamed the unused scene index variable to `_`, mentioned that it is unused and clarified why that is.

test/lib/mayaUsd/render/mayaToHydra/cpp/testUsdStageLayerMuting.py
debloip-adsk: Thanks, that import was even already factored out at the top so those were useless.

ppt-adsk (approved):

PR #20, opened 2023-11-24
HYDRA-703 :  Fix mesh adapter transform update

debloip-adsk:
This PR fixes an issue with MeshAdapters where their transform was not being dirtied properly, and thus not updated correctly in the viewport. Since this issue was not MeshAdapter-specific and only highlighted a larger-scale problem, I've fixed the problem in the other places where it appeared.

### DIFF
--- a/lib/mayaHydra/hydraExtensions/adapters/cameraAdapter.cpp
+++ b/lib/mayaHydra/hydraExtensions/adapters/cameraAdapter.cpp
@@ -102,8 +102,8 @@ void MayaHydraCameraAdapter::CreateCallbacks()
         dag,
         +[](MObject& transformNode, MDagMessage::MatrixModifiedFlags& modified, void* clientData) {
             auto* adapter = reinterpret_cast<MayaHydraCameraAdapter*>(clientData);
-            adapter->MarkDirty(HdCamera::DirtyTransform);
             adapter->InvalidateTransform();
+            adapter->MarkDirty(HdCamera::DirtyTransform);
         },
         reinterpret_cast<void*>(this),
         &status);
--- a/lib/mayaHydra/hydraExtensions/adapters/dagAdapter.cpp
+++ b/lib/mayaHydra/hydraExtensions/adapters/dagAdapter.cpp
@@ -82,8 +82,8 @@ void _TransformNodeDirty(MObject& node, MPlug& plug, void* clientData)
         // that dirty as well...
         if (adapter->IsVisible(false)) {
             // Transform can change while dag path is hidden.
-            adapter->MarkDirty(HdChangeTracker::DirtyVisibility | HdChangeTracker::DirtyTransform);
             adapter->InvalidateTransform();
+            adapter->MarkDirty(HdChangeTracker::DirtyVisibility | HdChangeTracker::DirtyTransform);
         } else {
             adapter->MarkDirty(HdChangeTracker::DirtyVisibility);
         }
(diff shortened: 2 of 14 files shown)


lib/mayaHydra/hydraExtensions/adapters/cameraAdapter.cpp
```diff
@@ -102,8 +102,8 @@ void MayaHydraCameraAdapter::CreateCallbacks()
         dag,
         +[](MObject& transformNode, MDagMessage::MatrixModifiedFlags& modified, void* clientData) {
             auto* adapter = reinterpret_cast<MayaHydraCameraAdapter*>(clientData);
-            adapter->MarkDirty(HdCamera::DirtyTransform);
             adapter->InvalidateTransform();
+            adapter->MarkDirty(HdCamera::DirtyTransform);
```
debloip-adsk, line 106: There were a couple places where we were dirtying the transform before invalidating it. I've unified all instances of this sequence of operations to first invalidate the transform and then dirty it.

lib/mayaHydra/hydraExtensions/adapters/dagAdapter.cpp
```diff
@@ -210,9 +210,9 @@ void MayaHydraDagAdapter::CreateCallbacks()
 void MayaHydraDagAdapter::MarkDirty(HdDirtyBits dirtyBits)
 {
     if (dirtyBits != 0) {
-        GetSceneProducer()->GetRenderIndex().GetChangeTracker().MarkRprimDirty(GetID(), dirtyBits);
+        GetSceneProducer()->MarkRprimDirty(GetID(), dirtyBits);
```
debloip-adsk, line 213: This line is where everything originates, as it is what caused (and now fixed) the issue for MeshAdapter; however, this line being the problem only highlighted a larger problem, which is that we were still using the Hydra 1.0 way of dirtying prims in some places in our Maya-native SceneIndex (and this is why this PR may be a bit larger than expected).

lib/mayaHydra/hydraExtensions/mayaHydraLibInterface.h
```diff
@@ -26,7 +26,7 @@
 
 PXR_NAMESPACE_OPEN_SCOPE
 
-using SceneIndicesVector                                        = std::vector<HdSceneIndexBasePtr>;//Be careful, these are not not Ref counted. Elements could become dangling
+using SceneIndicesVector = std::vector<HdSceneIndexBaseRefPtr>;
```
debloip-adsk, line 29: See this comment and discussion : https://github.com/Autodesk/maya-hydra/pull/4#discussion_r1394377638. Doing this now because I added a utility observer class along with the unit test for this issue, and that class needs its observed SceneIndex to remain valid for its whole lifetime.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -943,18 +943,33 @@ void MayaHydraSceneIndex::_AddPrimAncestors(const SdfPath& path)
...
+
+void MayaHydraSceneIndex::MarkInstancerDirty(const SdfPath& id, HdDirtyBits dirtyBits)
+{
+    _MarkPrimDirty(id, dirtyBits, HdDirtyBitsTranslator::InstancerDirtyBitsToLocatorSet);
+}
+
+void MayaHydraSceneIndex::_MarkPrimDirty(
+    const SdfPath&          id,
+    HdDirtyBits             dirtyBits,
+    DirtyBitsToLocatorsFunc dirtyBitsToLocatorsFunc)
```
debloip-adsk, line 968: Broke up the previous MarkPrimDirty method into separate prim-category-specific methods : the reason for this is because there does not seem to be a reliable way to determine which DirtyBitsToLocatorSet method to use purely based off a prim's type. While this does look more like Hydra 1.0 in terms of API design, it also matches more closely how HdChangeTracker handles these calls for its own emulated scene index. When our different adapters all move to using Hydra 2.0 APIs, we can change this.

debloip-adsk:
FYI @ppt-adsk @roopavr-adsk @lanierd-adsk

lilike-adsk (dismissed):
good catch!

lilike-adsk (approved):

PR #23, opened 2023-12-06
HYDRA-718 : Fix added prims being at the wrong place when saving/loading a scene.

lanierd-adsk:


lanierd-adsk:
@lilike-adsk @ppt-adsk @debloip-adsk

roopavr-adsk (approved):

PR #26, opened 2023-12-13
Selection change functors now take scene index as argument.

ppt-adsk:
Static functors were capturing the selection scene index once, on construction.  This is incorrect, as the selection scene index is destroyed and re-created on switching Hydra renderers.

### DIFF
--- a/lib/mayaHydra/mayaPlugin/renderOverride.cpp
+++ b/lib/mayaHydra/mayaPlugin/renderOverride.cpp
@@ -941,23 +941,24 @@ void MtohRenderOverride::SelectionChanged(
     //    reading the Maya selection must be done from the Maya selection
     //    changed callback, not the UFE selection changed callback.
     using SnOp = Ufe::SelectionCompositeNotification::Op;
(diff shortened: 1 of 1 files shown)


lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -941,37 +941,38 @@ void MtohRenderOverride::SelectionChanged(
     //    reading the Maya selection must be done from the Maya selection
     //    changed callback, not the UFE selection changed callback.
     using SnOp = Ufe::SelectionCompositeNotification::Op;
-    static auto appendSn = [this](const SnOp& op) {
-        _selectionSceneIndex->AddSelection(op.item->path());
```
ppt-adsk, line 945: Capturing the selection scene index once, on creation of the static functor, is incorrect, as the selection scene index is destroyed and re-created.  We were therefore calling member functions on a destroyed object, causing crashes.

lanierd-adsk (approved):

PR #31, opened 2023-12-20
Fix testFlowViewportAPI image diff thresholds

debloip-adsk:
The tests in testFlowViewportAPI were expecting renders that matched exactly 1:1 the reference images, but renders can vary slightly between GPUs, so there was a slight difference between the reference images and OSX renders. This PR adds in some small thresholds to accept slight differences between a render and its reference image.


lanierd-adsk (approved):
Thanks for the change !

PR #33, opened 2023-12-21
HYDRA-764 : Add prim instancing unit test

debloip-adsk:
This PR adds a test for the fix for HYDRA-764. Note that the fix itself is in MayaUSD, see this PR : https://github.com/Autodesk/maya-usd/pull/3534


test/testUtils/usdUtils.py
```diff
@@ -31,6 +33,20 @@
...
+        Args:
+            filePath (str): The file path to the USD file to load.
+        Returns :
+            A string representing the UFE path to the stage's proxy shape node
+    """
+    # Source the corresponding mel file, otherwise the procedure isn't available
+    mel.eval('source \"mayaUsd_createStageFromFile.mel\"')
+    loadStageCmd = f'mayaUsd_createStageFromFilePath(\"{filePath}\")'
+    # Replace backslashes with forward slashes to avoid them being interpreted
+    loadStageCmd = loadStageCmd.replace('\\', '/')
```
debloip-adsk, line 48: This isn't great but it seems like the simplest way to get it to work, this is also what is done in MayaUSD ([example](https://github.com/Autodesk/maya-usd/blob/dev/test/lib/testMayaUsdCreateStageCommands.py#L72-L74)).

test/lib/mayaUsd/render/mayaToHydra/cpp/testPrimInstancing.cpp
```diff
@@ -0,0 +1,68 @@
...
+    ASSERT_TRUE(instancerDataSource);
+
+    // Ensure the instancer prim exists and is populated
+    SdfPath instancerPath = instancerDataSource->GetTypedValue(0);
+    auto findInstancerPredicate
+        = [instancerPath](const HdSceneIndexBasePtr& sceneIndex, const SdfPath& primPath) -> bool {
+        return primPath == instancerPath;
+    };
+    PrimEntriesVector instancerPrims = inspector.FindPrims(findInstancerPredicate);
+    ASSERT_EQ(instancerPrims.size(), 1u);
```
debloip-adsk, line 53: This is the main check, as the instancer referenced by the instanceable prim was previously not present without the fix.

ppt-adsk (approved):

debloip-adsk:
FYI @lanierd-adsk @roopavr-adsk @lilike-adsk

PR #34, opened 2023-12-22
HYDRA-700 : Avoid Boost autolinking libraries

debloip-adsk:
This PR adds the `BOOST_ALL_NO_LIB` definition to the `mayaHydra_compile_config` CMake function, which is used by all our targets. This avoids Boost trying to automagically link certain dependencies we don't truly need (e.g. `boost_python311-vc142-mt-x64-1_81.lib`), which could cause linker errors.

I've tested this using VS2019, for which we previously manually specified the `BOOST_PYTHON_NO_LIB` definition to work around the issue. With this change, this is no longer needed, and things compile as expected.


cmake/compiler_config.cmake
```diff
@@ -131,9 +131,10 @@ function(mayaHydra_compile_config TARGET)
         )
     endif()
 
-    # Remove annoying TBB warnings.
+    # Common definitions
     target_compile_definitions(${TARGET}
         PRIVATE
-            TBB_SUPPRESS_DEPRECATED_MESSAGES
+            TBB_SUPPRESS_DEPRECATED_MESSAGES # Remove TBB deprecation warnings
+            BOOST_ALL_NO_LIB # Avoid Boost autolinking libraries
```
ppt-adsk, line 145: Thanks for doing this!  Can you check all the places where we use BOOST_DEBUG_PYTHON and BOOST_LINKING_PYTHON?  I believe we can remove all these, as there is no Boost Python our own mayaHydra code, only in code we include from OpenUSD.

cmake/compiler_config.cmake
```diff
@@ -136,5 +136,6 @@ function(mayaHydra_compile_config TARGET)
         PRIVATE
             TBB_SUPPRESS_DEPRECATED_MESSAGES # Remove TBB deprecation warnings
             BOOST_ALL_NO_LIB # Avoid Boost autolinking libraries
+            $<$<STREQUAL:${CMAKE_BUILD_TYPE},Debug>:BOOST_DEBUG_PYTHON> # Auto-select Boost python version
```
ppt-adsk: Why do we need this?  We shouldn't have any Boost Python.

cmake/compiler_config.cmake
debloip-adsk: So it looks like Python is still being linked despite this. I've tried removing the BOOST_*_PYTHON definitions, keeping TBB_USE_DEBUG and moving it to the common definitions, and making the common definitions PUBLIC, but despite this it's still trying to link Python. The minimal setup I've found is to just use BOOST_DEBUG_PYTHON in the common definitions, and when that's there, everything else can be removed and things compile again.

debloip-adsk:
FYI @lilike-adsk @lanierd-adsk @roopavr-adsk

cmake/compiler_config.cmake
debloip-adsk: Update : I think I might've had something wrong in my setup back then? I just retried it, but instead got some other errors that will no longer happen once https://github.com/Autodesk/maya-hydra/pull/37 is merged. I tried building using those changes and removing the remaining BOOST_DEBUG_PYTHON, and it worked. So once the other PR is merged I think we should be able to get rid of the remaining BOOST_DEBUG_PYTHON.

cmake/compiler_config.cmake
debloip-adsk: (updated since then, see [this comment](https://github.com/Autodesk/maya-hydra/pull/34#discussion_r1451023214))

ppt-adsk (dismissed):

ppt-adsk (approved):

PR #36, opened 2024-01-08
HYDRA-474 : Filtering scene index example can have prims get unfiltered at the wrong location

debloip-adsk:
This PR fixes the issue where filtered prims who are no longer filtered could reappear in wrong locations.

The approach taken is to make it so that the filtering fully appears as prim additions and removals to downstream scene indices. The reason for this is that we do not know what downstream scene indices might do with the information they get from a filtering scene index and/or its parents. They could full well call GetPrim once and cache the result, and then only relying on notifications for their purpose, which means the filtering scene index has to stay coherent across its API, and send notifications which match its observable state. It also wouldn't make much sense for downstream scene indices to receive notifications about prims which are supposed to be filtered out.

This approach has the caveat of requiring either 1) to maintain state in order to remember which prims are filtered or not, in order to send the correct prim removal/addition notifications, or 2) wastefully re-syncing prims. For example, suppose we receive a prim dirty notification, and the prim does not need to be filtered : how do we know whether to simply forward the dirty notification (because it was already not filtered), or instead send a prim added notification (because it was previously filtered)? Either we keep track of whether the prim was previously filtered, or we always send prim added notifications, which acts as a full (potentially wasteful) re-sync (see this comment in USD).

### DIFF
--- a/lib/flowViewport/API/samples/fvpFilteringSceneIndexExample.cpp
+++ b/lib/flowViewport/API/samples/fvpFilteringSceneIndexExample.cpp
@@ -21,9 +21,6 @@
 #include <pxr/imaging/hd/primvarsSchema.h>
 #include <pxr/imaging/hd/meshSchema.h>
 #include <pxr/imaging/hd/tokens.h>
-#include <pxr/imaging/hd/xformSchema.h>
-
-#include <iostream>
 
 PXR_NAMESPACE_OPEN_SCOPE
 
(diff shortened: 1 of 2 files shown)


debloip-adsk:
We'll be waiting for input from Pixar on this, I've started a thread on the AOUSD forums : https://forum.aousd.org/t/pruning-prims-based-on-their-data-source-contents-and-handling-dirty-notifications/1111

PR #38, opened 2024-01-12
HYDRA-691 : Move mayaUsd code for managing stages inside mayaHydra

lanierd-adsk:


lanierd-adsk:
FYI : @roopavr-adsk @lilike-adsk @debloip-adsk

lanierd-adsk:
This won't be merged before our release is done.

lanierd-adsk:
FYI : @lilike-adsk , @roopavr-adsk , @debloip-adsk

doc/build.md
```diff
@@ -80,6 +80,14 @@ There are four arguments that must be passed to the script:
 |  --devkit-location    | directory where Maya devkit is installed.                                             |
 | workspace_location    | directory where the project use as a workspace to build and install plugin/libraries  |
 
+Optional arguments are :
+| Flags                 | Description                                                                                         |
+|--------------------   |---------------------------------------------------------------------------------------------------- |
+|   --mayausd-location  | directory where the mayaUSD.mod file is. By providing this, you will enable more features for       |
+|                       | usd stages when using an hydra render delegate, such as : hide/delete the stage when the proxy shape|
+|                       | node is hidden/deleted, or applying a transform on the proxy shape node will apply it on the stage. |
```
ppt-adsk: This sounds awkward, as if the version without --mayausd-location is buggy.  Would such a configuration be useful in production?  It would seem far preferable for customers to have a fully-working integration, or nothing at all.  Otherwise, QA has to test 2 configurations for USD, one which is arguably broken.

lib/mayaHydra/hydraExtensions/CMakeLists.txt
```diff
@@ -30,31 +30,54 @@ set(HEADERS
...
-        MAYAHYDRALIB_EXPORT
-        $<$<BOOL:${IS_MACOSX}>:OSMac_>
-        # Copy-pasted from lib/mayaUsd/CMakeLists.txt
-)
-
+if(DEFINED MAYAUSD_LOCATION)
+    target_compile_definitions(${TARGET_NAME} 
+        PRIVATE 
+            MAYAHYDRALIB_EXPORT
+            MAYAHYDRALIB_MAYAUSD_IS_USED_TO_BUILD
```
ppt-adsk: A bit wordy for a boolean variable.  How about:
```suggestion
            MAYAHYDRALIB_MAYAUSD_ENABLED
```

lib/mayaHydra/hydraExtensions/CMakeLists.txt
```diff
@@ -30,31 +30,54 @@ set(HEADERS
...
 # compiler configuration
 # -----------------------------------------------------------------------------
-target_compile_definitions(${TARGET_NAME} 
-    PRIVATE 
-        MAYAHYDRALIB_EXPORT
-        $<$<BOOL:${IS_MACOSX}>:OSMac_>
-        # Copy-pasted from lib/mayaUsd/CMakeLists.txt
-)
-
+if(DEFINED MAYAUSD_LOCATION)
```
ppt-adsk: I don't think you need this conditional, could simply be
```suggestion
$<$<BOOL:${MAYAUSD_LOCATION}>:MAYAHYDRALIB_MAYAUSD_ENABLED>
```

lib/mayaHydra/hydraExtensions/CMakeLists.txt
```diff
@@ -30,31 +30,54 @@ set(HEADERS
...
-            ${PXR_INCLUDE_DIRS}
-            ${BOOST_INCLUDE_DIR}
-            ${CMAKE_BINARY_DIR}/include
-        PRIVATE
-            $<$<BOOL:${UFE_FOUND}>:${UFE_INCLUDE_DIR}>
-    )
+# -----------------------------------------------------------------------------
+# include directories
+# -----------------------------------------------------------------------------
+if(DEFINED MAYAUSD_LOCATION)
```
ppt-adsk: See above using a CMake generator expression.

lib/mayaHydra/hydraExtensions/CMakeLists.txt
```diff
@@ -94,14 +116,43 @@ endif()
     # -----------------------------------------------------------------------------
     # link libraries
     # -----------------------------------------------------------------------------
-if(BUILD_MAYAUSD_LIBRARY)
+if(DEFINED MAYAUSD_LOCATION)
```
ppt-adsk: Again, a generator expression should be used.

lib/mayaHydra/hydraExtensions/sceneIndex/CMakeLists.txt
```diff
@@ -32,6 +33,7 @@ set(HEADERS
     mayaHydraSceneIndexDataFactoriesSetup.h
     mayaHydraMayaFilteringSceneIndexData.h
     mayaHydraMayaFilteringSceneIndexDataConcreteFactory.h
+    mayaHydraMayaUsdProxyShapeSceneIndex.h
```
ppt-adsk: With our recent changes to the coding guidelines I believe these file names should be:
```suggestion
    mhMayaUsdProxyShapeSceneIndex.h
```

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraMayaUsdProxyShapeSceneIndex.h
```diff
@@ -0,0 +1,111 @@
...
+
+//Usd/Hydra headers
+#include <pxr/base/tf/declarePtrs.h>
+#include <pxr/usdImaging/usdImaging/stageSceneIndex.h>
+#include <pxr/usdImaging/usdImaging/sceneIndices.h>// In USD 23.11+
+#include <pxr/imaging/hd/filteringSceneIndex.h>
+
+//Maya headers
+#include <maya/MObjectHandle.h>
+#include <maya/MMEssage.h>
```
ppt-adsk: This won't compile on Linux, where the file system is case-sensitive.
```suggestion
#include <maya/MMessage.h>
```

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraMayaUsdProxyShapeSceneIndex.h
```diff
@@ -0,0 +1,111 @@
...
+
+namespace MAYAHYDRA_NS_DEF {
+
+class MayaUsdProxyShapeSceneIndex;
+TF_DECLARE_WEAK_AND_REF_PTRS(MayaUsdProxyShapeSceneIndex);
+
+/// <summary>
+/// Simply wraps single stage scene index for initial stage assignment and population
+/// </summary>
+class MayaUsdProxyShapeSceneIndex : public HdSingleInputFilteringSceneIndexBase
```
ppt-adsk:
```suggestion
class MhMayaUsdProxyShapeSceneIndex : public HdSingleInputFilteringSceneIndexBase
```

build.py
```diff
@@ -556,7 +556,7 @@ def Package(context):
                     help="Directory where Maya is installed.")
 
 parser.add_argument("--mayausd-location", type=str,
-                    help="Directory where MayaUsd is installed.")
+                    help="Directory where mayaUSD.mod file is.")
```
seando-adsk: Actually I think the previous help was correct. This flag should be used to specify where the MayaUsd install is. Then you should create a `cmake/modules/FindMayaUsd.cmake` file where search for the new api library you want to link with and the include folder (setting two new cmake variables for these (similar to FindUfe.cmake), such as: MAYAUSD_INCLUDE_DIR and MAYAUSD_API_LIBRARY.

I think you could also add cmake code in this new FindMayaUsd.cmake that looks for the mayaUsd.mod file and if found adds the path to MAYA_MODULE_PATH.

doc/build.md
seando-adsk: I don't think its where the .mod is, but instead simply where MayaUsd is (the root folder).

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -60,10 +59,6 @@
 
 #include <ufeExtensions/Global.h>
 
-#if defined(MAYAUSD_VERSION)
-#include <mayaUsd/render/px_vp20/utils.h>
-#include <mayaUsd/utils/hash.h>
-#else
 namespace MayaUsd {
 // hash combiner taken from:
 // http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0814r0.pdf
```
ppt-adsk: This is all dead code, and should be removed.  We should never have code in the MayaUsd namespace in mayaHydra.

lib/mayaHydra/hydraExtensions/CMakeLists.txt
```diff
@@ -30,31 +30,54 @@ set(HEADERS
...
+            # Copy-pasted from lib/mayaUsd/CMakeLists.txt
+    )
+else()
+    target_compile_definitions(${TARGET_NAME} 
+        PRIVATE 
+            MAYAHYDRALIB_EXPORT
+            $<$<BOOL:${IS_MACOSX}>:OSMac_>
+            # Copy-pasted from lib/mayaUsd/CMakeLists.txt
+    )
+endif()
```
seando-adsk: Once you create the FindMayaUsd.cmake it will set MAYAUSD_FOUND (to TRUE for cmake). Then this can be simplified with a cmake generator expression.
```suggestion
target_compile_definitions(${TARGET_NAME} 
    PRIVATE 
        MAYAHYDRALIB_EXPORT
        $<$<BOOL:${IS_MACOSX}>:OSMac_>
        $<$<BOOL:${MAYAUSD_FOUND}>:MAYAHYDRALIB_MAYAUSD_IS_USED_TO_BUILD>
        # Copy-pasted from lib/mayaUsd/CMakeLists.txt
)
```

lib/mayaHydra/hydraExtensions/CMakeLists.txt
```diff
@@ -30,31 +30,54 @@ set(HEADERS
...
+            PUBLIC
+                ${MAYA_INCLUDE_DIRS}
+                ${PYTHON_INCLUDE_DIR}
+                ${PXR_INCLUDE_DIRS}
+                ${BOOST_INCLUDE_DIR}
+                ${CMAKE_BINARY_DIR}/include
+            PRIVATE
+                $<$<BOOL:${UFE_FOUND}>:${UFE_INCLUDE_DIR}>
+        )
+     endif()
```
seando-adsk: Can simplify with generator expression.
```suggestion
    target_include_directories(${TARGET_NAME} 
        PUBLIC
            ${MAYA_INCLUDE_DIRS}
            ${PYTHON_INCLUDE_DIR}
            ${PXR_INCLUDE_DIRS}
            ${BOOST_INCLUDE_DIR}
            ${CMAKE_BINARY_DIR}/include
        PRIVATE
            $<$<BOOL:${UFE_FOUND}>:${UFE_INCLUDE_DIR}>
            $<$<BOOL:${MAYAUSD_FOUND}>:${MAYAUSD_INCLUDE_DIR}>
    )
```

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1099,12 +1097,24 @@ void MtohRenderOverride::_PopulateSelectionList(
...
+                    if (foundPlace == std::string::npos) {
+                        TF_CODING_ERROR("pickedPathAsString.find(sceneIndexPathPrefixAsString) returned std::string::npos !");
+                        return;
+                    }
+
+                    //Keep only the right part of the string after the prefix
+                    const size_t placeToKeep = foundPlace + sceneIndexPathPrefixAsString.length();
+                    localPathAsString = pickedPathAsString.substr(placeToKeep, pickedPathAsString.length() - placeToKeep);
+
+                    const SdfPath localPath(localPathAsString);
```
ppt-adsk: I can see that you went to great lengths to replace the existing code, but on reading through I don't understand why we need convert to string / find prefix / replace prefix.  This seems to be way too much work.  Additionally, the SdfPath documentation recommends against converting to strings to do processing:
https://openusd.org/release/api/class_sdf_path.html#details

lib/mayaHydra/hydraExtensions/CMakeLists.txt
```diff
@@ -94,14 +116,43 @@ endif()
     # -----------------------------------------------------------------------------
     # link libraries
     # -----------------------------------------------------------------------------
-if(BUILD_MAYAUSD_LIBRARY)
+if(DEFINED MAYAUSD_LOCATION)
+    find_library(MAYAUSD_API_LIBRARY NAMES mayaUsdAPI PATHS ${MAYAUSD_LOCATION}/lib NO_DEFAULT_PATH)
+    if(NOT MAYAUSD_API_LIBRARY)
+        message(WARNING "Could not find mayaUsdAPI library.")
+    endif()
```
seando-adsk: This should be in the new FindMayaUsd.cmake

seando-adsk: Can use generator expression again to add the library (if found). No need for the `if(DEFINED MAYAUSD_LOCATION)`

            $<$<BOOL:${MAYAUSD_FOUND}>:${MAYAUSD_API_LIBRARY}>

seando-adsk: I'm pretty sure the `if(DEFINED MAYAUSD_VERSION)` can be simplified down to a single generator expression line as well. This might/should work:
`$<$<BOOL:${MAYAUSD_FOUND}>:MAYAUSD_VERSION=${MAYAUSD_VERSION}>`

seando-adsk (changes requested):

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraMayaUsdProxyShapeSceneIndex.cpp
```diff
@@ -0,0 +1,134 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#include "mayaHydraMayaUsdProxyShapeSceneIndex.h"
+
+#if defined(MAYAHYDRALIB_MAYAUSD_IS_USED_TO_BUILD)
```
ppt-adsk: I would be more assertive and do:
```suggestion
#if not defined(MAYAHYDRALIB_MAYAUSD_IS_USED_TO_BUILD)
#error MAYAHYDRALIB_MAYAUSD_ENABLED must be defined!
#endif
```

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraMayaUsdProxyShapeSceneIndex.cpp
```diff
@@ -0,0 +1,134 @@
...
+            _populated = true;
+        }
+    }
+}
+
+Ufe::Path MayaUsdProxyShapeSceneIndex::InterpretRprimPath(
+    const HdSceneIndexBaseRefPtr& sceneIndex,
+    const SdfPath&                path)
+{
+    if (MayaUsdProxyShapeSceneIndexRefPtr proxyShapeSceneIndex = TfStatic_cast<MayaUsdProxyShapeSceneIndexRefPtr>(sceneIndex)) {
```
ppt-adsk: This line is suspicious.  TfStatic_cast does not do a dynamic cast; it is therefore equivalent to doing a nullptr check on the input sceneIndex.  I wonder if a dynamic cast was intended?

lib/flowViewport/API/fvpFilteringSceneIndexClient.h
```diff
@@ -58,7 +58,7 @@ namespace FVP_NS_DEF
         *  @param[in]      rendererNames is the names of the renderers you want this client to be associated to.
         *                  If there are several, separate them with for example a coma, like "GL, Arnold", we actually look for the renderer name in this string.
         *                  If you want your client to work on any renderer please use FvpViewportAPITokens->allRenderers.
-        *  @param[in]      dccNode is a MObject* for Maya, if you provide the pointer value, then we automatically track some events such as visibility changed, 
+        *  @param[in]      dccNode is a MObject* DAG node for Maya, if you provide the pointer value, then we automatically track some events such as visibility changed,
```
ppt-adsk, line 61: If it's a Dag node, shouldn't we be passing in an MDagPath?

ppt-adsk (commented):
We need a test for this.

lib/flowViewport/API/fvpFilteringSceneIndexClient.h
ppt-adsk: I'm concerned about our support for DAG nodes.  To correctly support Maya instances a DAG path must be used.  Furthermore, since visibility is hierarchical, and therefore can be changed by reparenting a node, tracking visibility changes needs to deal with a DAG path and track changes to it.

CMakeLists.txt
```diff
@@ -127,6 +127,13 @@ else()
     message(STATUS "UFE not found.")
 endif()
 
+find_package(MAYAUSD OPTIONAL_COMPONENTS)
```
ppt-adsk: Nit-pick: why the all-caps naming?  UFE and USD packages are uppercase, because their name are uppercase, but MayaUsd is not --- e.g. look at the Python and Qt6 package names

test/lib/mayaUsd/render/mayaToHydra/testMayaUsdAPIUsage.py
ppt-adsk, line 1: Thanks for the test!

ppt-adsk (dismissed):
Only minor comments, approving.  From an architecture standpoint I am concerned about our use of MObject rather than MDagPath, but this is a wider concern, and should not block this pull request.

ppt-adsk (dismissed):

CMakeLists.txt
```diff
@@ -127,6 +127,13 @@ else()
     message(STATUS "UFE not found.")
 endif()
 
+find_package(MayaUsd OPTIONAL_COMPONENTS)
```
seando-adsk: I don't think you need the `OPTIONAL_COMPONENTS` - that is used when you want to specific a list of components which are optional. Since you didn't add `REQUIRED` this package is optional.

cmake/modules/FindMayaUsd.cmake
```diff
@@ -0,0 +1,90 @@
+#
+# Simple module to find MayaUsd.
+#
+# This module searches for a valid MayaUsd installation.
+# It searches for MayaUsd's libraries and include header files.
+#
+# Variables that will be defined:
+# MAYAUSD_FOUND           Defined if a MayaUsd installation has been detected
```
seando-adsk:
```suggestion
# MayaUsd_FOUND           Defined if a MayaUsd installation has been detected
```

cmake/modules/FindMayaUsd.cmake
```diff
@@ -0,0 +1,90 @@
...
+
+find_path(MAYAUSD_LIBRARY_DIR
+        mayaUsd.lib
+    HINTS
+        ${MAYAUSD_LOCATION}
+    PATH_SUFFIXES
+        lib/
+    DOC
+        "MayaUsd libraries path"
+)
```
seando-adsk: Why do you need to find this path? You just need the API one (below) - right? If you do need this one, then you'll have to adjust for linux/mac because the extension is not .lib on those platforms.

cmake/modules/FindMayaUsd.cmake
```diff
@@ -0,0 +1,90 @@
...
+
+
+# Handle the QUIETLY and REQUIRED arguments and set MAYAUSD_FOUND to TRUE if
+# all listed variables are TRUE.
+include(FindPackageHandleStandardArgs)
+
+find_package_handle_standard_args(MAYAUSD
+    REQUIRED_VARS
+        MAYAUSD_INCLUDE_DIR
+        MAYAUSD_LIBRARY_DIR
```
seando-adsk: The required library is the API, so that should be the required variable, along with the include dir. These will only be set when MayaUsd_FOUND is true. That way you use `MayaUsd_FOUND` in the generator expression BOOL test.
```suggestion
        MAYAUSDAPI_LIBRARY
```

cmake/modules/FindMayaUsd.cmake
```diff
@@ -0,0 +1,90 @@
...
+)
+
+if(MAYAUSD_FOUND)
+    message(STATUS "MayaUsd found : ${MAYAUSD_LOCATION}")
+    message(STATUS "MayaUsd include dir: ${MAYAUSD_INCLUDE_DIR}")
+    message(STATUS "MayaUsd libraries dir: ${MAYAUSD_LIBRARY_DIR}")
+    message(STATUS "MayaUsdAPI.lib fullpath : ${MAYAUSDAPI_LIBRARY}")
```
seando-adsk: I would not specify .lib - that is Windows specific.
```suggestion
    message(STATUS "MayaUsdAPI library : ${MAYAUSDAPI_LIBRARY}")
```

cmake/modules/FindMayaUsd.cmake
```diff
@@ -0,0 +1,90 @@
...
+    #Add MAYAUSD_MOD_PATH (the path where maya USD .mod file is) to the MAYA_MODULE_PATH
+    
+    # Get the current value of the environment variable
+    set(CURRENT_MAYA_MODULE_PATH $ENV{MAYA_MODULE_PATH})
+
+    # Append the new path to the current value
+    set(MAYA_MODULE_PATH "${CURRENT_MAYA_MODULE_PATH}:${MAYAUSD_MOD_PATH}")
+
+    # Export the new value to the environment
+    set(ENV{MAYA_MODULE_PATH} ${MAYA_MODULE_PATH})
```
seando-adsk: This should be done in `cmake\test.cmake`

cmake/modules/FindMayaUsd.cmake
```diff
@@ -0,0 +1,90 @@
...
+else()
+    if(DEFINED MAYAUSD_LOCATION)
+        message(WARNING "Could not find MayaUsd library and include directories in the folder provided : ${MAYAUSD_LOCATION}")
+    endif()
```
seando-adsk: Do you need this message? Because you didn't use `QUIET` in the `find_package(MayaUsd)` line, cmake should output a message if the package is not found.

lib/mayaHydra/hydraExtensions/CMakeLists.txt
```diff
@@ -30,52 +30,30 @@ set(HEADERS
...
-        PRIVATE 
-            MAYAHYDRALIB_EXPORT
-            $<$<BOOL:${IS_MACOSX}>:OSMac_>
-            # Copy-pasted from lib/mayaUsd/CMakeLists.txt
-    )
-endif()
+target_compile_definitions(${TARGET_NAME} 
+    PRIVATE 
+        MAYAHYDRALIB_EXPORT
+        $<$<BOOL:${MAYAUSDAPI_LIBRARY}>:MAYAHYDRALIB_MAYAUSDAPI_ENABLED>
```
seando-adsk:
```suggestion
        $<$<BOOL:${MayaUsd_FOUND}>:MAYAHYDRALIB_MAYAUSDAPI_ENABLED>
```

lib/mayaHydra/hydraExtensions/CMakeLists.txt
```diff
@@ -30,52 +30,30 @@ set(HEADERS
...
             PUBLIC
                 ${MAYA_INCLUDE_DIRS}
                 ${PYTHON_INCLUDE_DIR}
                 ${PXR_INCLUDE_DIRS}
                 ${BOOST_INCLUDE_DIR}
                 ${CMAKE_BINARY_DIR}/include
-        ${MAYAUSD_LOCATION}/include
             PRIVATE
                 $<$<BOOL:${UFE_FOUND}>:${UFE_INCLUDE_DIR}>
+                $<$<BOOL:${MAYAUSDAPI_LIBRARY}>:${MAYAUSD_INCLUDE_DIR}>
```
seando-adsk:
```suggestion
                $<$<BOOL:${MayaUsd_FOUND}>:${MAYAUSD_INCLUDE_DIR}>
```

lib/mayaHydra/hydraExtensions/CMakeLists.txt
```diff
@@ -113,77 +91,40 @@ if(DEFINED MAYAHYDRA_VERSION)
...
+        usdSkel
+        usdUtils
+        ufeExtensions
+        flowViewport
+        ${MAYA_LIBRARIES}
+    PRIVATE
+        $<$<BOOL:${USD_001905_BUILD}>:hio>
+        ${PYTHON_LIBRARIES} 
+        $<$<BOOL:${UFE_FOUND}>:${UFE_LIBRARY}>
+        $<$<BOOL:${MAYAUSDAPI_LIBRARY}>:${MAYAUSDAPI_LIBRARY}>
```
seando-adsk:
```suggestion
        $<$<BOOL:${MayaUsd_FOUND}>:${MAYAUSDAPI_LIBRARY}>
```

test/lib/mayaUsd/render/mayaToHydra/CMakeLists.txt
```diff
@@ -27,6 +27,11 @@ set(TEST_SCRIPT_FILES
     cpp/testFlowViewportAPIFilterPrims.py
 )
 
+#Add this test only if the MAYAUSDAPI_LIBRARY has been found during compile time.
+if(MAYAUSDAPI_LIBRARY)
```
seando-adsk: The reason is because `MAYAUSDAPI_LIBRARY` is a required variable so it must be set when `MayaUsd_FOUND` is true.
```suggestion
if(MayaUsd_FOUND)
```

seando-adsk (changes requested):
Looking good. Just a few minor changes.

cmake/test.cmake
```diff
@@ -1,5 +1,15 @@
 set(MAYA_USD_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 
+if(MayaUsd_FOUND)
```
seando-adsk, line 3: So here this should be either `MAYAUSD_MOD_PATH` or if you must have MAYAUSD_MOD_PATH then add it to the `REQUIRED_VARS` in FindMayaUsd.cmake.

cmake/test.cmake
```diff
@@ -1,5 +1,15 @@
...
+    #Add MAYAUSD_MOD_PATH (the path where maya USD .mod file is) to the MAYA_MODULE_PATH
+    # Get the current value of the environment variable
+    set(CURRENT_MAYA_MODULE_PATH $ENV{MAYA_MODULE_PATH})
+    # Append the new path to the current value
+    set(MAYA_MODULE_PATH "${CURRENT_MAYA_MODULE_PATH}:${MAYAUSD_MOD_PATH}")
```
seando-adsk: Does Windows accept `:` in this env var? I think it would be `;` for Windows and `:` for Mac/Linux.

cmake/modules/FindMayaUsd.cmake
seando-adsk: Can you test this by comment this out and then using an invalid path for `MAYAUSD_LOCATION` - I think cmake should print a warning for you.

seando-adsk (changes requested):
Just one minor change left and one question to answer.

cmake/test.cmake
```diff
@@ -5,9 +5,16 @@ if(MayaUsd_FOUND)
     # Get the current value of the environment variable
     set(CURRENT_MAYA_MODULE_PATH $ENV{MAYA_MODULE_PATH})
     # Append the new path to the current value
-    set(MAYA_MODULE_PATH "${CURRENT_MAYA_MODULE_PATH}:${MAYAUSD_MOD_PATH}")
+    if(IS_MACOSX OR IS_LINUX) 
+        #Linux and OSX
+        set(MAYA_MODULE_PATH "${CURRENT_MAYA_MODULE_PATH}:${MAYAUSD_MOD_PATH}")
+    else()
+        #Windows
+        set(MAYA_MODULE_PATH "${CURRENT_MAYA_MODULE_PATH};${MAYAUSD_MOD_PATH}")
```
seando-adsk: Sorry I should have remembered this earlier. The ; needs to be escaped otherwise cmake will interpret it to mean the string is a list. This is documented on line 78 in this file and used that way on line 343.

seando-adsk (changes requested):
One small change needed in the test.cmake

seando-adsk (dismissed):

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -808,7 +810,7 @@ void MtohRenderOverride::_InitHydraResources(const MHWRender::MDrawContext& draw
     _initializationSucceeded = true;
 }
 
-void MtohRenderOverride::ClearHydraResources()
+void MtohRenderOverride::ClearHydraResources(bool fullReset)
```
ppt-adsk, line 816: Would be very valuable to have a code comment explaining how to use this new argument.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -894,7 +901,8 @@ void MtohRenderOverride::_RemovePanel(MString panelName)
     }
 
     if (_renderPanelCallbacks.empty()) {
-        ClearHydraResources();
+        static const bool fullReset = false;
```
ppt-adsk: With modern C++ this should be
```suggestion
        constexpr bool fullReset = false;
```

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1265,7 +1273,8 @@ void MtohRenderOverride::_ClearHydraCallback(void* data)
     if (!TF_VERIFY(instance)) {
         return;
     }
-    instance->ClearHydraResources();
+    static const bool fullReset = true;
```
ppt-adsk:
```suggestion
    constexpr bool fullReset = true;
```

ppt-adsk (dismissed):
Some minor comments which you can choose to do or not.

ppt-adsk (approved):

PR #39, opened 2024-01-16
HYDRA-765 - fix the double transform issue on instanced prims

lilike-adsk:
Fix the double transform issue on instanced prims in the flow viewport example


lilike-adsk:
FYI @debloip-adsk @roopavr-adsk @ppt-adsk

lanierd-adsk (approved):

PR #42, opened 2024-01-18
HYDRA-747 : Filtering scene index example can have prims get unfiltered at the wrong location

debloip-adsk:
The issue in this case was that when the prim was added, it was initially pruned out. Then, when changing the number of vertices so that it is no longer pruned, only a PrimsDirtied notification was sent on the mesh. This means that downstream scene indices were never made aware of the prim's xform data source, or of the prim's addition.

This issue is more general in that pruning scene indices such as this one need to be aware of which prims are initially pruned out, so that they can provide a coherent scene view to downstream scene indices. The current solution to this issue is to traverse the scene in the scene index ctor and store which prims are initially pruned, and update this set in PrimsAdded and PrimsRemoved. That way, in PrimsDirtied, we can always correctly determine for a given prim whether to forward the dirty notification, or if we should send a PrimsAdded/PrimsRemoved if a prim's pruning status changed.

I have asked a question on the AOUSD forums regarding this issue : https://forum.aousd.org/t/pruning-prims-based-on-their-data-source-contents-and-handling-dirty-notifications/1111. The conclusion is that scene indices should always present a coherent view to to their observers, and send notifications as needed to keep it as such. So in our example case, either we send PrimsAdded and PrimsRemoved notifications every time the vertices change, potentially causing additional prim re-syncs, or we traverse the scene hierarchy initially, store prims' filtering status and send notifications accordingly when they are dirtied, potentially incurring extra load-time cost (or whenever we'd modify the scene index chain). The current implementation uses the latter approach.


debloip-adsk:
FYI @ppt-adsk @lilike-adsk @roopavr-adsk @vlasovi

lib/flowViewport/API/samples/fvpFilteringSceneIndexExample.cpp
```diff
@@ -64,20 +66,112 @@ namespace
...
-            return prim;//return only non filtered prims
+    return IsFiltered(primPath) ? HdSceneIndexPrim() : _GetInputSceneIndex()->GetPrim(primPath);
+}
+
+SdfPathVector FilteringSceneIndexExample::GetChildPrimPaths(const SdfPath& primPath) const {
+    if (IsFiltered(primPath)) {
+        return SdfPathVector();
+    }
+
+    SdfPathVector childPaths = _GetInputSceneIndex()->GetChildPrimPaths(primPath);
```
lanierd-adsk, line 115: Can you please add a comment here to say you're automatically hiding the children of an hidden prim ? It will help in case of debugging if required one day...  Thanks !

lanierd-adsk (changes requested):
Can you please add a unit test for the different cases ? So we are sure it works and we don't break it any more.

lib/flowViewport/API/samples/fvpFilteringSceneIndexExample.cpp
debloip-adsk: Done!

debloip-adsk:
> Can you please add a unit test for the different cases ? So we are sure it works and we don't break it any more.

I've added a check for the renderer switching case which is the original bug that started this, as well as for filtering and unfiltering in similar cases. However, I can't really test that child prims of a filtered prim are also filtered out, because from what I see our mesh prims are always leaf prims, so they never have children.

lanierd-adsk (approved):

ppt-adsk:
> > Can you please add a unit test for the different cases ? So we are sure it works and we don't break it any more.
> 
> I've added a check for the renderer switching case which is the original bug that started this, as well as for filtering and unfiltering in similar cases. However, I can't really test that child prims of a filtered prim are also filtered out, because from what I see our mesh prims are always leaf prims, so they never have children.

This will always be the case for Maya, since shape nodes must be leaves, and is the rule for well-constructed USD scenes as well --- see
https://openusd.org/release/glossary.html#usdglossary-gprim

PR #51, opened 2024-01-31
HYDRA-806 - Create a unit test for translate, rotate, scale

vlasovi:


vlasovi:

test/lib/mayaUsd/render/mayaToHydra/testTransforms.py
```diff
@@ -0,0 +1,108 @@
+# Copyright 2020 Luma Pictures
```
lanierd-adsk: Can you please remove the Luma copyright and set 2024 for the copyright from Autodesk ?
Unless you've taken that test from Luma somewhere ?

test/lib/mayaUsd/render/mayaToHydra/testTransforms.py
```diff
@@ -0,0 +1,108 @@
...
+
+import fixturesUtils
+import mtohUtils
+import unittest
+
+class TestTransforms(mtohUtils.MtohTestCase):
+    _file = __file__
+
+    IMAGEDIFF_FAIL_THRESHOLD = 0.01
+    IMAGEDIFF_FAIL_PERCENT = 0.1
```
lanierd-adsk, line 26: I think we use 2 for the percentage error , but this may fail if you're checking the selection highlight of a node for example. But maybe 0.1 is a bit low. We'll see from the preflight.

test/lib/mayaUsd/render/mayaToHydra/testTransforms.py
```diff
@@ -0,0 +1,108 @@
...
+        usdProxyShapeUfePathString = mayaUsd_createStageWithNewLayer.createStageWithNewLayer()
+        stage = mayaUsd.lib.GetPrim(usdProxyShapeUfePathString).GetStage()
+
+        # Define a cube prim in the stage
+        parentA = "/ParentA"
+        childCube = "/ParentA/Cube"
+        parentPrimA = stage.DefinePrim(parentA, 'Xform')
+        childPrimCube = stage.DefinePrim(childCube, 'Cube')
+        
+        self.verifySnapshot("usd_cube_untransformed.png")
```
lanierd-adsk, line 79: You should rather use :
self.assertSnapshotClose("usd_cube_untransformed.png", self.IMAGE_DIFF_FAIL_THRESHOLD, self.IMAGE_DIFF_FAIL_PERCENT)
which will use the threshold and diff fail percent variables.

test/lib/mayaUsd/render/mayaToHydra/testTransforms.py
ppt-adsk: Did Luma really make a contribution to this file?  Also, the copyright date is wrong.

test/lib/mayaUsd/render/mayaToHydra/testTransforms.py
```diff
@@ -0,0 +1,108 @@
...
+class TestTransforms(mtohUtils.MtohTestCase):
+    _file = __file__
+
+    IMAGEDIFF_FAIL_THRESHOLD = 0.01
+    IMAGEDIFF_FAIL_PERCENT = 0.1
+
+    def setUp(self):
+        self.imageVersion = None
+        if maya.mel.eval("defaultShaderName") != "standardSurface1":
+            self.imageVersion = 'lambertDefaultMaterial'
```
ppt-adsk: Why do we need this?

ppt-adsk (commented):
Thanks for the test, really minor comments.

test/lib/mayaUsd/render/mayaToHydra/testTransforms.py
vlasovi: I copied this code from testVisibility.py. In older Mayas (probably 2022 and older) they used lambert material as default and that gives different shading.

lanierd-adsk (changes requested):

test/lib/mayaUsd/render/mayaToHydra/testTransforms.py
vlasovi: @lanierd-adsk This is what is called inside verifySnapshot

test/lib/mayaUsd/render/mayaToHydra/testTransforms.py
ppt-adsk: We don't support any Maya older than 2024, and all our new releases support only Maya 2025.  You can therefore remove this code.

test/lib/mayaUsd/render/mayaToHydra/testTransforms.py
lanierd-adsk: ok, sorry, I missed that.

lanierd-adsk (approved):

ppt-adsk (approved):

PR #54, opened 2024-02-01
HYDRA-780 : Image-based NURBS unit test

debloip-adsk:


test/lib/mayaUsd/render/mayaToHydra/testNurbsPrimitives.py
```diff
@@ -0,0 +1,113 @@
...
+    IMAGE_DIFF_FAIL_THRESHOLD = 0.1
+    IMAGE_DIFF_FAIL_PERCENT = 0.75
+
+    def compareSnapshot(self, referenceFilename, cameraDistance=10):
+        self.setBasicCam(cameraDistance)
+        cmds.refresh()
+        self.assertSnapshotClose(referenceFilename, self.IMAGE_DIFF_FAIL_THRESHOLD, self.IMAGE_DIFF_FAIL_PERCENT)
+
+    def setupScene(self, nurbsCreationCallable):
+        cmds.loadPlugin('ArubaTessellator') # Loaded by default in Maya, changes NURBS vertices calculations
```
ppt-adsk: Feels awkward that this is done for each test (even though nothing bad will happen, just a plugin already loaded warning).  If this is the proper default behavior for mayaHydra tests, should we move this up into mtohUtils?

test/lib/mayaUsd/render/mayaToHydra/testNurbsPrimitives.py
```diff
@@ -0,0 +1,113 @@
...
+        self.setHdStormRenderer()
+        nurbsResult = nurbsCreationCallable()
+        cmds.refresh()
+        return nurbsResult
```
ppt-adsk: Nit-pick: looks like all the call sites are using the index 1 value from the return list, may as well save some typing and return nurbsResult[1].

ppt-adsk (dismissed):
Couple of minor points you may want to have a look at, otherwise fine, thanks for the test!

debloip-adsk:
FYI @lilike-adsk @lanierd-adsk @roopavr-adsk @vlasovi

test/lib/mayaUsd/render/mayaToHydra/testNurbsPrimitives.py
debloip-adsk: Good point, I've moved it to fixturesUtils as it is the common code between the base unit test class and the image-based one.

test/lib/mayaUsd/render/mayaToHydra/testNurbsPrimitives.py
debloip-adsk: Yeah I originally had it that way in case I needed the nurbs node, but in the end I didn't. Refactored it to directly return the makeNurb node.

ppt-adsk (approved):

PR #55, opened 2024-02-02
Adjust image diff threshold for testMayaLights

debloip-adsk:


lilike-adsk (commented):
is this for MacOS specifically? I'm tweaking the threshold for another test case,

debloip-adsk:
Yes for macOS specifically, it's causing failures in preflights e.g. for #54, I've checked the output images and it does match the expected result so it looks like just a threshold issue

lilike-adsk:
> Yes for macOS specifically, it's causing failures in preflights e.g. for #54, I've checked the output images and it does match the expected result so it looks like just a threshold issue

If that, could we disable the test on MacOS first? I found if we use a pretty large threshold, we might miss the real failures. Meanwhile, we need to investigate why MacOS produced different images than Win/Linux for most test cases.

test/lib/mayaUsd/render/mayaToHydra/testMayaLights.py
```diff
@@ -20,12 +20,23 @@
...
+import platform
+
 class TestMayaLights(mtohUtils.MtohTestCase): #Subclassing mtohUtils.MtohTestCase to be able to call self.assertSnapshotClose
     # MayaHydraBaseTestCase.setUpClass requirement.
     _file = __file__
 
-    IMAGE_DIFF_FAIL_THRESHOLD = 0.01
-    IMAGE_DIFF_FAIL_PERCENT = 0.2
+    @property
+    def imageDiffFailThreshold():
```
lilike-adsk: great, miss "self" as parameter?

debloip-adsk: Yes definitely, though somehow it ran fine locally without it. Thanks for catching it

lilike-adsk (approved):

PR #56, opened 2024-02-02
HYDRA-795 - add test to cover arnold lights

lilike-adsk:
Add test to cover arnold lights

Note: this test is skipped until Arnold plugin can be loaded in our test environment


test/lib/mayaUsd/render/mayaToHydra/testArnoldLights.py
```diff
@@ -0,0 +1,93 @@
...
+        #No Light mode
+        #cmds.modelEditor(panel, edit=True, displayLights="none")
+        #cmds.refresh()
+        #self.assertSnapshotClose("noLight" + imageSuffix + ".png", self.imageDiffFailThreshold, self.imageDiffFailPercent)
+
+    #Test arnold lights (e.g., aiSkyDomeLight,etc.) with a maya native sphere and usd sphere.
+    @unittest.skipUnless(mtohUtils.checkForMayaUsdPlugin(), "Requires Maya USD Plugin.")
+    def test_ArnoldLights(self):
+        cmds.file(new=True, force=True)
+
```
lanierd-adsk, line 80: I would rather use the following code to try to load mtoa and do the tests only if it was successfully loaded :
if (mayaUtils.loadPlugin("mtoa"))

So when we can load mtoa, your tests are run automatically and skipped when it cannot be loaded.

test/lib/mayaUsd/render/mayaToHydra/CMakeLists.txt
```diff
@@ -16,6 +16,8 @@ set(INTERACTIVE_TEST_SCRIPT_FILES
     testMayaLights.py
     testFlowViewportAPI.py
     testStageVariants.py
+# Skip until MtoA plugin can be loaded.
+#   testArnoldLights.py
```
lanierd-adsk: I woul keep it and rather use something in the test. Please see the comment below about :
if (mayaUtils.loadPlugin("mtoa"))

lanierd-adsk (changes requested):
I would keep the test alive and use 
if (mayaUtils.loadPlugin("mtoa"))
to skip it if mtoa cannot be loaded.

test/testUtils/mtohUtils.py
```diff
@@ -39,6 +39,9 @@ def checkForPlugin(pluginName: str):
 def checkForMayaUsdPlugin():
     return checkForPlugin('mayaUsdPlugin')
 
+def checkForMtoAPlugin():
+    return checkForPlugin('mtoa')
```
lanierd-adsk, line 43: Thanks for the updates. Does this checkForPlugin loads the plugin or just checks if it is loaded ?

lilike-adsk: I saw the code, seems it would load the plugin

lanierd-adsk (approved):

PR #60, opened 2024-02-05
HYDRA-783 : Curve tools unit tests

debloip-adsk:
This PR tests the various curve/arcs tools. Note that this mostly tests the commands that are used by the tools, as we can't interact with the viewport directly, so the tests don't map 1:1 with the tools, but it does cover their usage.


debloip-adsk:
FYI @lanierd-adsk @roopavr-adsk @lilike-adsk @vlasovi

ppt-adsk (approved):
Thanks for the test!

PR #63, opened 2024-02-08
Maya is no longer using a Preview Release number

seando-adsk:
Remove code (currently not used) that detected this.


ppt-adsk (approved):
Thanks for handling this.

PR #65, opened 2024-02-09
Add code coverage build support.

ppt-adsk:
Added support for code coverage using Clang.

A new build flag, --build-coverage, creates a new variant, called Coverage.  The build therefore creates instrumented object files in the Coverage variant directory.  The build flag must also be used to run the unit tests, which then produce profiling data.

llvm-profdata merge can then be used to combine the profiling data from each test run into a single file, which can then be used to produce HTML output using llvm-cov show.


ppt-adsk:
Most of the code changes are to implement workarounds to avoid crashes in the code coverage build, which have yet to be fully investigated:
-     Avoid directly calling _GetInputSceneIndex() for filtering scene indices, which crashes in the coverage build. Replaced by a workaround.
-     Leak scene indices at file new or process exit.  Proper scene index cleanup in this cases crashes in the coverage build, so leaking is the workaround.

test/lib/mayaUsd/render/mayaToHydra/testMeshes.py
```diff
@@ -29,6 +30,27 @@ class TestMeshes(mtohUtils.MayaHydraBaseTestCase):
...
+    def setUpClass(cls):
+        if cls._file is None:
+            raise ValueError("Subclasses of MayaHydraBaseTestCase must "
+                             "define `_file = __file__`")
+
+        meshAdapter = os.getenv('MAYA_HYDRA_USE_MESH_ADAPTER', 0)
+        fixturesUtils.setUpClass(cls._file, 'mayaHydra',
+                                 initializeStandalone=False,
+                                 suffix='_meshAdapter' if meshAdapter else '')
+
```
debloip-adsk, line 53: I think this should be in the base MayaHydraBaseTestCase setUpClass method, otherwise we'll run into the same issue if we add other tests that run with both RenderItemAdapter and MeshAdapter

debloip-adsk (changes requested):

test/lib/mayaUsd/render/mayaToHydra/testMeshes.py
debloip-adsk: After discussion, we'll leave it like this for now, see how things evolve and refactor as needed.

debloip-adsk (dismissed):

debloip-adsk (approved):

ppt-adsk:
All tests pass in coverage build, even when run in parallel.

PR #70, opened 2024-02-14
Added code coverage documentation.

ppt-adsk:
Provide documentation on Windows code coverage using clang / LLVM.

### DIFF
--- a/doc/codeCoverageInstructions.md
+++ b/doc/codeCoverageInstructions.md
@@ -1,0 +1,71 @@
+# Code Coverage For Windows
+
+Maya-hydra has support for obtaining code coverage on Windows platforms.
+
+The support uses two main tools: Clang for the compilation of maya-hydra and the LLVM toolset for the parsing of code coverage
+information and the generation of a code coverage report.
+
+## Prerequisites
+
+To install Clang and the LLVM toolset, you can install an optional module with Visual Studio.  Refer to these instructions: [Install Clang and LLVM Toolset](https://learn.microsoft.com/en-us/cpp/build/clang-support-msbuild?view=msvc-170)
+
+## Documentation References
+- [llvm-profdata and llvm-cov show](https://llvm.org/docs/CommandGuide/llvm-cov.html)
+- [Compiling clang with coverage](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html)
+
+## Building the Coverage Variant
+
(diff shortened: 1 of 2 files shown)


doc/codeCoverageInstructions.md
```diff
@@ -0,0 +1,71 @@
...
+- [Compiling clang with coverage](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html)
+
+## Building the Coverage Variant
+
+The maya-hydra build has a Coverage variant that can be used with the following stages: clean,build,install,test (note that the test stage depends on the install stage).  Clang is used with code coverage instrumentation flags enabled (-fprofile-instr-generate -fcoverage-mapping) so that when tests are run after a successful install stage, code coverage data files will be generated.
+
+To build the coverage variant you can run:
+
+```
+python build.py --generator=Ninja --stages clean,configure,build,install --maya-location <maya_location> --build-coverage --pxrusd-location <pxrusd_location> --devkit-location <devkit_location> --build-args="-DPYTHON_INCLUDE_DIR=<python_include_dir>,-DPython_EXECUTABLE=<python_executable>,-DPYTHON_LIBRARIES=<python_libraries>,-DBUILD_WITH_PYTHON_3=ON,-DBUILD_WITH_PYTHON_3_VERSION=<version_of_python>,-DQT_VERSION=6.5.0,-DCMAKE_WANT_MATERIALX_BUILD=ON,-DCMAKE_PREFIX_PATH=<cmake_prefix_path>" <installation_location>
```
debloip-adsk: Should we refer to the [build doc](https://github.com/Autodesk/maya-hydra/blob/dev/doc/build.md) here? Regardless, we could simplify the example slightly :
- `BUILD_WITH_PYTHON_3` no longer exists
- `BUILD_WITH_PYTHON_3_VERSION` is optional, by default automatically determined based on the Maya version
- `QT_VERSION` : not sure if this is even actually used (the CMake script uses `WANT_QT_VERSION`), but in any case, it is not required, as the Qt version to use is also automatically determined based on the Maya version

Same goes for the other example below

debloip-adsk (approved):
Thanks for the doc! I was able to generate and view the HTML report. I did get a `warning: 63 functions have mismatched data` when running the `llvm-cov` command, is that expected? The report looks fine to me and I can navigate individual files, so not sure what that's about

ppt-adsk:
> Thanks for the doc! I was able to generate and view the HTML report. I did get a `warning: 63 functions have mismatched data` when running the `llvm-cov` command, is that expected? The report looks fine to me and I can navigate individual files, so not sure what that's about

I get the same, but I haven't figured out which functions are involved.  Appears to be a bug with llvm-cov for dynamic libraries, possibly all platform:
https://github.com/llvm/llvm-project/issues/32849

PR #72, opened 2024-02-14
HYDRA-833 - Maya MaterialX material bindings on Maya geometry

vlasovi:
- Support MaterialX on Maya native prims
- Autotest


vlasovi:
FYI @lanierd-adsk @ppt-adsk @roopavr-adsk @debloip-adsk

test/lib/mayaUsd/render/mayaToHydra/testMaterialXOnNative.py
```diff
@@ -0,0 +1,41 @@
...
+        self.assertSnapshotClose(imageName, 
+                                 self.IMAGEDIFF_FAIL_THRESHOLD,
+                                 self.IMAGEDIFF_FAIL_PERCENT)
+
+    @unittest.skipUnless(mtohUtils.checkForPlugin('LookdevXMaya'), "Requires LookDevX Plugin.")
+    def test_MaterialX(self):
+        mayaUtils.openTestScene("testMaterialX", "RedMtlxSphere.ma")
+        self.setBasicCam(2)
+        self.setHdStormRenderer()
+        self.verifySnapshot("RedMtlxSphere.png")
```
lilike-adsk, line 38: are we missing to pass IMAGEDIFF_FAIL_THRESHOLD and IMAGEDIFF_FAIL_PERCENT here?

lib/mayaHydra/hydraExtensions/adapters/materialAdapter.cpp
```diff
@@ -231,10 +246,81 @@ class MayaHydraShadingEngineAdapter : public MayaHydraMaterialAdapter
...
+        return true;
+    }
+
     VtValue GetMaterialResource() override
     {
         TF_DEBUG(MAYAHYDRALIB_ADAPTER_MATERIALS)
             .Msg("MayaHydraShadingEngineAdapter::GetMaterialResource(): %s\n", GetID().GetText());
+        
+        HdMaterialNetworkMap materialXNetworkMap;
+        if (PopulateMaterialXNetworkMap(materialXNetworkMap)) {
```
lilike-adsk, line 320: is there any reason not to do the translation inside MayaHydraMaterialNetworkConverter? I feel it's better to move the code into MayaHydraMaterialNetworkConverter as MatX is also one kind of Material.

test/lib/mayaUsd/render/mayaToHydra/testMaterialXOnNative.py
vlasovi: No, they are passed from inside verifySnapshot method

lib/mayaHydra/hydraExtensions/adapters/materialAdapter.cpp
vlasovi: The interface of MayaHydraMaterialNetworkConverter seems to be developped specifically for UsdPreviewSurface network. I think we would have hard times adapting and rewriting it for the needs of MaterialX. For example method HdMaterialNode* GetMaterial would need to change its output type as well as all other public methods have no meaning for MaterialX (like GetPreviewShaderParams). I don't see any reason for readapting  MayaHydraMaterialNetworkConverter for MaterialX

test/lib/mayaUsd/render/mayaToHydra/testMaterialXOnNative.py
lilike-adsk: right, sorry, I missed that. :)

lib/mayaHydra/hydraExtensions/adapters/materialAdapter.cpp
lilike-adsk: ok, sounds like having a base materialConverter and diferent sub materialConverters would make the architecture looks better, as the actual conversion is the responsibility of materialConverter, not materialAdapter.

I'm fine with the current code, just trying to make it easier to maintain,  we could also refactor it later when we need more materialConverters in the future.

lilike-adsk (approved):

PR #75, opened 2024-02-15
HYDRA-307: Add the FootPrint node as an Hydra example

lanierd-adsk:


lanierd-adsk:
FYI : @lilike-adsk @debloip-adsk @roopavr-adsk @vlasovi

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
...
+# compiler configuration
+# -----------------------------------------------------------------------------
+target_compile_definitions(${TARGET_NAME}
+    PRIVATE 
+        # M3dView needs this (technically, MNativeWindowHdl.h)
+        # private atm because M3dView is only used in .cpp right now
+        # gcc will NOT set this automatically
+        $<$<BOOL:${IS_LINUX}>:LINUX>
+        # Not sure if msvcc sets this automatically, but won't hurt to redefine
+        $<$<BOOL:${IS_WINDOWS}>:_WIN32>
```
ppt-adsk: _WIN32 is defined by the compiler, so you shouldn't need this:
https://stackoverflow.com/questions/662084/whats-the-difference-between-the-win32-and-win32-defines-in-c

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
...
+# include directories
+# -----------------------------------------------------------------------------
+if(DEFINED MAYAUSD_VERSION)
+    target_compile_definitions(${TARGET_NAME}
+        PRIVATE
+            MAYAUSD_VERSION=${MAYAUSD_VERSION}
+    )
+endif()
+
+if(DEFINED MAYAHYDRA_VERSION)
```
ppt-adsk: This should not be optional, and is defined unconditionally in the top-level CMakeLists.txt.  Therefore, I suggest:
```suggestion
if(NOT DEFINED MAYAHYDRA_VERSION)
    message(FATAL_ERROR "No MAYAHYDRA_VERSION defined, aborting configuration." )
endif()
```

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
...
+        PRIVATE
+            MAYAHYDRA_VERSION=${MAYAHYDRA_VERSION}
+    )
+endif()
+if(DEFINED MAYAHYDRA_CUT_ID)
+    target_compile_definitions(${TARGET_NAME}
+        PRIVATE
+            MAYAHYDRA_CUT_ID=${MAYAHYDRA_CUT_ID}
+    )
+endif()
```
ppt-adsk: Where would this be defined?  I see no definition in the maya-hydra repo.  I also wonder what this would be useful for if it was defined.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
...
+        mayaUsd_add_rpath(rpath "../../${MAYAUSD_TO_USD_RELATIVE_PATH}/lib")
+    elseif(DEFINED PXR_USD_LOCATION)
+        mayaUsd_add_rpath(rpath "${PXR_USD_LOCATION}/lib")
+    endif()
+    if (IS_LINUX AND DEFINED MAYAUSD_TO_USD_RELATIVE_PATH)
+        mayaUsd_add_rpath(rpath "../../${MAYAUSD_TO_USD_RELATIVE_PATH}/lib64")
+    endif()
+    if(IS_MACOSX AND DEFINED MAYAUSD_TO_USD_RELATIVE_PATH)
+        mayaUsd_add_rpath(rpath "../../../../Maya.app/Contents/MacOS")
+    endif()
```
ppt-adsk: Nit-pick: the MAYAUSD_TO_USD_RELATIVE_PATH is already checked at line 74, could place these two OS-dependent conditionals there.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
```diff
@@ -0,0 +1,296 @@
...
+
+    @property
+    def imageDiffFailPercent(self):
+        return 0.2
+
+    def setupScene(self):
+        cmds.file(new=True, force=True)
+        testFile = mayaUtils.openTestScene( #Is an empty scene that is used to zoom in more on the foot print nodes, it also has Storm already set as the renderer
+                "testFootPrintNode",
+                "testFootPrintNode.ma")
```
ppt-adsk: If this scene is empty, and you're setting Storm as the renderer at line 47, what is the added value of testFootPrintNode.ma beyond doing file new, which you're already doing at line 42?  Unless I'm missing something, I would remove testFootPrintNode.ma.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
```diff
@@ -0,0 +1,296 @@
...
+    @property
+    def imageDiffFailThreshold(self):
+        return 0.01
+
+    @property
+    def imageDiffFailPercent(self):
+        return 0.2
+
+    def setupScene(self):
+        cmds.file(new=True, force=True)
```
ppt-adsk, line 43: MayaHydraBaseTestCase.setUp() already does this, so it's done before each test method, no need to repeat this.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
```diff
@@ -0,0 +1,296 @@
...
+        testFile = mayaUtils.openTestScene( #Is an empty scene that is used to zoom in more on the foot print nodes, it also has Storm already set as the renderer
+                "testFootPrintNode",
+                "testFootPrintNode.ma")
+        cmds.refresh()
+        self.setHdStormRenderer()#set Storm as the renderer (even if the scene is supposed to have it already)
+
+    #Test adding primitives
+    def test_AddingPrimitives(self):
+        with PluginLoaded('mayaHydraFootPrintNode'):
```
ppt-adsk, line 52: Not yours, but we should really add support in setUpClass() for this.  Potentially loading and unloading the mayaHydra plugin at each test method is wasteful.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
```diff
@@ -0,0 +1,296 @@
...
+            self.setupScene()
+        
+            #Create a mayaHydraFootPrintNode node which adds a dataProducerSceneIndex
+            footPrintNodeName = cmds.createNode("MhFootPrint")
+            self.assertFalse(footPrintNodeName == None)
```
ppt-adsk: I'm not sure what this is testing.  If a plugin is not found, a node of type "unknown" is created, with name "unknown1".  Therefore, I think this assertion will always succeed.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
```diff
@@ -0,0 +1,296 @@
...
+            
+            #Increase its size
+            cmds.setAttr(footPrintNodeName + '.size', 5)
+            cmds.refresh()
+            
+            #When the node above is created, its compute method is not called automatically, so work around to trigger a call to compute
+            cmds.setAttr(footPrintNodeName + '.dummyInput', 2)#setting this will set dirty the dummyOutput attribute
```
ppt-adsk: This shouldn't be necessary.  If the dummyOutput correctly depends on the size input, then setting size will dirty dummyOutput.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
```diff
@@ -0,0 +1,296 @@
...
+            cmds.getAttr(footPrintNodeName + '.dummyOutput')#getting this value will trigger a call to compute
+            cmds.refresh()
+            #Original images are located for example in maya-hydra\test\lib\mayaUsd\render\mayaToHydra\FlowViewportAPITest
```
ppt-adsk: I would remove this, it's a maintenance burden.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
```diff
@@ -0,0 +1,296 @@
...
+            self.assertSnapshotClose("add_NodeCreated.png", self.imageDiffFailThreshold, self.imageDiffFailPercent)
+
+            #Move the transform node, the added prims should move as well
+            # Get the transform node of the mayaHydraFootPrintNode
+            transformNode = cmds.listRelatives(footPrintNodeName, parent=True)[0]
+            self.assertFalse(transformNode == None)
```
ppt-adsk: Better to use assertIsNotNone(), error message on failure is much more informative:
https://codereview.doctor/features/python/best-practice/use-assert-is-not-none

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
```diff
@@ -0,0 +1,296 @@
...
+            #Select the transform node
+            cmds.select(transformNode)
```
ppt-adsk: Nit-pick: moving the selected object is fine, but you can just pass the transformNode as an argument to the move command.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
```diff
@@ -0,0 +1,296 @@
...
+            self.assertSnapshotClose("add_NodeMovedAfterDeletionAndUndo.png", self.imageDiffFailThreshold, self.imageDiffFailPercent)
+
+            #Switch to VP2
+            self.setViewport2Renderer()
+            #Switch back to Storm
+            self.setHdStormRenderer()
+            self.assertSnapshotClose("add_VP2AndThenBackToStorm.png", self.imageDiffFailThreshold, self.imageDiffFailPercent)
+
+            #Finish by a File New command
+            cmds.file(new=True, force=True)
```
ppt-adsk: No harm, but do we really need this?  setUp() will do a file new before every test method.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
```diff
@@ -0,0 +1,296 @@
...
+    def test_FootPrintAttributes(self):
+        with PluginLoaded('mayaHydraFootPrintNode'):
+            self.setupScene()
+        
+            #Create a mayaHydraFootPrintNode node which adds a dataProducerSceneIndex and a Filtering scene index
+            footPrintNodeName = cmds.createNode("MhFootPrint")
+            self.assertFalse(footPrintNodeName == None)
+
+            #When the node above is created, its compute method is not called automatically, so work around to trigger a call to compute
+            cmds.setAttr(footPrintNodeName + '.dummyInput', 2)#setting this will set dirty the dummyOutput attribute
```
ppt-adsk: I question the need for this.  We should be able to use the normal inputs to the footprint node.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
```diff
@@ -0,0 +1,296 @@
...
+            cmds.setAttr(footPrintNodeName2 + '.size', 1.5)
+            cmds.setAttr(footPrintNodeName2 + '.color', 0.0, 1.0, 0.0, type="double3")
+            cmds.refresh()
+            
+            #Move the transform node, the added prims should move as well
+            # Get the transform node of the mayaHydraFootPrintNode
+            transformNode2 = cmds.listRelatives(footPrintNodeName2, parent=True)[0]
+            self.assertFalse(transformNode2 == None)
+            #Select the transform node
+            cmds.select(transformNode2)
```
ppt-adsk: Shouldn't need to select to move.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
```diff
@@ -0,0 +1,296 @@
...
+
+    #Test loading a scene
+    def test_Load(self):
+        with PluginLoaded('mayaHydraFootPrintNode'):
+            cmds.file(new=True, force=True)
+            testFile = mayaUtils.openTestScene( 
+                "testFootPrintNode",
+                "testFootPrintNodeSaved.ma")
+            cmds.refresh()
+            self.assertSnapshotClose("loadingFootPrintScene.png", self.imageDiffFailThreshold, self.imageDiffFailPercent)
```
ppt-adsk, line 244: What is this testing?  Loading an empty scene?  Loading a scene should already be tested by Maya, I don't think we need to replicate this test.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -0,0 +1,666 @@
...
+
+    /// Init flag to do things only once for this node
+    bool _init = false;
+};
+
+namespace 
+{
+    // Foot print data
+    constexpr int soleVertsCount = 21;
+    constexpr int heelVertsCount = 17;
```
ppt-adsk: Both are unused, and feel like a maintenance burden.  You can always query solePoints.size().

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -0,0 +1,666 @@
...
+                                                      };
+
+    //For the maya bounding box
+    static const MPoint corner1( -0.17, 0.0, -0.7 );
+    static const MPoint corner2( 0.17, 0.0, 0.3 );
+    
+    //Create the Hydra primitive and add it to the retained scene index
+    void _CreateAndAddPrim(const HdRetainedSceneIndexRefPtr& retainedSceneIndex, const SdfPath& primPath, const VtArray<GfVec3f>& points, const VtIntArray& faceVertexCount, const VtIntArray& faceVertexIndices, const GfVec3f& scale, const GfVec3f& displayColor)
+    {
+        using _PointArrayDs = HdRetainedTypedSampledDataSource<VtArray<GfVec3f>>;
```
ppt-adsk, line 244: Repeated at line 411, can factor out to class level.

ppt-adsk: Repeated at line 412, can be factored out to class level.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -0,0 +1,666 @@
...
+    GfVec3f _GetColor() const;
+    ///Create the Hydra foot print primitives
+    void _CreateAndAddFootPrintPrimitives();
+    ///Remove the Hydra foot print primitives
+    void _RemoveFootPrintPrimitives();
+    ///Update the MObject of this node
+    void _UpdateThisMObject();
+
+    ///Counter to make the hydra primitives unique
+    static int _counter;
```
ppt-adsk: Does this need to be thread-safe (i.e. atomic)?

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -0,0 +1,666 @@
...
+        _CreateAndAddFootPrintPrimitives();
+        _init = true;
+    }
+
+    //The MObject can change if the node gets deleted and deletion being undone, so always update it in our records
+    MObject currentMObj = thisMObject();
+    if (_oldMObject.isNull() || _oldMObject != currentMObj){
+        
+        _thisMObject = currentMObj;     
+        _oldMObject  = _thisMObject;
```
ppt-adsk: Usually an MObjectHandle is used for this.  Would this meet your requirements?

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -0,0 +1,666 @@
...
+    unitFn.setWorldSpace( true );
+
+    mColor = nAttr.create("color", "col", MFnNumericData::k3Double, 1.0);
+    MAKE_INPUT(nAttr);
+    CHECK_MSTATUS ( nAttr.setDefault(0.0, 0.0, 1.0) );
+
+    //Create dummy input attribute to trigger a call to the compute function on demand. as it's in the compute fonction that we add our scene indices
+    mDummyInput = nAttr.create("dummyInput", "dI", MFnNumericData::kInt, 1.0);
+    MAKE_INPUT(nAttr);
+    CHECK_MSTATUS ( nAttr.setDefault(1) );
```
ppt-adsk: I question the need for this.  If the size and color attributes properly affect the output, we shouldn't need this dummy input.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -0,0 +1,666 @@
...
+        MFnNumericData fnData(oDouble3);
+        fnData.getData( outVal[0], outVal[1], outVal[2] );
+    }
+}//end of anonymous namespace
+
+//Static variables init
+int     MhFootPrint::_counter = 0;
+MObject MhFootPrint::mSize;
+MObject MhFootPrint::mColor;
+MTypeId MhFootPrint::id( 0x90573 );
```
ppt-adsk: I trust you registered this plugin ID wherever it now needs to be registered.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -0,0 +1,666 @@
...
+        return {(float)color[0], (float)color[1], (float)color[2]};
+    }
+
+    return GfVec3f(0.f,0.f,1.f);
+}
+
+MStatus MhFootPrint::compute( const MPlug& plug, MDataBlock& dataBlock)
+{
+    if( ! _init){
+        _CreateAndAddFootPrintPrimitives();
```
ppt-adsk: It seems strange to me that we need to use the Dataflow Graph compute method to register Hydra drawing.  Why can't this be done on node creation?

ppt-adsk (changes requested):
Great work David!  Some changes and questions sprinkled in.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/CMakeLists.txt
lanierd-adsk: I removed it.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: Right, I updated it.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: Good catch, I removed them.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: Thanks, the ones at line 411/412 were not needed, I removed them.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: No, I 've used that number almost randomly, let's take this offline.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: Because I need a valid MObject to use in the flow viewport API, that is not the case in the creation, it is invalid.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: Right, I am doing the change.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: Right. But this is done for tests as when creating a node by python nothing calls its compute function. And I'd like not to have to change its size or its color just for it to be displayed correctly so I am using this dummyinput/output to trigger a call to compute without changing its actual attributes.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
lanierd-adsk: It's actually a "placed camera" scene so we see better the footprint. When I used to fit the camera on the selection, it produced different results depending on the OS hence the usage of that empty scene but which places the camera near to the footprint.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
lanierd-adsk: ok, I didn't know that, I am updating it, thanks.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
lanierd-adsk: That is to avoid changing the default value of the attributes.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
lanierd-adsk: Ha yes, I want to do this to check because we could have a crash when cleaning up everything, that has happended in the past.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: I've fixed that now.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
lanierd-adsk: No, it loads a scene with the footprint node and checks if it is displayed correctly.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
ppt-adsk: Don't forget to create a pull request in Maya to indicate that you've taken 0x80101.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
ppt-adsk: But the "requires" line of testFootPrintNode.ma does not mention any footprint node, and I don't see any footprint node in the file.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
ppt-adsk: Got it, would be a nice code comment to add.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
ppt-adsk: But can't we do this programmatically?  Just get the camera and place it where you want.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
ppt-adsk: Can we move this code to node creation time?  How about using a path?  There does not need to be a node if we use a Ufe::Path, and an MDagPath can be built up.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -567,6 +567,20 @@ MBoundingBox MhFootPrint::boundingBox() const
 
 void* MhFootPrint::creator()
 {
+    int	isMayaHydraLoaded = false;
+    // Validate that the gpuCache plugin is loaded.
+    MGlobal::executeCommand( "pluginInfo -query -loaded mayaHydra", isMayaHydraLoaded );
+	if( ! isMayaHydraLoaded){
```
ppt-adsk: Indentation is awkward.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -567,6 +567,20 @@ MBoundingBox MhFootPrint::boundingBox() const
...
+        MStatus status;
+	    MString errorString = MStringResource::getString(rMayaHydraNotLoadedStringError, status);	     
+        if (! status){
+            status.perror("Cannot retrieve the rMayaHydraNotLoadedStringError string, but you need to load mayaHydra before creating this node");
+        }else{
+	        MGlobal::displayError(errorString);
```
ppt-adsk: Strange that we'd have 2 ways to report errors.

ppt-adsk (commented):
Looks pretty good, but I'd still like to explore the use of the compute() method to do scene index work.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: Let's take this offline, I tried unsuccessfully so far.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: It doesn't appear like that under Visual C++ but I tried to redo the indentation.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: They are 2 different errors, one because we didn't find the localized string, the other because maya hydra is not loaded.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
lanierd-adsk: It's not the same file that you mentionned, this file is :
"testFootPrintNode**Saved**.ma" not "testFootPrintNode.ma"

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
lanierd-adsk: I prefer loading a scene which has the same result on all platforms.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: Thanks, I've added one PR for that in maya.

lib/mayaHydra/flowViewportAPIExamples/flowViewportAPIMayaLocator/flowViewportAPIMayaLocator.cpp
```diff
@@ -271,7 +271,7 @@ namespace
 }//end of anonymous namespace
 
 //Initialization of static members
-MTypeId FlowViewportAPIMayaLocator::id( 0x00080100 );
+MTypeId FlowViewportAPIMayaLocator::id( 0x00080101 );
```
ppt-adsk, line 274: As per
https://git.autodesk.com/maya3d/maya/pull/27570
this is no longer correct, should be in the range 0x58000086 - 0x5800008F

test/lib/mayaUsd/render/mayaToHydra/testFlowViewportAPI.py
```diff
@@ -102,7 +102,7 @@ def test_AddingPrimitives(self):
             self.setHdStormRenderer()
             self.assertSnapshotClose("add_VP2AndThenBackToStorm.png", self.IMAGE_DIFF_FAIL_THRESHOLD, self.IMAGE_DIFF_FAIL_PERCENT)
 
-            #Finish by a File New command
+            #Finish by a File New command to check that it's not crashing when cleaning up everything
```
ppt-adsk: Sorry, should have caught this earlier, since you're calling file new after each test, you should factor this out into a tearDown() method:
https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDown

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
```diff
@@ -105,7 +105,7 @@ def test_AddingPrimitives(self):
             self.setHdStormRenderer()
             self.assertSnapshotClose("add_VP2AndThenBackToStorm.png", self.imageDiffFailThreshold, self.imageDiffFailPercent)
 
-            #Finish by a File New command
+            #Finish by a File New command to check that it's not crashing when cleaning up everything'
```
ppt-adsk: Same as testFlowViewportAPI.py, should use tearDown() for this.

ppt-adsk (changes requested):

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
ppt-adsk: Right, but why do we use MGlobal::displayError() in one case, and MStatus::perror() in the other?

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
ppt-adsk: Looks like it's fixed!

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
ppt-adsk: It seems very heavy-handed.  From what I can see the only relevant change in this file is moving the camera, which can trivially be done with
`cmds.move(7.714, 5.786, 7.714, 'persp')`
Also, the file introduces a dependence on the mayaUsdPlugin, which is incorrect.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
ppt-adsk: With the recent move to postConstructor, do we still need the dummyInput / dummyOutput?

lib/mayaHydra/flowViewportAPIExamples/flowViewportAPIMayaLocator/flowViewportAPIMayaLocator.cpp
lanierd-adsk: This is being done in another PR specific to the flow viewport node : https://github.com/Autodesk/maya-hydra/pull/85/files

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
lanierd-adsk: I just tried in the tests to remove these calls but it doesn't work. It is still needed.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
lanierd-adsk: I don't want to modify the defaults values to trigger a compute

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: I removed one of them.

test/lib/mayaUsd/render/mayaToHydra/testFlowViewportAPI.py
```diff
@@ -38,6 +38,11 @@ class TestFlowViewportAPI(mtohUtils.MtohTestCase): #Subclassing mtohUtils.MtohTe
     IMAGE_DIFF_FAIL_THRESHOLD = 0.1
     IMAGE_DIFF_FAIL_PERCENT = 2
 
+    @classmethod
```
ppt-adsk, line 41: This is not equivalent to what you had before: previously, you did a file new at the end of each test method.  As I suggested in
https://github.com/Autodesk/maya-hydra/pull/75/files/e3184227bf913bbbb1f48fb2e75ad2b8b5944239..eff7ba8689bcfe5db288d514517fba122b1e90a5#r1499561088
the proper equivalent is tearDown():
https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDown
If you're O.K. with tearDownClass() then that's fine too.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
```diff
@@ -38,12 +38,15 @@ def imageDiffFailThreshold(self):
     def imageDiffFailPercent(self):
         return 0.1
 
+    @classmethod
+    def tearDownClass(cls):
```
ppt-adsk: Same comment as in testFlowViewportAPI.py, should you be using tearDown() here rather than tearDownClass()?

test/lib/mayaUsd/render/mayaToHydra/testFlowViewportAPI.py
lanierd-adsk: You're right, I want to use tearDown.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: I integrated your changes where you removed the dummy attributes, thanks for your contribution.

test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py
ppt-adsk: Apologies, sorry for the confusion!

ppt-adsk (changes requested):
Looks really good, don't forget to change tearDownClass() to tearDown() in testFlowViewportAPI.py!

lanierd-adsk:
> Looks really good, don't forget to change tearDownClass() to tearDown() in testFlowViewportAPI.py!

Yes this is being made in another PR : https://github.com/Autodesk/maya-hydra/pull/85

ppt-adsk:
> > Looks really good, don't forget to change tearDownClass() to tearDown() in testFlowViewportAPI.py!
> 
> Yes this is being made in another PR : #85

O.K., change is not in that pull request yet, but if that's where you want to make it sounds good.

ppt-adsk (approved):

PR #76, opened 2024-02-16
HYDRA-811 - add test for template and reference

lilike-adsk:
add test to cover
1.display under template and reference mode
2.selection under template and reference mode


lilike-adsk:
FYI @lanierd-adsk @debloip-adsk @roopavr-adsk @ppt-adsk

vlasovi (approved):
Looks good

PR #77, opened 2024-02-16
Add Unit Test for Maya Shading Modes

roopavr-adsk:
Unit Test for the following shading modes (Shading menu in the viewport toolbar) for only Maya data

Flat Shade All
Bounding Box
X-Ray
X-Ray Joints
Backface Culling
Smooth Wireframe


test/lib/mayaUsd/render/mayaToHydra/testMayaShadingModes.py
```diff
@@ -0,0 +1,92 @@
...
+        cmds.modelEditor(panel, edit=True, displayAppearance="boundingBox")
+        cmds.refresh()
+        self.assertSnapshotClose("boundingBox" + ".png", self.imageDiffFailThreshold, self.imageDiffFailPercent)
+        cmds.modelEditor(panel, edit=True, displayAppearance="smoothShaded")
+
+        #Wirefame
+        cmds.modelEditor(panel, edit=True, smoothWireframe=True)
+        cmds.refresh()
+        self.assertSnapshotClose("wireframe" + ".png", self.imageDiffFailThreshold, self.imageDiffFailPercent)
+        cmds.modelEditor(panel, edit=True, smoothWireframe=False)
```
debloip-adsk: This doesn't seem to actually use wireframes : is that intended? Looking at the docs this option is to smooth the wireframes when they are there, the option to actually use wireframes is `displayAppearance="wireframe"`

debloip-adsk (dismissed):

PR #81, opened 2024-02-21
HYDRA-811 - add test for cover display & seleciton for object template

lilike-adsk:
Update the current testTemplateAndReference.py test to only cover display & seleciton when object template is on, as there's another test to cover Template&Reference on Display layer.

Note: A selection&highlighting issue was found and logged as https://jira.autodesk.com/browse/HYDRA-905


lilike-adsk:
FYI @lanierd-adsk @debloip-adsk @ppt-adsk @vlasovi

roopavr-adsk (approved):

PR #84, opened 2024-02-22
Made proxy shape scene index prefixes unique, and support rename / reparent

ppt-adsk:
When adding to the scene index scene, proxy shapes have to create a unique name under their rprim parent.  When translating application scene paths to rprim paths, any change in the application scene path to scene index prefix path must be tracked by the path interface scene index.  We therefore track renaming the proxy shape, or reparenting the proxy shape itself or an ancestor.


debloip-adsk (approved):
Approving, presumably the preflights are going to keep failing due to the compile-time MayaUSD API not found? If so and the tests pass locally, it might be worth force merging, as I have changes upcoming for HYDRA-834 which will likely have a few conflicts to resolve.

ppt-adsk:
> Approving, presumably the preflights are going to keep failing due to the compile-time MayaUSD API not found? If so and the tests pass locally, it might be worth force merging, as I have changes upcoming for HYDRA-834 which will likely have a few conflicts to resolve.

That would be awkward...  It would mean that all pre-flights would fail until the mayaUsdAPI support is added.  Could you not merge your branch with this branch and resolve conflicts that arise?

debloip-adsk:
Ah yeah I didn't think this through at all now did I... Problem is I was about to open a PR for HYDRA-834, so by merging both branches it couldn't be merged as well. Though we might want to wait until we have a USD build with the configurable float/double printing to run in our preflights before merging the PR for HYDRA-834 anyways, otherwise we'll be merging tests that aren't run yet, but will suddenly be in the future. Just feels a bit annoying that we have a couple dependent things floating around

PR #88, opened 2024-02-26
HYDRA-834 : Add support for text-based data source dump comparisons in tests

debloip-adsk:
This PR allows us to compare data sources dumped with `HdDebugPrintDataSource` in C++ tests. The printing of floats and doubles can be configured if the USD build used supports it, which allows us to print them with a certain precision and deal with floating point discrepancies between platforms. The compile definition `CONFIGURABLE_DECIMAL_STREAMING_AVAILABLE` indicates whether support for configurable float/double streaming is available.

Bundled with this PR is a test for some of the NURBS primitives, which while already tested using image diffing by #54, is the test that originally exposed the floating-point printing issue. This allows us to both test the changes to support data source comparison and make sure we successfully handle the original issue.


debloip-adsk:
FYI @lanierd-adsk @lilike-adsk @roopavr-adsk @vlasovi

ppt-adsk (approved):

PR #89, opened 2024-02-27
HYDRA-803 - add test for payload & reference

lilike-adsk:
add test for payload & reference


lilike-adsk:
FYI. @roopavr-adsk @ppt-adsk @vlasovi

debloip-adsk (approved):
Approving with a minor comment for the future : I think we should generally strive to have the objects in reference images take up a decent amount of space, so that we can detect potentially more subtle errors, e.g. the OSX one we've been having. I was a bit surprised to see we didn't need special OSX thresholds on this test, though I'm not sure if that's actually due to some of the images having small objects or instead due to using USD meshes.

lilike-adsk:
> Reference in n

there're no wires rendered, that's why it survived on Mac

debloip-adsk:
Ah true, forgot about that specificity, thanks!

PR #91, opened 2024-03-01
HYDRA-762 - Investigate why the tumbling performance of USD scene is slower than corresponding Maya Scene under Hydra Storm

vlasovi:
Fixing double-rendering of USD prims + autotest

### DIFF
--- a/lib/adskHydraSceneBrowser/test/adskHydraSceneBrowserTestFixture.cpp
+++ b/lib/adskHydraSceneBrowser/test/adskHydraSceneBrowserTestFixture.cpp
@@ -25,7 +25,6 @@
 #include <pxr/base/gf/vec3f.h>
 #include <pxr/base/vt/array.h>
 #include <pxr/base/vt/value.h>
-#include <pxr/imaging/hd/tokens.h>
 
 #include <gtest/gtest.h>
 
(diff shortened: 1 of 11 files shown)


vlasovi:
FYI : @lilike-adsk @roopavr-adsk @debloip-adsk @lanierd-adsk

lilike-adsk:
Nice catch! @vlasovi, I assume now, we can get similar fps on USD scene as corresponding Maya scene , right?

vlasovi:
> Nice catch! @vlasovi, I assume now, we can get similar fps on USD scene as corresponding Maya scene , right?

Yes

lib/adskHydraSceneBrowser/test/adskHydraSceneBrowserTestFixture.cpp
```diff
@@ -90,6 +91,60 @@ void AdskHydraSceneBrowserTestFixture::SetReferenceSceneIndex(
     sceneIndex = referenceSceneIndex;
 }
 
+void AdskHydraSceneBrowserTestFixture::VerifySceneCorrectness()
+{
+    // Traverse prim hierarchy
+    QTreeWidgetItemIterator itPrimsTreeWidget = GetIteratorForTree(_primHierarchyWidget);
```
ppt-adsk: Thanks for the test, looks good, but it feels out of place to tie it to the scene browser.  I asked @debloip-adsk and he agreed with me, saying that 
> with APIs like GetChildPrimPaths, GetPrim, Get methods for container data sources (or GetElement for vector data sources) + casts, and GetValue/GetTypeValue methods on data sources, the whole scene index scene hierarchy can be inspected.

Here's an example you might like:
https://github.com/Autodesk/maya-hydra/blob/dev/test/lib/mayaUsd/render/mayaToHydra/cpp/testPrimInstancing.cpp
It shouldn't change the logic of the test.

ppt-adsk (changes requested):
Looks good to me, though I wouldn't tie the test to the scene browser.

ppt-adsk (approved):
Great work, thanks for the test!

PR #94, opened 2024-03-04
HYDRA-843 : Add picking test + Link C++ tests with Qt for UI/viewport interaction

debloip-adsk:
This PR adds a unit test for picking. To do so, it also adds the ability to interact with the viewport (or more largely the UI as a whole) by linking the C++ tests plugin with Qt.


cmake/compiler_config.cmake
```diff
@@ -102,6 +102,51 @@ set(MSVC_DEFINITIONS
...
     # since it was added in Visual Studio 2015.
     HAVE_SNPRINTF
+
+    # In debug builds, Boost's wrap_python.hpp will #undef _DEBUG when BOOST_DEBUG_PYTHON is not defined,
+    # and will then include pyconfig.h :
+    # https://github.com/boostorg/python/blob/boost-1.81.0/include/boost/python/detail/wrap_python.hpp#L23-L57
+    # pyconfig.h will then autolink the Python lib; however, since _DEBUG was #undef'd, it will try to
+    # link against the release version of the Python lib, causing a linker error :
+    # https://github.com/python/cpython/blob/v3.11.4/PC/pyconfig.h#L269-L286
+    $<$<STREQUAL:${CMAKE_BUILD_TYPE},Debug>:BOOST_DEBUG_PYTHON>
```
debloip-adsk, line 112: Moved this from `target_compile_options` to here

cmake/compiler_config.cmake
```diff
@@ -102,6 +102,53 @@ set(MSVC_DEFINITIONS
...
+    # 11. Since there is no header guard for math.h, it is re-included.
+    # 12. As _USE_MATH_DEFINES is now defined, corecrt_math_defines.h does get included this time.
+    # 13. corecrt_math_defines.h defines the macros without checking if they have already been defined :
+    #     macro redefinition -> warning -> warning treated-as-error -> error -> compilation fails.
+    # In our case, step 9 (_USE_MATH_DEFINES getting defined late) happens in MTypes.h, and step 10 
+    # (math.h re-inclusion) in some other Maya headers, for example MFloatVector.h.
+    # By defining _USE_MATH_DEFINES from the get-go, we ensure that the math macros get defined by
+    # corecrt_math_defines.h the first time around. Afterwards, even if math.h does not have a
+    # header guard, the one in corecrt_math_defines.h avoids redefining the macros.
+    _USE_MATH_DEFINES
```
debloip-adsk, line 151: This is the new fix for the error I was getting in preflights. This might only be a problem when using Qt, but I'm putting this in the general `compiler_config.cmake` so we don't have to re-discover this again. There's also another discussion to be had about why this error only occurred in preflights, I'll follow up on that separately.

lib/adskHydraSceneBrowser/lib/CMakeLists.txt
```diff
@@ -136,10 +136,6 @@ target_include_directories(${TARGET_NAME}
 # -----------------------------------------------------------------------------
 # compiler configuration
 # -----------------------------------------------------------------------------
-# QT_NO_KEYWORDS prevents Qt from defining the foreach, signals, slots and emit macros.
-# this avoids overlap between Qt macros and boost, and enforces using Q_ macros.
-set_target_properties(Qt6::Core PROPERTIES INTERFACE_COMPILE_DEFINITIONS QT_NO_KEYWORDS)
```
debloip-adsk: Moved this to the top-level CMake in the Qt section to avoid having to copy-paste this in each Qt-enabled target

debloip-adsk:
Note : preflight is currently failing due to not finding LookdevX, a previous preflight did pass and the changes since then were just comments

cmake/compiler_config.cmake
seando-adsk: Are you sure this works? Generator expressions are only known at build stage. I don't think it works in a cmake `set()`

lib/adskHydraSceneBrowser/lib/CMakeLists.txt
seando-adsk: I actually don't agree with this change. I think that only the libraries in maya-hydra that need Qt should have the qt target property and the qt linked libs.

ppt-adsk: I like sharing Qt-related CMake code, but I agree with @seando-adsk that Qt dependencies should be reserved to those targets that need it.

debloip-adsk: Is that not the case? This line only runs if Qt was found (it is inside an `if (Qt6_FOUND)`), and from my understanding this is setting a property on the `Qt6::Core` target, not maya-hydra targets. I don't think this links Qt to maya-hydra targets either, does it?

seando-adsk: Yes it is setting a property on the Qt6::Core library. But here it would only do it for this library. By moving it to the make CMakeLists.txt you are setting that property for every library in MayaHydra. My feeling is that the individual libraries (in MayaHydra) that use Qt should each decide whether or not they want this property set.

cmake/compiler_config.cmake
debloip-adsk: The generator expression just gets stored as-is in the variable, and since we only use the variable in `target_compile_options`, it ends up spitting the generator expression back out, so it's pretty much equivalent. This could cause issues if we start trying to use the variable in pre-build stages and expecting it to be evaluated, but that seems somewhat unlikely to me. We have a couple generator expressions in MSVC_FLAGS already too (e.g. using `BUILD_STRICT_MODE` to treat warnings as errors). And a good indication that it works is that debug builds should have failed had it not worked (unless the issue this is fixing is no longer there, which I doubt).

lib/adskHydraSceneBrowser/lib/CMakeLists.txt
debloip-adsk: Ah that makes sense yeah, I've reverted it to be on a per-target basis, thanks

cmake/compiler_config.cmake
seando-adsk: Ah yes sorry I missed the fact that we already have generator expressions in these cmake `set()`. I'm okay with this change then.

seando-adsk (approved):

ppt-adsk (approved):
Great work, thanks for the test!  Looking forward to the increase in code coverage!

PR #99, opened 2024-03-06
Adjust image diff fail percent for shading modes test

debloip-adsk:
Seems like since MSAA was enabled this test can sometimes fail in preflights (Windows only it seems) due to a very minimal change along the wireframes. Possibly different Windows machines with different GPUs?


roopavr-adsk:
Looks like a needed change. Did you notice this on your or any particular Windows machines?

debloip-adsk:
Doesn't seem to happen on my particular machine, I encountered this when running preflights for #94

roopavr-adsk (approved):

PR #100, opened 2024-03-12
HYDRA-574 : Remove SceneDelegate code path

roopavr-adsk:
Removed Scene Delegate code path option/dependency. Updated related docs and unit tests to reflect this change. 

This change will break compatibility if clients have MAYA_HYDRA_ENABLE_NATIVE_SCENE_INDEX env var set.


lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright 2019 Luma Pictures
```
lilike-adsk: maybe double check with Sylvain, are we fine to change the copyright info here?

lilike-adsk (commented):
nice cleaning work!

lib/mayaHydra/mayaPlugin/renderOverride.cpp
roopavr-adsk: Yep. I asked the same question on the side. I would think we may need to keep it. Will update when I have more info.

lib/mayaHydra/hydraExtensions/mayaHydraSceneProducer.h
ppt-adsk, line 1: I think that we should get rid of MayaHydraSceneProducer.  I believe it was introduced as a wrapper for scene index / scene delegate, and now that we only have scene index, it only forwards calls to scene index.  My intuition is that we should move the render index proxy into MayaHydraSceneIndex, but this would have to be investigated.

lib/mayaHydra/hydraExtensions/mayaHydraSceneProducer.cpp
ppt-adsk, line 1: As per my comments in mayaHydraSceneProducer.h, I think we should get rid of MayaHydraSceneProducer.

lib/mayaHydra/hydraExtensions/mayaHydraSceneProducer.h
lilike-adsk: Good catch, Pierre! yes, MayaHydraSceneProducer could be removed together.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
roopavr-adsk: keeping it is the way to go.

lib/mayaHydra/hydraExtensions/mayaHydraSceneProducer.cpp
roopavr-adsk: Good point. I had a quick peek at what this change would entail and seems like simple enough change. There are just a couple of places where I feel we might need to be a bit careful (scene population and object lifecycle management). 

How about we introduce this change in a follow up PR? This way debugging could be easier to track in case of regression?

ppt-adsk (approved):

lilike-adsk (approved):

PR #102, opened 2024-03-18
Add mayaHydra plugin commands to query CPU/GPU memory usage

roopavr-adsk:
CPU and GPU usage can now be queried with mayaHydra using python commands when the plugin is loaded.

Sample usage:

```
import maya.cmds as cmds

cpuMem = cmds.mayaHydra(currentProcessMemory=1)
gpuMem = cmds.mayaHydra(hdGPUMem=1)
print("\nCPU Memory(MB): ", cpuMem[0])
print("\nGPU Memory(MB): ", gpuMem[0])
```

### DIFF
--- a/lib/mayaHydra/hydraExtensions/mixedUtils.h
+++ b/lib/mayaHydra/hydraExtensions/mixedUtils.h
@@ -30,7 +30,61 @@
 #include <maya/MMatrix.h>
 #include <maya/MObject.h>
 
+#if defined(_WIN32)
+#include <windows.h>
+#include "psapi.h"
+#elif defined(__linux__)
+#include <fstream>
+#include <sstream>
+#elif defined(__APPLE__)
+#include <mach/task.h>
+#include <mach/mach.h>
+#endif
+
 namespace MAYAHYDRA_NS_DEF {
+
+constexpr int MB = 1024 * 1024;
+
+// Function to get memory usage of the current process
+MAYAHYDRALIB_API
+inline int getProcessMemory()
+{
+#if defined(_WIN32)
+    // see https://learn.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-getprocessmemoryinfo
+    PROCESS_MEMORY_COUNTERS_EX pmc;
+    GetProcessMemoryInfo(GetCurrentProcess(), (PROCESS_MEMORY_COUNTERS*)&pmc, sizeof(pmc));
+    return static_cast<int>(pmc.WorkingSetSize / MB); // Working Set Size in MB
+
+#elif defined(__linux__)
+    // https://man7.org/linux/man-pages/man5/proc.5.html
+    // When a process accesses this magic symbolic link, it
+    // resolves to the process's own /proc/pid directory.
+    std::ifstream processFile("/proc/self/status");
+    std::string line;
+    unsigned long long memoryUsage = 0;
+
+    while (std::getline(processFile, line)) {
+        if (line.compare(0, 6, "VmRSS:") == 0) {
+            std::istringstream iss(line.substr(7));
+            iss >> memoryUsage;
+            break;
+        }
+    }
+    return  static_cast<int>(memoryUsage / MB);
+    
+#elif defined(__APPLE__)
+    // https://developer.apple.com/documentation/kernel/1537934-task_info
+    task_vm_info_data_t vmInfo;
+    mach_msg_type_number_t count = TASK_VM_INFO_COUNT;
+    kern_return_t kernResult = task_info(mach_task_self(), TASK_VM_INFO, (task_info_t)&vmInfo, &count);
+
+    if (kernResult == KERN_SUCCESS) {
+        return static_cast<int>(vmInfo.phys_footprint / MB); // Physical footprint in MB
+    } else {
+        return 0; // Unable to retrieve memory usage
+    }
+#endif
+}
 
 /**
  * @brief Converts a Maya matrix to a double precision GfMatrix.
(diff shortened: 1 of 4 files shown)


lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -340,6 +340,26 @@ void MtohRenderOverride::UpdateRenderGlobals(
...
+}
+
+int MtohRenderOverride::GetHdStGPUMemory()
+{   
+    int totalGPUMemory = 0;
+    std::lock_guard<std::mutex> lock(_allInstancesMutex);
+    for (auto* instance : _allInstances) {
+        totalGPUMemory += instance->_GetHdStGPUMemory().UncheckedGet<int>();
+    }
+    return totalGPUMemory;
```
roopavr-adsk: I think Hydra reports this in bytes but it's not clear from the code. I could convert it to MB like in the code for CPU.

lib/mayaHydra/mayaPlugin/viewCommand.cpp
```diff
@@ -41,6 +42,12 @@ constexpr auto _listRenderersLong = "-listRenderers";
 constexpr auto _listActiveRenderers = "-lar";
 constexpr auto _listActiveRenderersLong = "-listActiveRenderers";
 
+constexpr auto _currentProcessMemory = "-cpm";
+constexpr auto _currentProcessMemoryLong = "-currentProcessMemory";
+
+constexpr auto _hdStGPUMem = "-hdm";
+constexpr auto _hdStGPUMemLong = "-hdStGPUMem";
```
ppt-adsk: Awkward that this is Storm-centric.  What if the renderer is not Storm?  This to me feels like a renderer capability that we should be able to interrogate.  For certain software-only renderers this measure will be irrelevant.  For GPU-enabled renderers some renderers will be able to provide the information, others not.

lib/mayaHydra/hydraExtensions/mixedUtils.h
```diff
@@ -30,8 +30,62 @@
...
+
+// Function to get memory usage of the current process
+MAYAHYDRALIB_API
+inline int getProcessMemory()
+{
+#if defined(_WIN32)
+    // see https://learn.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-getprocessmemoryinfo
+    PROCESS_MEMORY_COUNTERS_EX pmc;
+    GetProcessMemoryInfo(GetCurrentProcess(), (PROCESS_MEMORY_COUNTERS*)&pmc, sizeof(pmc));
+    return static_cast<int>(pmc.WorkingSetSize / MB); // Working Set Size in MB
```
lilike-adsk: Please get the Virtual memory (pmc.PrivateUsage) used by current process, instead of physical memory.

lib/mayaHydra/hydraExtensions/mixedUtils.h
```diff
@@ -30,8 +30,62 @@
...
+#elif defined(__linux__)
+    // https://man7.org/linux/man-pages/man5/proc.5.html
+    // When a process accesses this magic symbolic link, it
+    // resolves to the process's own /proc/pid directory.
+    std::ifstream processFile("/proc/self/status");
+    std::string line;
+    unsigned long long memoryUsage = 0;
+
+    while (std::getline(processFile, line)) {
+        if (line.compare(0, 6, "VmRSS:") == 0) {
```
lilike-adsk: same as above

lib/mayaHydra/hydraExtensions/mixedUtils.h
```diff
@@ -30,8 +30,62 @@
...
+    return  static_cast<int>(memoryUsage / MB);
+    
+#elif defined(__APPLE__)
+    // https://developer.apple.com/documentation/kernel/1537934-task_info
+    task_vm_info_data_t vmInfo;
+    mach_msg_type_number_t count = TASK_VM_INFO_COUNT;
+    kern_return_t kernResult = task_info(mach_task_self(), TASK_VM_INFO, (task_info_t)&vmInfo, &count);
+
+    if (kernResult == KERN_SUCCESS) {
+        return static_cast<int>(vmInfo.phys_footprint / MB); // Physical footprint in MB
```
lilike-adsk: same as above

lib/mayaHydra/mayaPlugin/renderOverride.h
```diff
@@ -125,6 +125,9 @@ class MtohRenderOverride : public MHWRender::MRenderOverride
     MStatus setup(const MString& destination) override;
     MStatus cleanup() override;
 
+    // Utility function to get HdStorm GPU memory usage stats
+    static int GetHdStGPUMemory();
```
lilike-adsk: probably better to rename it as GetUsedGPUMemory, assuming _renderDelegate->GetRenderStats() is renderer agnostic

lib/mayaHydra/mayaPlugin/viewCommand.cpp
```diff
@@ -41,6 +42,12 @@ constexpr auto _listRenderersLong = "-listRenderers";
 constexpr auto _listActiveRenderers = "-lar";
 constexpr auto _listActiveRenderersLong = "-listActiveRenderers";
 
+constexpr auto _currentProcessMemory = "-cpm";
+constexpr auto _currentProcessMemoryLong = "-currentProcessMemory";
+
+constexpr auto _hdStGPUMem = "-hdm";
```
lilike-adsk: remove "St"

lilike-adsk (changes requested):

lilike-adsk (dismissed):
looks good!

lilike-adsk (approved):

lib/mayaHydra/mayaPlugin/viewCommand.cpp
roopavr-adsk: The "storm" centric-ness in the flag names have been removed in the latest commit. It is now just `hdGPUMem`. 

GetRenderStats() from a renderer's Render Delegate could potentially be any arbitrary data(VtDictionary). As of now, since our tests are mainly with Storm(also the only GPU renderer), the code only returns a valid value in that case else it returns 0. This behavior could be updated in future as we start testing against more offline/gpu renderers.

Hydra has tokens like `HdPerfTokens->gpuMemoryUsed` which we could possibly leverage to construct a better interface. I think it's best we do that when we have additional data points with more renderers

PR #103, opened 2024-03-19
HYDRA-931 : implement basics of bounding box display style

lanierd-adsk:


lanierd-adsk:
FYI : @ppt-adsk @roopavr-adsk @debloip-adsk @vlasovi

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -134,6 +135,19 @@ void replaceSelectionTask(PXR_NS::HdTaskSharedPtrVector* tasks)
...
+// return true if we need to recreate the filtering scene indices chain because of a change, false otherwise.
+bool NeedToRecreateTheSceneIndicesChain(unsigned int currentDisplayStyle, unsigned int previousDisplayStyle)
+{
+    const bool currentlyUsingBBoxDisplayStyle = currentDisplayStyle & MHWRender::MFrameContext::kBoundingBox;
+    const bool previouslyUsingBBoxDisplayStyle = previousDisplayStyle & MHWRender::MFrameContext::kBoundingBox;
+    if (currentlyUsingBBoxDisplayStyle != previouslyUsingBBoxDisplayStyle){
+        return true;
+    }
+
+    return false;
```
ppt-adsk, line 148: Nit-pick: can be simply `return currentlyUsingBBoxDisplayStyle != previouslyUsingBBoxDisplayStyle;`

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -611,15 +625,34 @@ MStatus MtohRenderOverride::Render(
...
+            M3dView view;
+            if (M3dView::getM3dViewFromModelPanel(panelName, view)){
+                MDagPath dpath;
+                view.getCamera(dpath);
+                MFnCamera viewCamera(dpath);
+                cameraName = viewCamera.name().asChar();
+            }
+            const Fvp::InformationInterface::ViewportInformation hydraViewportInformation(std::string(panelName.asChar()), cameraName);
+            manager.AddViewportInformation(hydraViewportInformation, _renderIndexProxy, _lastFilteringSceneIndexBeforeCustomFiltering);
+        }
```
ppt-adsk, line 655: Why do we need to re-create the scene index chain?  Can't we just dirty the whole scene and redraw with the new style?

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -953,6 +989,14 @@ void MtohRenderOverride::_CreateSceneIndicesChainAfterMergingSceneIndex()
     // selection highlighting.
     wfSi->addExcludedSceneRoot(_ID);
     _lastFilteringSceneIndexBeforeCustomFiltering  = wfSi;
+    
+    const unsigned int currentDisplayStyle = drawContext.getDisplayStyle();
+    
+    //Are we using Bounding Box display style ?
+    if (currentDisplayStyle & MHWRender::MFrameContext::kBoundingBox){
+        //Insert the bounding box filtering scene index which converts geometries into a bounding box using the extent attribute
+        _lastFilteringSceneIndexBeforeCustomFiltering = Fvp::BboxSceneIndex::New(_lastFilteringSceneIndexBeforeCustomFiltering);
```
ppt-adsk, line 998: Couldn't this always be in the chain and simply be a pass-through if bounding box mode isn't on?

ppt-adsk (commented):
Looks good to me, only minor comments, plus one question.  Great work!

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lanierd-adsk: Right, but this function is here as we will have some others cases in the near future, so I want to keep it generic.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lanierd-adsk: There is 2 possibilitites here.
1) We always have a BBox scene index with a flag to enable/disable it. 
2) we remove/add the BBox scene index when needed so we recreate the chain. 
In the near future, we'll have some more cases like override the scene with the default material which will likely be another scene index. 
I think it's better to completely remove the useless scene indices (so using solution 2) from above) instead of keeping them in the chain and disabling them as it will take some indirection for nothing.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lanierd-adsk: Please see my comment above.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
ppt-adsk: Makes sense.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
ppt-adsk: O.K., that's a valid choice.  Can we add this as an architectural-level code comment in the code, and possibly in a Markdown file somewhere, that this is what we expect in terms of architecture?

ppt-adsk (approved):

lilike-adsk (approved):

PR #104, opened 2024-03-19
Fix code coverage build issue

roopavr-adsk:
Change in this PR:

1. previous code in renderOverride.cpp has a unique_ptr's method function release(). Since our codebase cleanup, we changed the type of MayaHydraSceneIndex to Pixar's internal counted ref pointer. The equivalent method seems to be to use Reset()
2. Remove a spurious export/import on a member function which seems to be a relic and an uncaught bug.
3. Exclude PDB generation for a example code.


ppt-adsk (commented):
LGTM.

PR #106, opened 2024-03-21
HYDRA-927 - Expose complexity settings in globalRenderSettings

vlasovi:
Implement global refinement level for USD primitives
Automated test for the feature


vlasovi:

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -622,6 +622,10 @@ MStatus MtohRenderOverride::Render(
         _mayaHydraSceneIndex->PreFrame(drawContext);
     }
 
+    if (_displayStyleSceneIndex) {
```
lanierd-adsk, line 977: I think you should wait for my PR to be merged (it's ready to be merged) as there are functions to handle different cases for display styles and you could put the creation of this scene index in there.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -945,6 +949,10 @@ void MtohRenderOverride::_CreateSceneIndicesChainAfterMergingSceneIndex()
         _sceneIndexRegistry.reset(new MayaHydraSceneIndexRegistry(_renderIndexProxy));
     }
 
+    // Add display style scene index
```
lanierd-adsk, line 1308: Yes I do that in a function now, you should use it. My PR is https://github.com/Autodesk/maya-hydra/pull/103

lib/mayaHydra/hydraExtensions/mayaHydraParams.h
```diff
@@ -39,6 +39,7 @@ struct MayaHydraParams
     float motionSampleStart = 0;
     float motionSampleEnd = 0;
     bool  displaySmoothMeshes = true;
+    int   refineLevel = 0;
```
lanierd-adsk, line 42: What about the refineLevel which is adapters, MayaHydraAdapter::GetDisplayStyle().refineLevel > 0)
Is what you do replaces this ? Or is this something else ?

lilike-adsk: I assume the added one is the global refine level, while the one in DisplayStyle is per-primitive, but it's not enabled yet due to MRenderItem was already refined. HdsiLegacyDisplayStyleOverrideSceneIndex can be used to override the per-primitive one.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -945,6 +949,10 @@ void MtohRenderOverride::_CreateSceneIndicesChainAfterMergingSceneIndex()
...
+    _lastFilteringSceneIndexBeforeCustomFiltering = _displayStyleSceneIndex =
+            HdsiLegacyDisplayStyleOverrideSceneIndex::New(_lastFilteringSceneIndexBeforeCustomFiltering);
```
lilike-adsk: Maybe worth checking the perf impact (switch to Hydra/First Frame) after introducing this SceneIndex. We expect, by default, with refine level is 0, the overhead should be trivial

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lanierd-adsk: The Pr has been merged, please rebase your PR on the latest code and have a look at the function :
NeedToRecreateTheSceneIndicesChain and _CreateSceneIndicesChainAfterMergingSceneIndex you could put your code in both.

lib/mayaHydra/hydraExtensions/mayaHydraParams.h
vlasovi: As far as I understand MayaHydraAdapter is used only for Maya native prims. The new scene index is supposed to be used only for USD prims. I will add "addExcludedSceneRoot" to it in my next commit to exclude Maya prims

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1299,13 +1304,20 @@ void MtohRenderOverride::_CreateSceneIndicesChainAfterMergingSceneIndex(const MH
...
+            Fvp::DisplayStyleOverrideSceneIndex::New(_lastFilteringSceneIndexBeforeCustomFiltering);
+    _displayStyleSceneIndex->addExcludedSceneRoot(mayaNativeRoot); // Maya native prims don't use global refinement
+
     auto wfSi = TfDynamic_cast<Fvp::WireframeSelectionHighlightSceneIndexRefPtr>(Fvp::WireframeSelectionHighlightSceneIndex::New(_lastFilteringSceneIndexBeforeCustomFiltering, _selection));
     wfSi->SetDisplayName("Flow Viewport Wireframe Selection Highlight Scene Index");
     
     // At time of writing, wireframe selection highlighting of Maya native data
     // is done by Maya at render item creation time, so avoid double wireframe
     // selection highlighting.
-    wfSi->addExcludedSceneRoot(_ID);
```
vlasovi, line 1308: @ppt-adsk while debugging around I found that the call that you implemented to wfSi->addExcludedSceneRoot doesn't receive the proper path. In my case the exclude path was "/MayaHydraViewportRenderer/_MayaHydra_HdStormRendererPlugin_0000023A87F10B80" while all the Maya native primitives were at the path "/MayaHydraViewportRenderer/rprims" so essentially we were drawing the selection highlight two times for those. Here I fix the bug

lib/mayaHydra/mayaPlugin/renderOverride.cpp
vlasovi: @lilike-adsk With refine level set to 0, I haven't noticed any performance impact of this change. The FPS maybe degraded with refine levels more than 0, but this is expected.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
ppt-adsk: @vlasovi thanks for the fix!  Can we better document what _ID is supposed to be used for?  Also, `/MayaHydraViewportRenderer` is duplicated at line 566.  Can we factor this out?

lanierd-adsk (dismissed):

ppt-adsk (approved):

PR #108, opened 2024-03-25
HYDRA-861 & HYDRA-932 : Fix MtoA & LookdevX loading in tests + Related test improvements

debloip-adsk:
This PR contains the following changes : 
- Fixed MtoA and LookdevX loading
- Reworked plugin loading in tests to make it so failing to load a plugin fails the test instead of skipping it. This is to avoid the MtoA/LookdevX situation where the tests weren't working yet reported as passing.
- Merged `MtohTestCase` into `MayaHydraBaseTestCase`, as the distinction was no longer really useful and was leading to code duplication
- Converted absolute paths in test scenes to relative paths
- Converted .usd files to .usda for readability


test/lib/mayaUsd/render/mayaToHydra/README.md
```diff
@@ -26,7 +26,7 @@ For example, to run testVisibility on RelWithDebInfo :
...
+The tests require MayaUSD in order to be run. In order for the test framework to find and load MayaUSD, the `-DMAYAUSD_LOCATION` CMake flag must be set to your MayaUSD installation directory when building MayaHydra. For convenience, you can use the `--mayausd-location` parameter when using the [build.py](../../../../../build.py) script, which will set the CMake flag for you with the given argument. The specified path should point to the directory where your MayaUSD `.mod` file resides. Note that the test framework will not find MayaUSD using the MAYA_MODULE_PATH environment variable.
```
debloip-adsk, line 29: (see [`mtohUtils.py`](https://github.com/Autodesk/maya-hydra/pull/108/files#r1538235005) for the corresponding code changes) I made it so we load MayaUSD by default for all tests, since we expect MayaHydra to be built with it anyways.

test/lib/mayaUsd/render/mayaToHydra/README.md
```diff
@@ -54,29 +54,29 @@ Looking to add a C++ test suite? See the corresponding [README.md](./cpp/README.
 To add a new Python-only test suite : 
 
 1. Create a new test[...].py file in the current folder.
-2. Create a test class that derives from unittest.TestCase (doesn't need to directly inherit from it).
-3. Add `test_[...]` methods for each test case in your test suite.
-4. Add the following snippet at the bottom of your file :
+2. Create a test class that derives from `mtohUtils.MayaHydraBaseTestCase`.
+3. Add the line `_file = __file__` in your class definition.
+4. If needed, set `_extraPluginsToLoad` to list any plugins that need to be loaded for your test. MayaUSD does not need to be specified, it will be loaded automatically.
```
debloip-adsk: (see [`mtohUtils.py`](https://github.com/Autodesk/maya-hydra/pull/108/files#r1538235005) for the corresponding code changes) The `_extraPluginsToLoad` variable will be used to list any plugins required by the test suite. As opposed to the previous method of skipping a test if it failed to load, this method will try to load the listed plugins, but fail the test suite immediately if one of them fails to load.

test/lib/mayaUsd/render/mayaToHydra/README.md
```diff
@@ -54,29 +54,32 @@ Looking to add a C++ test suite? See the corresponding [README.md](./cpp/README.
...
     if __name__ == '__main__':
         fixturesUtils.runTests(globals())
     ```
-5. Add the name of your test[...].py file to the list of `TEST_SCRIPT_FILES` in the [current folder's CMakeLists.txt](./CMakeLists.txt)
+7. Add the name of your test[...].py file to the list of `TEST_SCRIPT_FILES` in the [current folder's CMakeLists.txt](./CMakeLists.txt)
 
 Some important notes :
-- By default, the mayaHydra plugin will not be loaded. An easy way to load it is to have your test class derive from `MayaHydraBaseTestCase`, which will automatically load mayaHydra when running tests. Don't forget to add the line `_file = __file__` in your class.
-- After loading mayaHydra, the renderer will still be on Viewport 2.0. If you want to use another renderer, your test will have to switch to it. If you want to use HdStorm, an easy way to do it is to have your test class derive from `MayaHydraBaseTestCase` and call `self.setHdStormRenderer()`
+- Before each test, a new file will be opened and the renderer will be switched to Hydra. If you need to switch between renderers, you can use the `self.setHdStormRenderer()` and `self.setViewport2Renderer()` methods.
```
debloip-adsk, line 69: Removed most of the previous info as it was outdated, and removed the mention of "MayaHydra not being loaded by default unless deriving from `MayaHydraBaseTestCase`", since rarely (if ever) a test will want to derive from `unittest.TestCase` directly. I've also adjusted the steps to reflect that.

test/lib/mayaUsd/render/mayaToHydra/README.md
```diff
@@ -54,29 +54,32 @@ Looking to add a C++ test suite? See the corresponding [README.md](./cpp/README.
...
 
 Image comparison is done using [idiff from OpenImageIO](https://openimageio.readthedocs.io/en/latest/idiff.html).
 
 Some utilities exist to facilitate image comparison tests. Here is the simplest way to get going : 
-1. Make your test class derive from `MtohTestCase`.
-2. Don't forget to add the line `_file = __file__` in your class.
-3. Create a folder called [TestName]Test (e.g. for testVisibility, create a folder VisibilityTest)
-4. Put the images you want to use for comparison in that folder
-5. Call `self.assertImagesClose`, `self.assertImagesEqual` and/or `self.assertSnapshotClose` with the file names of the images to compare with.
+1. Make sure your test class derives from `mtohUtils.MayaHydraBaseTestCase` (and has the line `_file = __file__`).
```
debloip-adsk, line 81: See [`mtohUtils.py`](https://github.com/Autodesk/maya-hydra/pull/108/files#r1538231177) for the corresponding code changes and explanation of merging `MtohTestCase` into `MayaHydraBaseTestCase`

test/lib/mayaUsd/render/mayaToHydra/README.md
```diff
@@ -88,10 +91,4 @@ To create a reference snapshot, an easy way is to first write your test, and the
...
-    ```@unittest.skipUnless(mtohUtils.checkForMayaUsdPlugin(), "Requires Maya USD Plugin.")```
-    - `MayaHydraBaseTestCase` is a base class you can use for your test classes that will provide you with some useful functionality, such as automatically loading the mayaHydra plugin. It also provides a method to use the HdStorm renderer (`setHdStormRenderer`), and another to create a simple scene with a cube (`makeCubeScene`).
-    - `MtohTestCase` is a class you can use as a base for your test classes that will provide you with image comparison functionality (see [Image comparison](#Image-comparison) section). Note that this class already derives from MayaHydraBaseTestCase, so it will also automatically load the mayaHydra plugin.
+Utility files are located under [/test/testUtils](../../../../testUtils/). Note that at the time of writing this (2023/08/16), many of the utils files and their contents were inherited from the USD plugin, and are not all used.
```
debloip-adsk, line 96: Removed most info as it is outdated or not very relevant. The plugin loading info has been moved into the steps to add tests + the best practices section

test/testUtils/mtohUtils.py
debloip-adsk, line 1: Most differences here are merging the `MtohTestCase`-specific code into `MayaHydraBaseTestCase`. Historically, the `MayaHydraBaseTestCase` class was a read-only class for tests that didn't output anything, and `MtohTestCase` was the image-diffing subclass that set up an output directory. However, since code coverage was added, `MayaHydraBaseTestCase` now also sets up an output directory, which is kind of duplicated from `MtohTestCase`. The plugin loading rework code would also have needed to be duplicated in both `setupClass` methods of `MayaHydraBaseTestCase` and `MtohTestCase`, so it felt simpler to just merge the two. Additionally, the data-source dumping setup that should be coming soon will also need pretty much the same setup as `MtohTestCase`, so I feel like there is not much point to keeping separate classes, and just merging them into one will be easier to maintain.

test/testUtils/mtohUtils.py
```diff
@@ -36,27 +39,44 @@ def checkForPlugin(pluginName: str):
...
 
-class MayaHydraBaseTestCase(unittest.TestCase):
-    '''Base class for mayaHydra unit tests without image comparison.'''
+    _inputDir = None
 
+    # Variables to be set in subclasses
     _file = None
-
-    DEFAULT_CAM_DIST = 24
+    _extraPluginsToLoad = []
```
debloip-adsk: New variable to list what plugins a test suite needs

test/testUtils/mtohUtils.py
```diff
@@ -36,27 +39,44 @@ def checkForPlugin(pluginName: str):
...
+        cls._testDir = os.path.abspath('.')
+
+        if MAYAUSD_PLUGIN_NAME not in cls._extraPluginsToLoad:
+            cls._extraPluginsToLoad.append(MAYAUSD_PLUGIN_NAME)
+
+        for pluginToLoad in cls._extraPluginsToLoad:
+            # If a plugin fails to load, the entire test suite will be immediately aborted.
+            # Note that in the case of mtoa, the plugin might load successfully but not
+            # initialize properly, which means issues will only be caught in the actual tests.
+            cmds.loadPlugin(pluginToLoad)
```
debloip-adsk: Plugin-loading code that will fail the test suite if one fails to load (with one asterisk, see the comment). We add MayaUSD to the list if it is not specified, since we expect MayaHydra to have been built with it. This allows most tests to not have to define `_extraPluginsToLoad` if all they need is MayaUSD, which will be quite common.

test/testSamples/testMayaDisplayModes/testMayaDisplayModes.ma
debloip-adsk, line 1: Removed Arnold references as per HYDRA-932. Also removed the USD stage references as there is no associated stage.

debloip-adsk:
FYI @ppt-adsk @vlasovi @lanierd-adsk @roopavr-adsk

test/lib/mayaUsd/render/mayaToHydra/README.md
```diff
@@ -54,29 +54,34 @@ Looking to add a C++ test suite? See the corresponding [README.md](./cpp/README.
...
 
 Image comparison is done using [idiff from OpenImageIO](https://openimageio.readthedocs.io/en/latest/idiff.html).
 
 Some utilities exist to facilitate image comparison tests. Here is the simplest way to get going : 
-1. Make your test class derive from `MtohTestCase`.
-2. Don't forget to add the line `_file = __file__` in your class.
-3. Create a folder called [TestName]Test (e.g. for testVisibility, create a folder VisibilityTest)
-4. Put the images you want to use for comparison in that folder
-5. Call `self.assertImagesClose`, `self.assertImagesEqual` and/or `self.assertSnapshotClose` with the file names of the images to compare with.
+1. Make sure your test class derives from `mtohUtils.MayaHydraBaseTestCase` (and has the line `_file = __file__`).
```
lilike-adsk, line 81: side question: is it possible to avoid the copying of "_file = __file__" each time?

test/testUtils/mtohUtils.py
```diff
@@ -36,27 +39,44 @@ def checkForPlugin(pluginName: str):
...
+            inputDirName = os.path.splitext(os.path.basename(cls._file))[0]
+            inputDirName = testUtils.stripPrefix(inputDirName, 'test')
+            if not inputDirName.endswith('Test'):
+                inputDirName += 'Test'
+            cls._inputDir = os.path.join(inputPath, inputDirName)
+
+        cls._testDir = os.path.abspath('.')
+
+        if MAYAUSD_PLUGIN_NAME not in cls._extraPluginsToLoad:
+            cls._extraPluginsToLoad.append(MAYAUSD_PLUGIN_NAME)
```
lilike-adsk: Is it better to set _extraPluginsToLoad = [MAYAUSD_PLUGIN_NAME] in MayaHydraBaseTestCase, so the derived test case has an opportunity to not load mayaUSDPlugin if needed?

test/testUtils/mtohUtils.py
```diff
@@ -36,27 +39,44 @@ def checkForPlugin(pluginName: str):
...
+        if cls._inputDir is None:
+            inputDirName = os.path.splitext(os.path.basename(cls._file))[0]
+            inputDirName = testUtils.stripPrefix(inputDirName, 'test')
+            if not inputDirName.endswith('Test'):
+                inputDirName += 'Test'
+            cls._inputDir = os.path.join(inputPath, inputDirName)
+
+        cls._testDir = os.path.abspath('.')
+
+        if MAYAUSD_PLUGIN_NAME not in cls._extraPluginsToLoad:
```
ppt-adsk: Nit-pick: in Maya terminology plugins that are needed for a scene are described with the `requires` command:
https://help.autodesk.com/cloudhelp/2024/ENU/Maya-Tech-Docs/Commands/requires.html
Maybe for familiarity we could call this list `_requiredPlugins`?  Just a suggestion, I'm fine with keeping what you have as well.

ppt-adsk (commented):
Great work, thanks for the cleanup!

test/lib/mayaUsd/render/mayaToHydra/README.md
debloip-adsk: So, this is used in order to set the `_inputDir` variable, which points to the input folder containing the reference images (or dumped data sources), by using the given file path and test class name. Since the input folders live in the same hierarchy as the test files, `__file__` is used to get the path to test file, which can then be used to construct the path to its corresponding input folder. I've tried other means of obtaining the path to the test file/class from the generic test setup code, such as by using [`inspect.getfile()`](https://docs.python.org/3/library/inspect.html#inspect.getfile), but it does not seem to work. With the relative paths between the test setup code and test files being known, we could in theory use these to construct the path to the input folders, but that would need to be changed along with any directory structure change, and we'd also need to account for C++ tests being in a subdirectory. At this point I think it'd be more trouble than it's worth, so I suggest staying with what we have.

test/testUtils/mtohUtils.py
debloip-adsk: Following our discussion in the scrum, I suggest leaving it as is currently, and adjust later if needed.

test/testUtils/mtohUtils.py
debloip-adsk: Adjusted the naming, thanks!

lilike-adsk (dismissed):
Looks good!

ppt-adsk (dismissed):

ppt-adsk (approved):

lilike-adsk (approved):

PR #110, opened 2024-03-26
HYDRA-837 : Fix color management handling in tests

debloip-adsk:
This PR fixes the color management handling in tests and updates reference images accordingly.


test/testUtils/mayaUtils.py
```diff
@@ -135,6 +135,9 @@ def isHydraRenderer():
 def openTestScene(*args):
     filePath = testUtils.getTestScene(*args)
     cmds.file(filePath, force=True, open=True)
+    # To err on the safe side, disable color management in case the scene had color management on.
+    # If/when we add color management tests, they will need to re-enable it manually.
+    cmds.colorManagementPrefs(edit=True, cmEnabled=False)
```
debloip-adsk, line 140: After starting work on adding tests for the grid and background color, there are a couple things that happen in the `snapshot` method that I think we could move to instead happen after each scene opening/creation, for example here disabling color management, but also things like disabling the grid, setting a default background color, etc. 

This'll keep the current behavior in the usual cases, but make it easier to change these settings should we need to (which we will for testing them). It'd also make the snapshot method be more of an actual snapshot of the current viewport, rather than it modifying a bunch of things unknowingly before actually taking the snapshot. 

Currently some of these settings are passed in as parameters to the `snapshot` function, but it doesn't really feel to me like that's where they should be, and I feel it's a bit awkward. To make sure those "post-scene opening" settings get applied consistently, I was thinking of maybe adding a script job or subscribing to the `MSceneMessage`s. 

That being said this is all for a future PR : for this current PR, I think this will suffice. I would prefer getting the color management setup functional first as it is going to be in contention with other PRs that may arise, and then refactor afterwards.

test/lib/mayaUsd/render/mayaToHydra/MayaDisplayLayersTest/coneLayerCustomWireframe.png
ppt-adsk, line 1: Not from this pull request, but I'm concerned about how little content there is in some of these reference images.  The image itself is already small (400x400, which is fair for a unit test), but the area in which there are meaningful pixels is tiny (on the order of 50x80 pixels, or 2.5% of the image area).  Is this intentional?  How can we be sure that we're covering problem cases with these reference images?

test/testUtils/mayaUtils.py
ppt-adsk: Isn't this doing a bit too much?  What if I really do want to have a particular color management setup in my file?  I'm also slightly uncomfortable with unconditionally setting multi-sampling.  I would expect that openTestScene() would open my file, not mess with it...

Perhaps we need a new function which does this, openTestSceneApplyTestSettings(), or an optional argument to openTestScene(), applyTestSettings=False?

debloip-adsk: Yeah there's no silver bullet here, so here are a couple thoughts :
1. I agree that `openTestScene()` would ideally just open a test scene. But in the same spirit, `snapshot` should also just take a snapshot, and not mess with the settings, which is what it currently does. However, the risk here is that if we make both of these methods not change anything in the scene, then we depend on test scenes being saved with the right settings, which while is the theoretically correct way to do things, seems to me like it would make it easy to miss if some settings weren't actually set properly. If we go this route, we'll need to make sure we pay very close attention to the scene files.
2. Following that, my idea was that we'd make `snapshot` not change the settings, and on each file open, we'd set those default settings, and if a test requires different values, it would then re-set them to the values it wants. For color management specifically, we're not changing the entire setup, we're just toggling it, so a test would just have to re-enable it. Also, even without this specific instance of disabling color management, currently it'll still be disabled by default in the `snapshot` method, which means you still wouldn't be able to test the color management setup, unless we add a color management status parameter to `assertSnapshotClose` (and if we add a parameter for color management, we should add one for the other settings like hud, grid, etc., which feels a bit misplaced and part of why I want to refactor this).
3. Allowing tests to opt-out of those default settings sounds like a good compromise if we ever need a specific scene setup (though I probably wouldn't have it be a parameter in `mayaUtils.openTestScene()`, since I think they should be applied on new files as well).
4. Yes I also think MSAA should be off as a default. Possibly it was set to on because Storm had it enabled by default, and our existing tests all used it?

debloip-adsk:
Note : the failures on OSX are likely aliasing issues due to reverting to the much smaller thresholds from before (the failures are with smaller differences than before), as OSX seems to have slightly different aliasing, maybe because of the Metal backend? Anyways, looking into it. If it is aliasing, then that would be a good point to the fact that MSAA should likely be disabled by default, though we can tackle that separately.

test/testUtils/mayaUtils.py
ppt-adsk: I like (1) best, it's slightly riskier, but I'd rather have people be deliberate about their settings, and paying attention to your data is good, but this is open to a team discussion.  For the moment I wouldn't change anything and leave this pull request as is.

ppt-adsk (dismissed):

test/lib/mayaUsd/render/mayaToHydra/MayaDisplayLayersTest/coneLayerCustomWireframe.png
debloip-adsk: I agree, this likely just has to do with the default distance of the camera being a bit too far in some cases. In most cases we can probably use `self.setBasicCam` to place the camera closer to have objects take up more space. I've logged HYDRA-944 to keep track of reviewing such images and updating if needed.

test/lib/mayaUsd/render/mayaToHydra/testMayaLights.py
```diff
@@ -26,7 +26,14 @@ class TestMayaLights(mtohUtils.MayaHydraBaseTestCase): #Subclassing mtohUtils.Ma
...
     IMAGE_DIFF_FAIL_THRESHOLD = 0.01
-    IMAGE_DIFF_FAIL_PERCENT = 0.2
+
+    @property
+    def IMAGE_DIFF_FAIL_PERCENT(self):
+        # HYDRA-943 : Shadows don't seem to work properly on OSX, possibly only with point lights.
+        # Only the "all lights with shadows" check fails, and shadows with point light only is skipped.
+        if platform.system() == "Darwin":
+            return 3
+        return 0.2
```
debloip-adsk: Unfortunately had to bring back the workaround in this case, as the higher threshold we used hid another issue. This has likely been happening for a while, unsure if this is only a test issue or if it is also an issue when using MayaHydra normally.

debloip-adsk:
Updated the thresholds, and I'm not even sure if it was actually due to aliasing differences. I honestly did not see any obvious visual differences in the images, even when zooming in. The thresholds were probably just too small to account for the different Mac GPU, I've increased them very slightly to make the tests pass.

ppt-adsk (dismissed):

ppt-adsk (approved):

test/testUtils/mayaUtils.py
lilike-adsk: I like Pierre's idea to be more flexible, probably we can have an applySettings() that can be used in OpenScene() and NewScene()?
Is it possible to have a default settings (E.g., MSAA=ON, CM=OFF, etc.) and a skip flag in our base test class or util?
So in the concrete tests, we can override the default settings (E.g., MSAA=OFF, CM=ON) Or enable skip flag to use scene settings directly if you want?

PR #112, opened 2024-03-28
HYDRA-536 & HYDRA-537 : Background color and grid tests

debloip-adsk:


test/testUtils/mtohUtils.py
```diff
@@ -88,7 +88,7 @@ def setUp(self):
         # Maya is not closed/reset between each test of a test suite,
         # so open a new file before each test to minimize leftovers
         # from previous tests.
-        cmds.file(new=True, force=True)
+        mayaUtils.openNewScene()
```
debloip-adsk, line 91: In the [discussion on the color management PR](https://github.com/Autodesk/maya-hydra/pull/110/files#r1541134556), we talked about relying on the saved scene settings; however some settings are application-level and not saved per-scene, and some tests don't load a scene but instead construct it programmatically. So to remedy this, the suggestion of an optional parameter seems to me like the simplest compromise. I also added an analogous `mayaUtils.openNewScene` function, which opens a new scene and applies the test settings, and which is called before each test in replacement of `cmds.file(new...)` here. This, way the test settings are applied before each test and each new empty scene opened, but if we load an existing scene, we can let the scene's settings take over.

Note that in the spirit of favoring saved scene settings, we could consider removing the option of applying test settings in `mayaUtils.openTestScene`, and always relying on the saved scene settings, however for that we'd need to do a pass on our scene files to update them with the correct settings, which I feel would be better suited for a different PR (potentially alongside making MSAA disabled by default).

test/lib/mayaUsd/render/mayaToHydra/testBackgroundColor.py
```diff
@@ -0,0 +1,49 @@
...
+        self.compareBackgroundColor("background_green.jpg")
+        cmds.displayRGBColor('background', 0, 0, 0.5)
+        self.compareBackgroundColor("background_blue.jpg")
+        
+        cmds.displayRGBColor('background', 0, 0, 0)
+        self.compareBackgroundColor("background_black.jpg")
+        cmds.displayRGBColor('background', 0.5, 0.5, 0.5)
+        self.compareBackgroundColor("background_gray.jpg")
+        cmds.displayRGBColor('background', 1, 1, 1)
+        self.compareBackgroundColor("background_white.jpg")
```
debloip-adsk, line 46: It seems like we must use .jpg for playblasts to render the background color, using .png always rendered the background as transparent.

test/lib/mayaUsd/render/mayaToHydra/testMtohBasicRender.py
debloip-adsk, line 1: This test hasn't been enabled in over a year; see [the test CMakeLists](https://github.com/Autodesk/maya-hydra/blob/63334cf1d91c140f0fc1b598872d05a2b3a276b9/test/lib/mayaUsd/render/mayaToHydra/CMakeLists.txt#L5-L7). I updated it on principle in case we bring it back and based on how I think it should be updated, but I haven't actually run it.

test/lib/mayaUsd/render/mayaToHydra/testMtohDagChanges.py
debloip-adsk, line 1: This test hasn't been enabled in over a year; see [the test CMakeLists](https://github.com/Autodesk/maya-hydra/blob/63334cf1d91c140f0fc1b598872d05a2b3a276b9/test/lib/mayaUsd/render/mayaToHydra/CMakeLists.txt#L5-L7). I updated it on principle in case we bring it back and based on how I think it should be updated, but I haven't actually run it.

test/testUtils/imageUtils.py
debloip-adsk, line 1: Removed a bunch of settings changes that were happening under the hood and moved them to an optional step when opening scenes (except `camera` as that one was never used). I kept `defaultRenderGlobals.imageFormat` because this is set based on the extension of the file name being passed in.

test/testUtils/mayaUtils.py
debloip-adsk, line 1: Removed some unused methods, added `openNewScene`, and added `applyTestSettings`, which mostly applies the test settings that were previously in `snapshot`.

test/lib/mayaUsd/render/mayaToHydra/testMtohBasicRender.py
```diff
@@ -26,8 +26,6 @@ class TestSnapshot(mtohUtils.MayaHydraBaseTestCase):
     _file = __file__
 
     def test_flat_orange(self):
-        cmds.file(new=1, f=1)
-
```
debloip-adsk, line 30: This test hasn't been enabled in over a year; see [the test CMakeLists](https://github.com/Autodesk/maya-hydra/blob/63334cf1d91c140f0fc1b598872d05a2b3a276b9/test/lib/mayaUsd/render/mayaToHydra/CMakeLists.txt#L5-L7). I updated it on principle in case we bring it back and based on how I think it should be updated, but I haven't actually run it.

test/lib/mayaUsd/render/mayaToHydra/testMtohDagChanges.py
```diff
@@ -271,7 +272,7 @@ def test_node_creation_undo(self):
 
         cmds.undoInfo(state=0)
         try:
-            cmds.file(new=1, f=1)
+            mayaUtils.openNewScene()
```
debloip-adsk, line 275: This test hasn't been enabled in over a year; see [the test CMakeLists](https://github.com/Autodesk/maya-hydra/blob/63334cf1d91c140f0fc1b598872d05a2b3a276b9/test/lib/mayaUsd/render/mayaToHydra/CMakeLists.txt#L5-L7). I updated it on principle in case we bring it back and based on how I think it should be updated, but I haven't actually run it.

test/testUtils/imageUtils.py
```diff
@@ -15,8 +15,6 @@
 #
```
debloip-adsk, line 15: Removed a bunch of settings changes that were happening under the hood and moved them to an optional step when opening scenes (except `camera` as that one was never used). I kept `defaultRenderGlobals.imageFormat` because this is set based on the extension of the file name being passed in.

test/testUtils/mayaUtils.py
```diff
@@ -29,6 +29,7 @@
     pass
```
debloip-adsk, line 30: Removed some unused methods, added `openNewScene`, and added `applyTestSettings`, which mostly applies the test settings that were previously in `snapshot`.

test/testUtils/mtohUtils.py
debloip-adsk: Also with regards to @lilike-adsk's comment on providing a way to override certain default settings or using a skip flag, I think a simpler way is that if a test really needs specific settings, it can override `setUp` method and set whatever settings it needs, which means we don't have to implement a specific mechanism for this.

debloip-adsk:
FYI @lilike-adsk @lanierd-adsk @roopavr-adsk @vlasovi

ppt-adsk (approved):

PR #113, opened 2024-03-29
HYDRA-934 : Fix reload of a scene with a Usd stage.

lanierd-adsk:


lanierd-adsk:
FYI : @debloip-adsk @lilike-adsk @vlasovi @roopavr-adsk

lanierd-adsk:
@ppt-adsk  : unfortunately, I could not find a way to repro the problem with a unit test, the test always passes with image comparison (a C++ would be the same as the prims are present and visible).

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1218,6 +1218,10 @@ void MtohRenderOverride::_InitHydraResources(const MHWRender::MDrawContext& draw
     //Add the scene index as an input scene index of the merging scene index
     _renderIndexProxy->InsertSceneIndex(_mayaHydraSceneIndex, SdfPath::AbsoluteRootPath());
     
+    if (!_sceneIndexRegistry) {
```
lanierd-adsk, line 1221: This change is not really related to the bug but I think it's better to pull out this from the _CreateSceneIndicesChain function as it will be ignored most of the time.

ppt-adsk (approved):

PR #115, opened 2024-04-03
USD pick with kind support, with test.

ppt-adsk:
USD defines a prim-level metadatum called Kind, to establish a model hierarchy, for organization and ease of navigation.
https://openusd.org/release/glossary.html#usdglossary-kind
This pull request implements Kind support on pick.


lib/mayaHydra/mayaPlugin/renderOverride.cpp
ppt-adsk, line 1: I realize that this file is becoming a picking kitchen sink.  Will think about extracting this in a later pull request.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -164,6 +169,42 @@ PXR_NAMESPACE_USING_DIRECTIVE
 
 static const SdfPath MAYA_NATIVE_ROOT = SdfPath("/MayaHydraViewportRenderer");
 
+//! \brief  Query the Kind to be selected from viewport.
+//! \return A Kind token (https://graphics.pixar.com/usd/docs/api/kind_page_front.html). If the
+//!         token is empty or non-existing in the hierarchy, the exact prim that gets picked
+//!         in the viewport will be selected.
+TfToken GetSelectionKind()
```
ppt-adsk, line 179: Copy pasted from
https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/render/vp2RenderDelegate/proxyRenderDelegate.cpp

test/lib/mayaUsd/render/mayaToHydra/CMakeLists.txt
```diff
@@ -47,8 +47,9 @@ set(INTERACTIVE_TEST_SCRIPT_FILES
     cpp/testSceneCorrectness.py
     cpp/testPrimInstancing.py
     cpp/testPicking.py
-    cpp/testPointInstancePicking.py
+    cpp/testUsdPointInstancePicking.py
```
ppt-adsk, line 50: Rename for consistency: these tests pick USD data.

test/lib/mayaUsd/render/mayaToHydra/cpp/CMakeLists.txt
```diff
@@ -35,8 +35,8 @@ target_sources(${TARGET_NAME}
         testSceneCorrectness.cpp
         testPrimInstancing.cpp
         testPicking.cpp
-        testPointInstancePicking.cpp
-        testUsdNativeInstancePicking.cpp
+        testUsdPointInstancePicking.cpp
+        testUsdPicking.cpp
```
ppt-adsk, line 39: The USD native instance picking C++ test code was not specific to native instancing, so renamed it for generality.

lib/mayaHydra/mayaPlugin/CMakeLists.txt
```diff
@@ -38,6 +38,7 @@ target_compile_definitions(${TARGET_NAME}
         $<$<BOOL:${IS_LINUX}>:LINUX>
         # Not sure if msvcc sets this automatically, but won't hurt to redefine
         $<$<BOOL:${IS_WINDOWS}>:_WIN32>
+        $<$<BOOL:${MayaUsd_FOUND}>:MAYAHYDRALIB_MAYAUSDAPI_ENABLED>
```
debloip-adsk, line 41: Question : The three added lines in this file are already present in the CMakeLists of mayaHydraLib/hydraExtensions, should we consider making them public there instead of duplicating them here, or do we want to keep it separate? (same goes for UFE, where we also repeat some include directories and link libraries across ufeExtensions, hydraExtensions and mayaPlugin)

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -287,6 +328,15 @@ HitPath pickInstancer(
     return {instancerPrimOrigin(primOrigin.instancerContexts.front()), -1};
 }
 
+Ufe::Path usdPathToUfePath(
+    const MayaHydraSceneIndexRegistrationPtr& registration,
+    const SdfPath&                            usdPath
+)
+{
+    return registration ? registration->interpretRprimPathFn(
```
debloip-adsk, line 339: `registration` is a reference here, so we shouldn't be checking for null

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -445,19 +495,45 @@ class UsdPickHandler : public MtohRenderOverride::PickHandlerBase {
...
+                return pickedMayaPath;
+            }
+
+            MayaUsdAPI::ProxyStage proxyStage{proxyShapeObj};
+            auto prim = proxyStage.getUsdStage()->GetPrimAtPath(pickedUsdPath);
+            prim = GetPrimOrAncestorWithKind(prim, snKind);
+            const auto usdPath = prim ? prim.GetPath() : pickedUsdPath;
+
+            return usdPathToUfePath(registration, usdPath);
+        }(pickedUsdPath);
```
debloip-adsk: Suggestion : I find the lambda nested in a tertiary a bit hard to read personally, I would propose moving the `(instanceNdx >= 0)` case inside the lambda as well and handle it with an early return like the other cases, so that it's a bit more consistent and `snMayaPath` is directly the result of the lambda.

debloip-adsk (changes requested):

lib/mayaHydra/mayaPlugin/renderOverride.cpp
debloip-adsk: Nevermind this, just saw that it's a reference to a pointer

debloip-adsk (dismissed):

lib/mayaHydra/mayaPlugin/CMakeLists.txt
ppt-adsk: I think it helps to keep the dependencies explicit, but I don't have a strong opinion here.

debloip-adsk (approved):

PR #116, opened 2024-04-08
HYDRA-955 : Pass normals and tangents to Hydra from render item and mesh adapters

lanierd-adsk:


lanierd-adsk:
FYI : @ppt-adsk @debloip-adsk @roopavr-adsk @vlasovi

lib/mayaHydra/hydraExtensions/adapters/meshAdapter.cpp
```diff
@@ -201,9 +250,25 @@ class MayaHydraMeshAdapter : public MayaHydraShapeAdapter
                 return {};
             }
             return GetPoints(mesh);
-        } else if (key == MayaHydraAdapterTokens->st) {
+        } 
+
+        if (key == HdTokens->normals) {
+            MStatus status;
```
lanierd-adsk: I am doing a change in this file to standardize the methods and avoid passing sometimes a MfnMesh and sometimes not.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -71,10 +71,13 @@ TF_DEFINE_ENV_SETTING(MAYA_HYDRA_USE_MESH_ADAPTER, false,
     "Use mesh adapter instead of MRenderItem for Maya meshes.");
 
 bool useMeshAdapter() {
-    static bool uma = TfGetEnvSetting(MAYA_HYDRA_USE_MESH_ADAPTER);
+    static const bool uma = TfGetEnvSetting(MAYA_HYDRA_USE_MESH_ADAPTER);
     return uma;
 }
 
+TF_DEFINE_ENV_SETTING(MAYA_HYDRA_PASS_NORMALS_TO_HYDRA, true,
```
lilike-adsk, line 78: With passing normal as true by default, is there any test case to be rebaselined? 
also, it's better to have a test case for normal if there's no one.

lib/mayaHydra/hydraExtensions/adapters/renderItemAdapter.cpp
```diff
@@ -258,42 +285,59 @@ void MayaHydraRenderItemAdapter::UpdateFromDelta(const UpdateFromDeltaData& data
...
+                                _uvs.clear();
+                                _uvs.resize(indices->size());
+                                float* uvs = (float*)mvb->map();
+                                for (int i = 0; i < indexCount; ++i){
+                                    _uvs[i].Set(&uvs[indicesData[i] * 2]);
+                                }
+                                mvb->unmap();
+                            }
+                            break;
+                            case MHWRender::MGeometry::kTangent:{
```
lilike-adsk, line 332: Didn't see notion for tangent type in Hydra, is mikkt tangent not supported in Hydra, right?
Also, it's good to have a test case to cover tangent.

lanierd-adsk: Yes only Mikkt is supported for tangents, but it's not explicitly mentioned in Hydra.
I left the code for tangents for a later PR but we need to modify the translation of materials to support tangents. At this time the tangents won't be used so I cannot add a test for them.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
lanierd-adsk: With the new normals, all the tests should pass without being rebaselined. I haven't added any tests since I think it will be difficult to differentiate between the generated normals and the passed normals, there are only a few details which are different.

ppt-adsk: Is there no way to generate a small synthetic scene (e.g. a mesh cube maybe?) with passed normals that would be completely different from generated normals (e.g. the opposite sign), and look at the normal values in the Hydra scene index scene, i.e. without image comparison?  Don't know if this is valid / realistic, just thinking out loud.

lanierd-adsk: You're right, I can use some scene from the bugs we have when not using the normals from OGS. 
I'll add this as a test. Thanks.

ppt-adsk (dismissed):

lilike-adsk (approved):

ppt-adsk (approved):

PR #117, opened 2024-04-16
HYDRA-826 : Global Default Material Override for Maya data

roopavr-adsk:
Added feature to support "Use Default Material" Maya viewport Shading mode option. The override is only for native Maya data.


test/testSamples/testMayaDisplayModes/testMayaDisplayModes.ma
```diff
@@ -1,19 +1,19 @@
 //Maya ASCII 2026ff01 scene
 //Name: testMayaDisplayModes.ma
-//Last modified: Wed, Mar 06, 2024 12:10:27 PM
+//Last modified: Tue, Apr 16, 2024 02:39:26 PM
```
lilike-adsk, line 3: why the scene needs an update?

test/testSamples/testMayaDisplayModes/testMayaDisplayModes.ma
```diff
@@ -236,14 +233,16 @@ createNode materialInfo -n "materialInfo1";
 	rename -uid "7A353D3A-4CE0-1AA9-B650-97B907DB1314";
 createNode file -n "file2";
 	rename -uid "024402E0-4F08-CE4B-8A38-2194E8643C1A";
-	setAttr ".ftn" -type "string" "./diffuse.png";
+	setAttr ".ftn" -type "string" "D:/repos/mayaHydra/ecg-maya-hydra/maya-hydra/test/testSamples/testMayaDisplayModes/diffuse.png";
```
lilike-adsk: This path needs to be relative.

roopavr-adsk: good catch.

test/testSamples/testMayaDisplayModes/testMayaDisplayModes.ma
roopavr-adsk: The prior ref images were too dark. This is because our test framework resets the default light intensity to 1 (from ~3) and that code also turns off Color Management (inside [applyTestSettings()](https://github.com/Autodesk/maya-hydra/blob/dev/test/testUtils/mayaUtils.py#L144)). To use default light intensity( for brighter ref image), the test scene is opened by not calling the utility fn like so: 
testFile = mayaUtils.openTestScene(
                "testMayaDisplayModes",
                "testMayaDisplayModes.ma", useTestSettings=False)
So the test scene was now modified where CM setting is manually turned off in the file.  Note that the CM needs to be off in our test for various reasons.

lilike-adsk (approved):

PR #118, opened 2024-04-18
HYDRA-310 : Support wireframe and wire on shaded and support selection highlight with the right colors

lanierd-adsk:
…


lanierd-adsk:
FYI : @lilike-adsk @debloip-adsk @roopavr-adsk @vlasovi

lib/flowViewport/selection/fvpSelection.h
```diff
@@ -83,6 +87,20 @@ class Selection
     PXR_NS::HdDataSourceBaseHandle
     GetVectorDataSource(const PXR_NS::SdfPath& primPath) const;
 
+    //Get the wireframe color of a primitive for selection highlighting, 
+    // this checks if the prim is selected or not and if it is selected, 
+    // it returns the lead or active color depending on the lead of the selection
+    FVP_API
+    PXR_NS::GfVec4f GetWireframeColor(const PXR_NS::SdfPath& primPath)const;
```
ppt-adsk: I'm uncomfortable with this interface.  I would argue for separation of concerns and keep the Fvp::Selection solely responsible for maintaining the Hydra selection.  It shouldn't know whether the selection is to be represented in wireframe, or any other way.

If this is the only way we currently have to share wireframe selection highlighting-related code between multiple scene index filters, maybe we should work on that?  Perhaps a base class that would provide shared services for any filtering scene index that requires wireframe selection highlighting services?

I wonder if now is the time to start discussing if we need to place parameters inside the Hydra scene graph itself, at known and fixed locations, for use by scene indices?  We might even be able to leverage the Hydra dependencies mechanisms to propagate changes in parameters.

lib/flowViewport/selection/fvpSelection.h
```diff
@@ -95,6 +113,18 @@ class Selection
...
+
+    //We need to keep an order of selected paths to deal with the lead of the selection, the lead is the last of the selected items
+    PXR_NS::SdfPathVector _selectedPaths;
+
+    bool _IsLastSelected(const PXR_NS::SdfPath& primPath)const;
+
+    //Colors used by wireframe selection highlighting
+    PXR_NS::GfVec4f _activeWireframeColor;
+    PXR_NS::GfVec4f _leadWireframeColor;
+    PXR_NS::GfVec4f _dormantWireframeColor;
```
ppt-adsk: Should these be const if they're never supposed to change?

test/lib/mayaUsd/render/mayaToHydra/DataProducerSelHighlightTest/Storm_BoundingBox_AllSelected.png
ppt-adsk, line 1: Why is only the flat plane showing through in bounding box mode?  Is it because the other objects have white selection highlighting (as they are not lead objects)?  I don't see any white.

lib/flowViewport/sceneIndex/fvpBBoxSceneIndex.h
```diff
@@ -43,8 +45,8 @@ class BboxSceneIndex : public PXR_NS::HdSingleInputFilteringSceneIndexBase
     using PXR_NS::HdSingleInputFilteringSceneIndexBase::_GetInputSceneIndex;
 
     FVP_API
-    static BboxSceneIndexRefPtr New(const PXR_NS::HdSceneIndexBaseRefPtr& inputSceneIndex){
-        return PXR_NS::TfCreateRefPtr(new BboxSceneIndex(inputSceneIndex));
+    static BboxSceneIndexRefPtr New(const PXR_NS::HdSceneIndexBaseRefPtr& inputSceneIndex, const std::shared_ptr<const Selection>& selection){
```
ppt-adsk: Minor: since you've included fvpSelectionFwd.h, could be const SelectionConstPtr&

lib/flowViewport/sceneIndex/fvpWireframeSelectionHighlightSceneIndex.h
```diff
@@ -101,6 +101,7 @@ class WireframeSelectionHighlightSceneIndex
     bool isExcluded(const PXR_NS::SdfPath& sceneRoot) const;
 
     std::set<PXR_NS::SdfPath> _excludedSceneRoots;
+    PXR_NS::HdContainerDataSourceHandle _highlightSelectedPrim(const PXR_NS::HdContainerDataSourceHandle& dataSource, const PXR_NS::SdfPath& primPath)const;
```
ppt-adsk: Minor: the other private methods in this class don't have a leading underscore, so the fact that this one does is inconsistent.

lib/flowViewport/selection/fvpSelection.h
```diff
@@ -83,6 +87,20 @@ class Selection
...
 
+    //Get the wireframe color of a primitive for selection highlighting, 
+    // this checks if the prim is selected or not and if it is selected, 
+    // it returns the lead or active color depending on the lead of the selection
+    FVP_API
+    PXR_NS::GfVec4f GetWireframeColor(const PXR_NS::SdfPath& primPath)const;
+
+    //Ignore changes, this is used to keep the selection when we want to switch between display modes
+    FVP_API
+    void IgnoreChanges(bool val){_ignoreChanges = val;}
```
ppt-adsk: I don't understand the purpose of this.  Why would a display mode switch change the selection?

lib/flowViewport/selection/fvpSelection.cpp
```diff
@@ -74,31 +76,58 @@ Selection::_PrimSelectionState::GetVectorDataSource() const
...
+Selection::Add(const SdfPath& primPath)
 {
+    if (_ignoreChanges) return false;
+
     if (primPath.IsEmpty()) {
         return false;
     }
 
     _pathToState[primPath].selectionSources.push_back(selectionBuilder.Build());
+    if (_selectedPaths.end() == std::find(_selectedPaths.begin(), _selectedPaths.end(), primPath)){
```
ppt-adsk: For repeated additions, this will cause an O(n^2) loop, which we shouldn't have.  The UFE selection already prevents duplicate items being added:
https://help.autodesk.com/view/MAYAUL/2024/ENU/?guid=MAYA_API_REF_ufe_ref_class_ufe_1_1_selection_html

lib/flowViewport/selection/fvpSelection.cpp
```diff
@@ -74,31 +76,58 @@ Selection::_PrimSelectionState::GetVectorDataSource() const
...
 {
-    return (!primPath.IsEmpty() && (_pathToState.erase(primPath) == 1));
+    if (_ignoreChanges) return false;
+
+    const bool res = (!primPath.IsEmpty() && (_pathToState.erase(primPath) == 1));
+    if( res){
+        auto it = std::find(_selectedPaths.begin(), _selectedPaths.end(), primPath);
+        if (it != _selectedPaths.end()){
+            _selectedPaths.erase(it);
+        }
```
ppt-adsk: For repeated removal this is O(n^2).

lib/flowViewport/selection/fvpSelection.h
```diff
@@ -95,6 +113,18 @@ class Selection
     // Maps prim path to data sources to be returned by the vector data
     // source at locator selections.
     std::map<PXR_NS::SdfPath, _PrimSelectionState> _pathToState;
+
+    //We need to keep an order of selected paths to deal with the lead of the selection, the lead is the last of the selected items
+    PXR_NS::SdfPathVector _selectedPaths;
```
ppt-adsk: Instead of keeping a vector of the selected paths, can't we just keep the last one?  That's the only one we care about, and maintaining and checking that is O(1).

test/lib/mayaUsd/render/mayaToHydra/DataProducerSelHighlightTest/Storm_Wireframe_AllSelected.png
ppt-adsk, line 1: What is the difference between this file and Storm_Wire_AllSelInStorm.png?

test/lib/mayaUsd/render/mayaToHydra/StageVariantsTest/oneCube.png
ppt-adsk, line 1: If this is the only object in the scene and it's selected, why is the selection highlighting white?

test/lib/mayaUsd/render/mayaToHydra/StageVariantsTest/oneCubeInstanceable.png
ppt-adsk, line 1: Same question, if this is the only object in the scene and it's selected, why is the selection highlighting white?

test/lib/mayaUsd/render/mayaToHydra/StageVariantsTest/threeCubes.png
ppt-adsk, line 1: Same question, if these are the only objects in the scene and they're selected, why is the selection highlighting white?

test/lib/mayaUsd/render/mayaToHydra/StageVariantsTest/twoCubes.png
ppt-adsk, line 1: Why is the selection highlighting color white?

test/lib/mayaUsd/render/mayaToHydra/StageVariantsTest/threeCubesWithDisplacement.png
ppt-adsk, line 1: Why is the selection highlighting color white?

lib/flowViewport/sceneIndex/fvpReprSelectorSceneIndex.h
```diff
@@ -0,0 +1,109 @@
...
+    , public Fvp::InputSceneIndexUtils<ReprSelectorSceneIndex>
+{
+public:
+    using ParentClass = PXR_NS::HdSingleInputFilteringSceneIndexBase;
+    using PXR_NS::HdSingleInputFilteringSceneIndexBase::_GetInputSceneIndex;
+
+    enum class RepSelectorType{
+        WireframeRefined, //Wireframe
+        WireframeOnSurface, //Wireframe on surface not refined
+        WireframeOnSurfaceRefined,//Wireframe on surface refined
```
ppt-adsk, line 49: Surely a dumb question, but what does "Refined" mean?

lib/flowViewport/sceneIndex/fvpReprSelectorSceneIndex.cpp
```diff
@@ -0,0 +1,119 @@
...
+                        Fvp::PrimvarDataSource::New(
+                            HdRetainedTypedSampledDataSource<VtVec4fArray>::New(
+                                                VtVec4fArray{_selection->GetWireframeColor(primPath)}),
+                                                HdPrimvarSchemaTokens->constant,
+                                                HdPrimvarSchemaTokens->color));
+        //Edit the cull style
+        edited.Set(HdLegacyDisplayStyleSchema::GetCullStyleLocator(),
+                        HdRetainedTypedSampledDataSource<TfToken>::New(HdCullStyleTokens->nothing));//No culling
+    
+        prim.dataSource = HdOverlayContainerDataSource::New({ edited.Finish(), _wireframeTypeDataSource});
```
ppt-adsk: Nice!  Learned something new :)

lib/flowViewport/sceneIndex/fvpWireframeSelectionHighlightSceneIndex.cpp
```diff
@@ -94,6 +105,40 @@ WireframeSelectionHighlightSceneIndex::GetChildPrimPaths(const SdfPath &primPath
...
+        if (HdTokenArrayDataSourceHandle ds =
+                styleSchema.GetReprSelector()) {
+            VtArray<TfToken> ar = ds->GetTypedValue(0.0f);
+            TfToken refinedToken = ar[0];
+            if(HdReprTokens->refined == refinedToken){
+                //Is in refined display style, apply the wire on top of shaded reprselector
+                return HdOverlayContainerDataSource::New({ edited.Finish(), sRefinedWireOnSurfaceDisplayStyleDataSource});
+            }
+        }else{
+            //No reprSelector found, assume it's in the Collection that we have set HdReprTokens->refined
```
ppt-adsk, line 137: I would appreciate a more detailed comment outlining how the Hydra scene index data sources can be set up here, just to help understand what kinds of inputs you're trying to deal with.

lib/flowViewport/sceneIndex/fvpSelectionSceneIndex.cpp
```diff
@@ -39,20 +39,36 @@
...
 namespace {
-const HdDataSourceLocatorSet selectionsSchemaDefaultLocator{HdSelectionsSchema::GetDefaultLocator()};
+    //Handle primsvars:overrideWireframeColor in Storm for wireframe selection highlighting color
+    TF_DEFINE_PRIVATE_TOKENS(
+         _primVarsTokens,
+ 
+         (overrideWireframeColor)    // Works in HdStorm to override the wireframe color
+     );
+
+    const HdDataSourceLocatorSet selectionsAndPrimvarsColorsLocatorSet{HdSelectionsSchema::GetDefaultLocator(),
```
ppt-adsk: Separation of concerns would imply that the selection scene index know nothing about how to highlight objects, whether it's done with wireframe, the fact that in Maya there is a lead object with a different color, etc.  All this is very Maya-centric.  The selection scene index should only know about how to translate changes in the application selection into changes in the Hydra scene index selection.

test/lib/mayaUsd/render/mayaToHydra/testDataProducerSelHighlight.py
```diff
@@ -0,0 +1,186 @@
...
+
+        #Torus
+        pTorusUfePath = ufe.Path([ufeSegment, usdUtils.createUfePathSegment("/pTorus1")])
+        pTorusUfeItem = ufe.Hierarchy.createItem(pTorusUfePath)
+        self.assertIsNotNone(pTorusUfeItem)
+
+        #Plane
+        pPlaneUfePath = ufe.Path([ufeSegment, usdUtils.createUfePathSegment("/pPlane1")])
+        pPlaneUfeItem = ufe.Hierarchy.createItem(pPlaneUfePath)
+        self.assertIsNotNone(pPlaneUfeItem)
```
ppt-adsk: Is this the same as line 53?  I know this is a test, but you might want to factor this out.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -340,6 +341,15 @@ Ufe::Path usdPathToUfePath(
         registration->pluginSceneIndex, usdPath) : Ufe::Path();
 }
 
+inline bool areDifferentForOneOfTheseBits(unsigned int val1, unsigned int val2, unsigned int bitsToTest)
+{
+    if ((val1 & bitsToTest) != (val2 & bitsToTest)) {
+        return true;
+    }
+
+    return false;
```
ppt-adsk: Nit-pick: could just be
```suggestion
    return  ((val1 & bitsToTest) != (val2 & bitsToTest));
```

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1043,6 +1040,10 @@ MStatus MtohRenderOverride::Render(
             //Create a HydraViewportInformation 
             const Fvp::InformationInterface::ViewportInformation hydraViewportInformation(std::string(panelName.asChar()), cameraName);
             manager.AddViewportInformation(hydraViewportInformation, _renderIndexProxy, _lastFilteringSceneIndexBeforeCustomFiltering);
+            //Update the selection since we may have added scene indices through manager.AddViewportInformation to the merging scene index
```
ppt-adsk: Shouldn't we do this then only when we add scene indices to the merging scene index?  Otherwise, we're setting the Hydra scene index selection at every render, which seems wasteful.

ppt-adsk (changes requested):
Amazing work!  I have some concerns about Maya selection highlighting concepts bleeding into the Flow Viewport Toolkit selection object and selection scene index, but you've done a fantastic job.

lib/flowViewport/selection/fvpSelection.h
lanierd-adsk: Not sure I can make them const since they are initialized with :
Selection::Selection()
{
    Fvp::ColorPreferences::getInstance().getColor(FvpColorPreferencesTokens->wireframeSelection, _leadWireframeColor);
    Fvp::ColorPreferences::getInstance().getColor(FvpColorPreferencesTokens->wireframeSelectionSecondary, _activeWireframeColor);
    Fvp::ColorPreferences::getInstance().getColor(FvpColorPreferencesTokens->polymeshDormant, _dormantWireframeColor);
}

And there is a possibility that they are updated by the user in the UI, but at this time we need to switch to VP2 and back to Hydra to update.

lib/flowViewport/sceneIndex/fvpBBoxSceneIndex.h
lanierd-adsk: Thanks, I fixed it in various places.

lib/flowViewport/sceneIndex/fvpReprSelectorSceneIndex.h
lanierd-adsk: For me refined means it has a better quality than the non refined.

lib/flowViewport/sceneIndex/fvpWireframeSelectionHighlightSceneIndex.cpp
lanierd-adsk: I've added a few more comments.

lib/flowViewport/sceneIndex/fvpReprSelectorSceneIndex.h
lanierd-adsk: But I think what you want to know is that "refined" means "shaded" in Hydra.

lib/flowViewport/selection/fvpSelection.cpp
lanierd-adsk: Do you have a suggestion to avoid this ?

lib/flowViewport/selection/fvpSelection.h
lanierd-adsk: That was the case, but let me check again since I did some modifications, maybe it's no longer needed.

lib/flowViewport/selection/fvpSelection.h
lanierd-adsk: I don't think I can do this : when you remove the last one from the selection I need to know which one was selected before, as it becomes the lead object now. Same case when you remove several objects from the selection, the lead may have changed. So I think we need an array of all selected objects.

lib/flowViewport/selection/fvpSelection.h
lanierd-adsk: Here is the reason why I need this :
When updating the internal scene indices chain (not using any custom filtering scene index from the flow viewport API) after the merging scene index. We need to remove from the renderIndex the last scene index of the chain. In cascade, this produces a callstack of :
 	flowViewport.dll!Fvp_v0::Selection::RemoveHierarchy(const pxrInternal_v0_23__pxrReserved__::SdfPath & primPath) Line 145	C++
>	flowViewport.dll!Fvp_v0::SelectionSceneIndex::_PrimsRemoved(const pxrInternal_v0_23__pxrReserved__::HdSceneIndexBase & sender, const pxrInternal_v0_23__pxrReserved__::TfSmallVector<pxrInternal_v0_23__pxrReserved__::HdSceneIndexObserver::RemovedPrimEntry,16> & entries) Line 344	C++
 	usd_hd.dll!pxrInternal_v0_23__pxrReserved__::HdSceneIndexBase::_SendPrimsRemoved(const pxrInternal_v0_23__pxrReserved__::TfSmallVector<pxrInternal_v0_23__pxrReserved__::HdSceneIndexObserver::RemovedPrimEntry,16> & entries) Line 144	C++
 	usd_hd.dll!pxrInternal_v0_23__pxrReserved__::HdMergingSceneIndex::RemoveInputScene(const pxrInternal_v0_23__pxrReserved__::TfRefPtr<pxrInternal_v0_23__pxrReserved__::HdSceneIndexBase> & sceneIndex) Line 188	C++
 	flowViewport.dll!Fvp_v0::RenderIndexProxy::RemoveSceneIndex(const pxrInternal_v0_23__pxrReserved__::TfRefPtr<pxrInternal_v0_23__pxrReserved__::HdSceneIndexBase> & inputScene) Line 119	C++

So RemoveHierarchy is called and we lose the selection.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lanierd-adsk: You are right, I modified the code to only do that when we have added data producer scene indices

test/lib/mayaUsd/render/mayaToHydra/DataProducerSelHighlightTest/Storm_BoundingBox_AllSelected.png
lanierd-adsk: It's from the github viewer, as the real image is correct and shows white boxes.

lib/flowViewport/sceneIndex/fvpSelectionSceneIndex.cpp
lanierd-adsk: We need to discuss about that.

lib/flowViewport/selection/fvpSelection.h
ppt-adsk: But can't we rely on the application selection for this?  On each application selection change, you can update the Hydra shadow of the last selected object.

lib/flowViewport/selection/fvpSelection.cpp
ppt-adsk: Yes, only keep a shadow of the last element.

lib/flowViewport/selection/fvpSelection.h
ppt-adsk: I'd like to discuss this one with you.  It seems like we want to reconfigure the scene index tree, but leave scene index prims alone.  Multiple scene indices may have this requirement.  I have an intuition that we could create an insulating scene index filter that would simply stop propagating prims removed messages while we reconfigure the pipeline.  This would provide a service for multiple filtering scene indices.

lib/flowViewport/selection/fvpSelection.h
ppt-adsk: Great point.  Since Fvp::ColorPreferences is a Ufe::Subject and is therefore observable, should we enter a JIRA ticket to do this work later, or do you want to do this in the current pull request?  Completely up to you, there are advantages and disadvantages to each approach (current PR is already big, but on the other hand you know pretty much what to do and there is less process to doing it now).

lib/flowViewport/sceneIndex/fvpReprSelectorSceneIndex.h
ppt-adsk: Could we add that to the code comment?

lanierd-adsk: I've added this in the .h and .cpp file :
refined means that it supports a "refineLevel" attribute in the displayStyle to get a more refined drawing, valid range is from 0 to 8.

lib/flowViewport/selection/fvpSelection.h
lanierd-adsk: Let's discuss about that

lib/flowViewport/selection/fvpSelection.h
lanierd-adsk: Let's discuss about that.

lib/flowViewport/selection/fvpSelection.h
lanierd-adsk: Le's discuss about that

lib/flowViewport/selection/fvpSelection.h
ppt-adsk, line 1: Are changes still required in this file?  The vec4f.h include seems unnecessary, and the default constructor will be implicitly declared by the compiler.

lib/flowViewport/selection/fvpSelection.cpp
ppt-adsk, line 1: Are the changes in this file still necessary?  Omitting the PXR_NS is fine, but perhaps I would just have removed the PXR_NAMESPACE_USING_DIRECTIVE.

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.h
```diff
@@ -0,0 +1,75 @@
...
+#include "mayaHydraLib/api.h"
+#include "mayaHydraLib/mayaHydra.h"
+#include "mayaHydraLib/sceneIndex/mhDirtyLeadObjectSceneIndex.h"
+
+//Flow viewport headers
+#include <flowViewport/sceneIndex/fvpPathInterface.h>
+
+//ufe
+#include <ufe/observer.h>
+#include <ufe/sceneItem.h>
```
ppt-adsk: Unused, should remove.

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.h
```diff
@@ -0,0 +1,75 @@
...
+    ~MhLeadObjectPathTracker();
+
+    MAYAHYDRALIB_API
+    bool isLeadObject(const PXR_NS::SdfPath& primPath) const;
+    
+    MAYAHYDRALIB_API
+    Ufe::Path getLeadObjectUfePath() const {return _leadObjectUfePath;}
+
+    MAYAHYDRALIB_API
+    void setNewLeadObjectSceneItem(const Ufe::Path& newLeadObjectUfePath);
```
ppt-adsk: Oddly named --- the accessor is `getLeadObjectUfePath()`, so I'd name this `setLeadObjectUfePath()`.

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.cpp
```diff
@@ -0,0 +1,169 @@
...
+#include "mhLeadObjectPathTracker.h"
+
+//Flow viewport headers
+#include <flowViewport/colorPreferences/fvpColorPreferences.h>
+#include <flowViewport/colorPreferences/fvpColorPreferencesTokens.h>
+#include <flowViewport/selection/fvpSelection.h>
+
+//ufe
+#include <ufe/globalSelection.h>
+#include <ufe/selection.h>
```
ppt-adsk: Unneeded, remove (Ufe::ObservableSelection isa Ufe::Selection).

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.cpp
```diff
@@ -0,0 +1,169 @@
...
+                    case Ufe::SelectionChanged::OpType::SelectionCompositeNotification://Fall into
+                    case Ufe::SelectionChanged::OpType::ReplaceWith: {
+                        UpdateLeadObjectSceneItem(leadObjectUfePath);
+                        break;
+                    }
+                    default: break;
+                }
+            }
+
+            void UpdateLeadObjectSceneItem(const Ufe::Path& currentLeadObjectUfePath) {
```
ppt-adsk: Oddly named, no scene item object is passed around.  I'd just name it UpdateLeadObject().

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.cpp
```diff
@@ -0,0 +1,169 @@
...
+                    auto newLeadObjectUfePath = globalSelection->back()->path();
+                    if (newLeadObjectUfePath != currentLeadObjectUfePath){
+                        _leadObjectPathInstance.setNewLeadObjectSceneItem(newLeadObjectUfePath);
+                    }
+                }else {
+                    _leadObjectPathInstance.ClearLeadObject();
+                }
+            }
+
+            MayaHydra::MhLeadObjectPathTracker& _leadObjectPathInstance;
```
ppt-adsk: Confusing name, how about
```suggestion
            MayaHydra::MhLeadObjectPathTracker& _leadObjectPathTracker;
```

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.cpp
```diff
@@ -0,0 +1,169 @@
...
+                    
+                    case Ufe::SelectionChanged::OpType::Insert://Fall into
+                    case Ufe::SelectionChanged::OpType::SelectionCompositeNotification://Fall into
+                    case Ufe::SelectionChanged::OpType::ReplaceWith: {
+                        UpdateLeadObjectSceneItem(leadObjectUfePath);
+                        break;
+                    }
+                    default: break;
+                }
+            }
```
ppt-adsk: I honestly wonder if all this code complexity is worth it...  We could just replace all this with `UpdateLeadObject()`, or better yet, just take the `UpdateLeadObject()` code and move it into `operator()()`.  In some sense, whenever there's a selection change, whatever the selection change is, we just update the lead object.  Code would be much simpler.

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.cpp
```diff
@@ -0,0 +1,169 @@
...
+    if(_dirtyLeadObjectSceneIndex && ! oldLeadObjectPrimPath.IsEmpty()){
+        _dirtyLeadObjectSceneIndex->dirtyLeadObjectRelatedPrims(oldLeadObjectPrimPath, _leadObjectPrimPath);
+    }
+}
+
+void MhLeadObjectPathTracker::ClearLeadObject()
+{
+    _leadObjectPrimPath   = SdfPath::EmptyPath();
+    _leadObjectUfePath    = Ufe::Path();
+}
```
ppt-adsk: I think there's a bug here, you're not dirtying the previous _leadObjectPrimPath.  I had reservations about this interface, and now I have even more reservations about it :)  Why not remove it completely, and have `setLeadObjectUfePath()` take an empty Ufe::Path to clear?  That way dirtying logic would be in a single method, `setLeadObjectUfePath()`.  This would also narrow your interface, which is usually a good thing.

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+void MhLeadObjectPathTracker::updateAfterDataProducerSceneIndicesLoaded()
```
ppt-adsk: I think the naming here is awkward; we shouldn't care about the clients of this method, we should only care about what it does.  Also, it doesn't update everything, just the prim path, so I'd call it `updatePrimPath()`.

lib/flowViewport/sceneIndex/fvpWireframeSelectionHighlightSceneIndex.h
```diff
@@ -104,6 +107,7 @@ class WireframeSelectionHighlightSceneIndex
     PXR_NS::HdContainerDataSourceHandle _HighlightSelectedPrim(const PXR_NS::HdContainerDataSourceHandle& dataSource, const PXR_NS::SdfPath& primPath)const;
 
     const SelectionConstPtr   _selection;
+    const WireframeColorInterface& _wireframeColorInterface;
```
ppt-adsk: As soon as I saw this reference to the object, I wondered about object lifescope: who owns it?  How can we be sure it won't disappear before the WireframeSelectionHighlightSceneIndex?  Since you're creating this using a shared pointer, why not pass it as a shared pointer and store it as a shared pointer?  This way all lifescope questions disappear.

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.cpp
```diff
@@ -0,0 +1,169 @@
...
+        const Ufe::Selection& selection = *(ufeSelection);
+        auto leadObjectSceneItem = selection.back();//get last selected
+        _leadObjectUfePath = leadObjectSceneItem->path();
+        //_leadObjectPrimPath can be empty with a valid _leadObjectUfePath when the lead object is in a data producer scene index not yet added to the merging scene index
+        //This is fixed at some point by calling updateAfterDataProducerSceneIndicesLoaded()
+        _leadObjectPrimPath = _pathInterface->SceneIndexPath(_leadObjectUfePath); 
+    }
+
+   // Add ourself as an observer to the selection
+   _ufeSelectionObserver = std::make_shared<GlobalSelectionChangedObs>(*this);
```
ppt-adsk: Could go into the initializer list, and then the _ufeSelectionObserver could be const.

lib/mayaHydra/hydraExtensions/mhWireframeColorInterfaceImp.h
```diff
@@ -0,0 +1,64 @@
...
+    //Get the wireframe color of a primitive for selection highlighting, 
+    // this checks if the prim is selected or not and if it is selected, 
+    // it returns the maya lead or active color depending on the lead of the selection
+    MAYAHYDRALIB_API
+    PXR_NS::GfVec4f getWireframeColor(const PXR_NS::SdfPath& primPath) const override;
+
+private:
+    //returns true if the sdfpath is selected or false if it is not. 
+    // If the function returns true, the outIsTheLastSelected is true if the prim was the last selected prim
+    bool _isSelected(const PXR_NS::SdfPath& primPath, bool& outIsTheLastSelected)const;
```
ppt-adsk: I feel this is a bit awkward; why not go with a 3-state enum?
```suggestion
    enum SelectionState {kLead, kActive, kDormant};
    SelectionState getSelectionState(const PXR_NS::SdfPath& primPath) const;
```

lib/mayaHydra/hydraExtensions/mhWireframeColorInterfaceImp.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+MhWireframeColorInterfaceImp::MhWireframeColorInterfaceImp(const std::shared_ptr<Fvp::Selection>& selection
+                                                         , const std::shared_ptr<MhLeadObjectPathTracker>& leadObjectPathTracker) 
+    : _selection(selection)
+    , _leadObjectPathTracker(leadObjectPathTracker)
+{ 
+    TF_AXIOM(_selection);
+    Fvp::ColorPreferences::getInstance().getColor(FvpColorPreferencesTokens->wireframeSelection, _leadWireframeColor);
+    Fvp::ColorPreferences::getInstance().getColor(FvpColorPreferencesTokens->wireframeSelectionSecondary, _activeWireframeColor);
+    Fvp::ColorPreferences::getInstance().getColor(FvpColorPreferencesTokens->polymeshDormant, _dormantWireframeColor);
```
ppt-adsk: Nit-pick: I'd just create a function in the unnamed namespace
```suggestion
    PXR_NS::GfVec4f getPreferencesColor(const PXR_NS::TfToken& token) {
        PXR_NS::GfVec4f color;
        Fvp::ColorPreferences::getInstance().getColor(token, color);
        return color;
}
```
which would allow you to add your color initializations in the MhWireframeColorInterfaceImp constructor initializer list.

lib/flowViewport/sceneIndex/fvpBlockPrimRemovalPropagationSceneIndex.h
```diff
@@ -0,0 +1,107 @@
...
+#define FLOW_VIEWPORT_SCENEINDEX_FVP_BLOCK_PRIM_REMOVAL_PROPAGATION_SCENE_INDEX_H
+
+//Local headers
+#include "flowViewport/api.h"
+#include "flowViewport/sceneIndex/fvpSceneIndexUtils.h"
+#include "flowViewport/selection/fvpSelectionFwd.h"
+#include "flowViewport/sceneIndex/fvpPathInterface.h"
+
+//Hydra headers
+#include <pxr/base/tf/declarePtrs.h>
```
ppt-adsk: Are we using this header?

lib/flowViewport/sceneIndex/fvpBlockPrimRemovalPropagationSceneIndex.h
```diff
@@ -0,0 +1,107 @@
...
+#include <pxr/imaging/hd/filteringSceneIndex.h>
+#include <pxr/imaging/hd/retainedDataSource.h>
```
ppt-adsk: Don't seem to be using this header.

lib/flowViewport/sceneIndex/fvpBlockPrimRemovalPropagationSceneIndex.h
```diff
@@ -0,0 +1,107 @@
...
+    FVP_API
+    void setPrimRemovalBlocked(bool blockPrimRemoval) { _blockPrimRemoval = blockPrimRemoval; }
+
+    //from PathInterface
+    FVP_API
+    PXR_NS::SdfPath SceneIndexPath(const Ufe::Path& appPath) const override{
+        auto pathInterface = dynamic_cast<const PathInterface*>(&*GetInputSceneIndex());
+        if (!pathInterface){
+            return {};
+        }
```
ppt-adsk: We should do this once, in the constructor, and it should be a TF_AXIOM.

lib/flowViewport/sceneIndex/fvpSelectionSceneIndex.cpp
ppt-adsk, line 1: Shouldn't we simply revert all changes to this file and move responsibility for this to the wireframe scene index?  The selection scene index shouldn't know anything about wireframe.

ppt-adsk (changes requested):
Looking really good, thanks for the changes!  There are still wireframe-related changes to the Fvp::SelectionSceneIndex; I was hoping we could remove those somehow and keep the SelectionSceneIndex unaware of how selection highlighting is done.

lib/flowViewport/sceneIndex/fvpBlockPrimRemovalPropagationSceneIndex.h
lanierd-adsk: Yes in renderoverride.h we have a member :
Fvp::BlockPrimRemovalPropagationSceneIndexRefPtr  _blockPrimRemovalPropagationSceneIndex;

lanierd-adsk: Right, I removed it, thanks.

lib/flowViewport/sceneIndex/fvpBlockPrimRemovalPropagationSceneIndex.h
lanierd-adsk: Also right, I removed it, thanks.

lib/flowViewport/sceneIndex/fvpBlockPrimRemovalPropagationSceneIndex.h
lanierd-adsk: Right, done. Thanks.

lib/flowViewport/sceneIndex/fvpSelectionSceneIndex.cpp
lanierd-adsk: This file is the fvpReprSelectorSceneIndex.h not the selection scene index, I've added it to this PR. Are you sure of what you meant ?

lib/flowViewport/sceneIndex/fvpSelectionSceneIndex.cpp
lanierd-adsk: This has been done already.

lib/flowViewport/sceneIndex/fvpWireframeSelectionHighlightSceneIndex.h
lanierd-adsk: You're right, I amde the change in the various places where this was used.

lib/flowViewport/selection/fvpSelection.cpp
lanierd-adsk: Right, I have completely reverted this file.

lib/flowViewport/selection/fvpSelection.h
lanierd-adsk: This file has been reverted completly.

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.cpp
lanierd-adsk: Done

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.cpp
lanierd-adsk: Done.

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.cpp
lanierd-adsk: this function has been removed.

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.cpp
lanierd-adsk: Right, done. Thanks.

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.cpp
lanierd-adsk: Right, thanks for catching this.

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.cpp
lanierd-adsk: You're right, done. Thanks.

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.h
lanierd-adsk: Done

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.h
lanierd-adsk: Done

lib/mayaHydra/hydraExtensions/mhWireframeColorInterfaceImp.cpp
lanierd-adsk: Done

lib/mayaHydra/hydraExtensions/mhWireframeColorInterfaceImp.h
lanierd-adsk: Thanks, done.

lib/flowViewport/sceneIndex/fvpSelectionSceneIndex.cpp
lanierd-adsk: I reverted fvpSelectionSceneIndex completely.

ppt-adsk (approved):

PR #120, opened 2024-04-23
Fixed time update, Bifrost stage update, and USD stage file update

ppt-adsk:
With tests for time update and USD stage file update.

### DIFF
--- a/lib/flowViewport/CMakeLists.txt
+++ b/lib/flowViewport/CMakeLists.txt
@@ -8,6 +8,7 @@ add_library(${TARGET_NAME} SHARED)
 target_sources(${TARGET_NAME} 
     PRIVATE
         debugCodes.cpp
+        fvpInstruments.cpp
         fvpUtils.cpp
         global.cpp
         tokens.cpp
@@ -16,7 +17,7 @@ target_sources(${TARGET_NAME}
 set(HEADERS
     api.h
     debugCodes.h
-    fvpUtils.h
+    fvpInstruments.h
     global.h
     tokens.h
 )
--- a/lib/flowViewport/CMakeLists.txt
+++ b/lib/flowViewport/CMakeLists.txt
@@ -18,6 +18,7 @@ set(HEADERS
     api.h
     debugCodes.h
     fvpInstruments.h
+    fvpUtils.h
     global.h
     tokens.h
 )
--- a/lib/flowViewport/fvpInstruments.h
+++ b/lib/flowViewport/fvpInstruments.h
@@ -1,4 +1,4 @@
-// Copyright 2023 Autodesk
+// Copyright 2024 Autodesk
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -12,27 +12,38 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
-#ifndef MH_CPPTESTS_CMD
-#define MH_CPPTESTS_CMD
+#ifndef FVP_INSTRUMENTS
+#define FVP_INSTRUMENTS
 
-#include <maya/MPxCommand.h>
+#include <flowViewport/api.h>
 
-class mayaHydraCppTestCmd : public MPxCommand
+#include <pxr/base/vt/value.h>
+
+#include <string>
+
+namespace FVP_NS_DEF {
+
+/// \class Instruments
+///
+/// A registry to measure Flow Viewport processing.
+///
+class Instruments
 {
 public:
-    static void*   creator() { return new mayaHydraCppTestCmd(); }
-    static MSyntax createSyntax();
 
-    MStatus doIt(const MArgList& args) override;
+    FVP_API
+    static Instruments& instance();
+
+    FVP_API
+    PXR_NS::VtValue get(const std::string& key) const;
+    FVP_API
+    void set(const std::string& key, const PXR_NS::VtValue& v);
+
+private:
+
+    Instruments() = default;
 };
 
-class mayaHydraInstruments : public MPxCommand
-{
-public:
-    static void*   creator() { return new mayaHydraInstruments(); }
-    static MSyntax createSyntax();
+} // namespace FVP_NS_DEF
 
-    MStatus doIt(const MArgList& args) override;
-};
-
-#endif // MH_CPPTESTS_CMD
+#endif // FVP_INSTRUMENTS
--- a/lib/mayaHydra/hydraExtensions/sceneIndex/mhMayaUsdProxyShapeSceneIndex.cpp
+++ b/lib/mayaHydra/hydraExtensions/sceneIndex/mhMayaUsdProxyShapeSceneIndex.cpp
@@ -16,15 +16,14 @@
 
 #include "mhMayaUsdProxyShapeSceneIndex.h"
 
-#if defined(MAYAHYDRALIB_MAYAUSDAPI_ENABLED)
+#include <flowViewport/fvpInstruments.h>
 
-//Maya headers
-#include <maya/MEventMessage.h>//For timeChanged callback
+#if defined(MAYAHYDRALIB_MAYAUSDAPI_ENABLED)
 
 //mayaHydra headers
 #include "ufeExtensions/Global.h"
 
-PXR_NAMESPACE_OPEN_SCOPE
+PXR_NAMESPACE_USING_DIRECTIVE
 
 namespace MAYAHYDRA_NS_DEF {
 
@@ -40,15 +39,16 @@ MayaUsdProxyShapeSceneIndex::MayaUsdProxyShapeSceneIndex(const MAYAUSDAPI_NS::Pr
 {
     TfWeakPtr<MayaUsdProxyShapeSceneIndex> ptr(this);
     _stageSetNoticeKey = TfNotice::Register(ptr, &MayaUsdProxyShapeSceneIndex::_StageSet);
+    _stageInvalidateNoticeKey = TfNotice::Register(ptr, &MayaUsdProxyShapeSceneIndex::_StageInvalidate);
     _objectsChangedNoticeKey = TfNotice::Register(ptr, &MayaUsdProxyShapeSceneIndex::_ObjectsChanged);
-    
-    _timeChangeCallbackId = MEventMessage::addEventCallback("timeChanged", onTimeChanged, this);
+
+    Fvp::Instruments::instance().set(kNbPopulateCalls, VtValue(_nbPopulateCalls));
 }
 
 MayaUsdProxyShapeSceneIndex::~MayaUsdProxyShapeSceneIndex()
 {
-    MMessage::removeCallback(_timeChangeCallbackId);
     TfNotice::Revoke(_stageSetNoticeKey);
+    TfNotice::Revoke(_stageInvalidateNoticeKey);
     TfNotice::Revoke(_objectsChangedNoticeKey);
 }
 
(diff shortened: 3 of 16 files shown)


lib/flowViewport/fvpInstruments.h
ppt-adsk, line 1: Added trivial string keyed dictionary to keep track of instrumentation values in mayaHydra.  Useful for reasoning and assertions.

test/testSamples/testUsdAnim/cubeAnim.usda
ppt-adsk, line 1: USD cube with animated translation.

lib/mayaHydra/hydraExtensions/sceneIndex/registration.cpp
```diff
@@ -204,6 +204,14 @@ PXR_NAMESPACE_OPEN_SCOPE
 // Remove this once the code has been moved to the MayaHydra namespace.
 using namespace MayaHydra;
 
+struct MayaUsdSceneIndexRegistration : public MayaHydraSceneIndexRegistration
```
ppt-adsk, line 207: This is the central construct in fixing HYDRA-930: we receive notice of time invalidation and immediately call compute() on the proxy shape output time, during dirty propagation.  This returns the previous time, which is incorrect.  Instead, simply refresh time for the USD data producer at render time, at which point the proxy shape output compute() produces the correct result.  This matches the way the VP 2 render delegate sets time.

lib/mayaHydra/hydraExtensions/sceneIndex/registration.h
```diff
@@ -59,6 +59,8 @@ struct MayaHydraSceneIndexRegistration
     SdfPath                     sceneIndexPathPrefix;
     MObjectHandle               dagNode;
     MayaHydraInterpretRprimPath interpretRprimPathFn = nullptr;
+
+    virtual void Update() = 0;
```
ppt-adsk, line 63: New hook method to upate the registered data provider at render time.  Used by USD registration to update time.

lib/mayaHydra/hydraExtensions/sceneIndex/mhMayaUsdProxyShapeSceneIndex.cpp
```diff
@@ -60,24 +60,35 @@ MayaUsdProxyShapeSceneIndexRefPtr MayaUsdProxyShapeSceneIndex::New(const MAYAUSD
...
-        return;
-    }
-    instance->UpdateTime();
-}
-
 void MayaUsdProxyShapeSceneIndex::UpdateTime()
 {
     if (_usdImagingStageSceneIndex && _dagNodeHandle.isValid()) {
-        _usdImagingStageSceneIndex->SetTime(_proxyStage.getTime());//We have the possibility to scale and offset the time in _proxyShapeBase
+        // _usdImagingStageSceneIndex->SetTime(_proxyStage.getTime());//We have the possibility to scale and offset the time in _proxyShapeBase
```
ppt-adsk: Apologies, no real change here, debugging only, will revert when I do review fixes.

lib/mayaHydra/hydraExtensions/sceneIndex/mhMayaUsdProxyShapeSceneIndex.h
```diff
@@ -36,19 +36,23 @@
 
 //Maya headers
 #include <maya/MObjectHandle.h>
-#include <maya/MMessage.h>
```
ppt-adsk, line 39: Updating time at every render, no need to be listening to time changes.

lib/mayaHydra/hydraExtensions/sceneIndex/mhMayaUsdProxyShapeSceneIndex.h
```diff
@@ -36,19 +36,23 @@
...
 
-//Ufe headers
-#include <ufe/path.h>
+UFE_NS_DEF {
+class Path;
+}
 
 #include <memory>
 
-PXR_NAMESPACE_OPEN_SCOPE
```
ppt-adsk, line 46: Class MayaUsdProxyShapeSceneIndex was both in the Pixar namespace and the mayaHydra namespace, which was a bit much, belongs in the mayaHydra namespace only.

lib/mayaHydra/hydraExtensions/sceneIndex/mhMayaUsdProxyShapeSceneIndex.h
```diff
@@ -60,6 +64,8 @@ class MayaUsdProxyShapeSceneIndex : public HdSingleInputFilteringSceneIndexBase
     using PXR_NS::HdSingleInputFilteringSceneIndexBase::_GetInputSceneIndex;
     using ParentClass = HdSingleInputFilteringSceneIndexBase;
 
+    static constexpr char kNbPopulateCalls[] = "MayaUsdProxyShapeSceneIndex:NbPopulateCalls";
```
ppt-adsk, line 67: Key for number of populate calls instrument.

lib/mayaHydra/hydraExtensions/sceneIndex/mhMayaUsdProxyShapeSceneIndex.h
```diff
@@ -95,24 +101,23 @@ class MayaUsdProxyShapeSceneIndex : public HdSingleInputFilteringSceneIndexBase
 private:
     void _ObjectsChanged(const MAYAUSDAPI_NS::ProxyStageObjectsChangedNotice& notice);
     void _StageSet(const MAYAUSDAPI_NS::ProxyStageSetNotice& notice);
+    void _StageInvalidate(const MAYAUSDAPI_NS::ProxyStageInvalidateNotice& notice);
```
ppt-adsk, line 104: Now listening for proxy shape stage invalidate notifications, which include integer stage ID increments, USD file path changes, etc.  See
https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/nodes/proxyShapeBase.cpp

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1173,6 +1173,12 @@ MStatus MtohRenderOverride::Render(
         _taskController->SetSelectionEnableOutline(false);
 
     _taskController->SetCollection(_renderCollection);
+
+    // Update all registered plugin before render.
+    for (auto& entry : _sceneIndexRegistry->GetRegistrations()) {
```
ppt-adsk, line 1178: Allow registered plugins to update before render.  Our USD registered data provider uses this to update time.

test/lib/mayaUsd/render/mayaToHydra/CMakeLists.txt
```diff
@@ -75,6 +75,13 @@ set(INTERACTIVE_TEST_SCRIPT_FILES_MESH_ADAPTER
     cpp/testMeshAdapterTransform.py
 )
 
+# Run the following tests with the VP2 render delegate disabled, to ensure 
+# that mayaHydra does not rely on computations made by VP2 for its use.
+set(INTERACTIVE_TEST_SCRIPT_FILES_DISABLE_VP2_RENDER_DELEGATE
```
ppt-adsk, line 80: Discovered that time computation (pulling on the proxy shape time output) was being done by the VP 2 render delegate, which meant that this value was being provided to the Maya Hydra viewport as a side effect.  Make sure that tests run without the VP 2 render delegate to insulate the test from this.  We should revisit this and perhaps turn of VP 2 render delegate updates when it is not the viewport renderer, while ensuring that correctness is maintained.

test/lib/mayaUsd/render/mayaToHydra/cpp/testHydraPrim.cpp
```diff
@@ -0,0 +1,45 @@
...
+    const Ufe::Path appPath(Ufe::PathString::path(argv[0]));
+
+    // Translate the application path into a scene index path using the
+    // selection scene index.
+    const auto snSi = findSelectionSceneIndexInTree(siRoot);
+    ASSERT_TRUE(snSi);
+
+    const auto sceneIndexPath = snSi->SceneIndexPath(appPath);
+
+    ASSERT_FALSE(sceneIndexPath.IsEmpty());
```
ppt-adsk, line 44: This is just testing that an application path object has a corresponding Hydra scene index prim.  This could be better done with a Python binding to a C++ function, rather than abusing our test framework test plugin command to do so.

test/lib/mayaUsd/render/mayaToHydra/cpp/mayaHydraCppTestsCmd.h
```diff
@@ -23,9 +23,16 @@ class mayaHydraCppTestCmd : public MPxCommand
     static void*   creator() { return new mayaHydraCppTestCmd(); }
     static MSyntax createSyntax();
 
-    static const MString name;
+    MStatus doIt(const MArgList& args) override;
+};
+
+class mayaHydraInstruments : public MPxCommand
```
ppt-adsk, line 29: Query the instruments dictionary from a Python test.  Would be much better done with a Python binding to a C++ function.

test/lib/mayaUsd/render/mayaToHydra/cpp/testUsdStageFromFile.py
ppt-adsk, line 1: Test that swapping USD files changes the Hydra scene index scene.

test/lib/mayaUsd/render/mayaToHydra/cpp/testUsdStageFromFile.py
```diff
@@ -0,0 +1,61 @@
...
+
+            populateCallsPre = cmds.mayaHydraInstruments("MayaUsdProxyShapeSceneIndex:NbPopulateCalls", q=True)
+
+            usdScenePath = testUtils.getTestScene('testStagePayloadsReferences', 'cone.usda')
+
+            cmds.setAttr('cubeShape.filePath', usdScenePath, type="string")
+
+            populateCallsPost = cmds.mayaHydraInstruments("MayaUsdProxyShapeSceneIndex:NbPopulateCalls", q=True)
+
+            self.assertEqual(populateCallsPre+1, populateCallsPost)
```
ppt-adsk, line 56: Assert that only one populate call was done to replace the USD scene.

test/lib/mayaUsd/render/mayaToHydra/cpp/mayaHydraCppTestsCmd.cpp
```diff
@@ -93,16 +96,82 @@ MStatus mayaHydraCppTestCmd::doIt( const MArgList& args )
...
+
+//============================================================
+// CLASS mayaHydraCppTests
+//============================================================
+
 MStatus initializePlugin( MObject obj ) 
 {
     MFnPlugin plugin( obj, "Autodesk", "1.0", "Any" );
     plugin.registerCommand( "mayaHydraCppTest", mayaHydraCppTestCmd::creator, mayaHydraCppTestCmd::createSyntax );
+    plugin.registerCommand( "mayaHydraInstruments", mayaHydraInstruments::creator, mayaHydraInstruments::createSyntax );
```
ppt-adsk, line 167: Query the instruments dictionary through a Maya command, for use in Python tests.  Would be better done through a Python binding of a C++ function.

test/lib/mayaUsd/render/mayaToHydra/cpp/testUsdAnim.py
ppt-adsk, line 1: Simple Python driver for the real time varying test in testUsdAnim.cpp.

test/lib/mayaUsd/render/mayaToHydra/cpp/testUsdAnim.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+    // Extract the Hydra xform matrix from the cube prim
+    // Loop over time, and check translation, from frames [0..10].  The z value
+    // of the translation has been set equal to time.
+    for (double t=0; t < 11.0; t+=1.0) {
+        ASSERT_EQ(MGlobal::viewFrame(t), MS::kSuccess);
+
+        GfMatrix4d cubeHydraMatrix;
+        ASSERT_TRUE(GetXformMatrixFromPrim(cubePrim, cubeHydraMatrix));
+        ASSERT_TRUE(GfIsClose(cubeHydraMatrix.ExtractTranslation()[2], t, std::numeric_limits<double>::epsilon()));
```
ppt-adsk, line 46: Run through the animation and check that the z translation value changes correctly.

lib/flowViewport/CMakeLists.txt
```diff
@@ -16,7 +17,7 @@ target_sources(${TARGET_NAME}
 set(HEADERS
     api.h
     debugCodes.h
-    fvpUtils.h
```
debloip-adsk, line 19: I'm assuming we're removing it from the list so we don't ship it, as it's only used for a workaround?

debloip-adsk (dismissed):
Looks good, thanks! I've got two higher-level questions though :
1) Currently, the instruments utility is always used/run, even in release builds; is there not a risk that if we use it too much or for some expensive instrumentation, that it starts negatively affecting performance? Should we consider modifying it to only run in debug builds? This would cause problems for the unit test that depends on it though.
2) For the Python bindings, should we not do them now, or if we need to look into it, log a ticket to that effect?

lib/mayaHydra/hydraExtensions/sceneIndex/mhMayaUsdProxyShapeSceneIndex.cpp
debloip-adsk: I approved for now since I have no hard changes to request and this is still functional, I'll re-approve once updated

lib/flowViewport/CMakeLists.txt
ppt-adsk: Good catch!  Purely copy-paste damage, reverted.

lib/mayaHydra/hydraExtensions/sceneIndex/mhMayaUsdProxyShapeSceneIndex.cpp
ppt-adsk: Reverted debugging change.

ppt-adsk:
> Looks good, thanks! I've got two higher-level questions though :
> 
>     1. Currently, the instruments utility is always used/run, even in release builds; is there not a risk that if we use it too much or for some expensive instrumentation, that it starts negatively affecting performance? Should we consider modifying it to only run in debug builds? This would cause problems for the unit test that depends on it though.
> 
>     2. For the Python bindings, should we not do them now, or if we need to look into it, log a ticket to that effect?

1. There is a risk, but I think it's manageable, and we can assess on a case by case basis.  I think the gain in understanding the release build behavior outweighs the risk of poor use.
2. I would log a ticket, and wait until Pixar switches to pybind11 for their Python bindings.  This would allow us to use pybind11 across UFE (already using pybind11), mayaHydra, and USD bindings, and provide for easy interchange.  For now, using a plugin Maya command is an acceptable workaround, in my opinion.  Python binding work entered as HYDRA-989.

debloip-adsk (approved):

PR #123, opened 2024-05-01
HYDRA-976 - update build doc

lilike-adsk:
1.add build info for maya 2025
2.add build info for building with mayaUSD

### DIFF
--- a/doc/build.md
+++ b/doc/build.md
@@ -23,11 +23,9 @@ Before building the project, consult the following table to ensure you use the r
 
 ***NOTE:*** Visit the online Maya developer help document under ***Setting up your build environment*** for additional compiler requirements on different platforms.
 
-#### 2. Download and Build MayaUSD 
+#### 2. Download and Build Pixar USD 
 
-Started from Maya 2025, by default, the project needs to be built with MayaUSD, to build MayaUSD, see the github page https://github.com/Autodesk/maya-usd/blob/dev/doc/build.md
-
-***NOTE:*** If you want to build the project with Pixar USD instead of MayaUSD or you're building the project with Maya 2024, see Pixar's official github page for instructions on how to build USD: https://github.com/PixarAnimationStudios/USD. Pixar has removed support for building Maya USD libraries/plug-ins in their github repository and ```build_usd.py```.
+See Pixar's official github page for instructions on how to build USD: https://github.com/PixarAnimationStudios/USD. Pixar has removed support for building Maya USD libraries/plug-ins in their github repository and ```build_usd.py```.
 
 |               |      ![](images/pxr.png)          |USD version used in Maya        
 |:------------: |:---------------:                  |:---------------:
@@ -74,28 +72,27 @@ There are at least four arguments that must be passed to the script:
 | Flags                 | Description                                                                           |
 |--------------------   |-------------------------------------------------------------------------------------- |
 |   --maya-location     | directory where Maya is installed.                                                    |
-|  --mayausd-location   | directory where MayaUSD is installed.                                                 |
+|  --pxrusd-location    | directory where Pixar USD Core is installed.                                          |
 |  --devkit-location    | directory where Maya devkit is installed.                                             |
 | workspace_location    | directory where the project use as a workspace to build and install plugin/libraries  |
 
 Optional arguments are :
 | Flags                 | Description                                                                                         |
 |--------------------   |---------------------------------------------------------------------------------------------------- |
-|   --pxrausd-location  | directory where Pixar USD Core is installed.                                                        |
-|                       | Note if this argument instead of mayausd-location is provided, some features won't be available for |
+|   --mayausd-location  | directory where MayaUSD is installed. By providing this, you will enable more features for          |
 |                       | usd stages when using an hydra render delegate, such as : hide/delete the stage when the proxy shape|
 |                       | node is hidden/deleted, or applying a transform on the proxy shape node will apply it on the stage. |
 
 
 ```
 Linux:
-➜ maya-hydra python build.py --maya-location /usr/autodesk/maya2025 --mayausd-location /usr/local/mayausd --devkit-location /usr/local/devkitBase /usr/local/workspace
+➜ maya-hydra python build.py --maya-location /usr/autodesk/maya2025 --pxrusd-location /usr/local/USD-Release --devkit-location /usr/local/devkitBase /usr/local/workspace
 
 MacOSX:
-➜ maya-hydra python build.py --maya-location /Applications/Autodesk/maya2025 --mayausd-location /opt/local/mayausd --devkit-location /opt/local/devkitBase /opt/local/workspace
+➜ maya-hydra python build.py --maya-location /Applications/Autodesk/maya2025 --pxrusd-location /opt/local/USD-Release --devkit-location /opt/local/devkitBase /opt/local/workspace
 
 Windows:
-C:\maya-hydra> python build.py --maya-location "C:\Program Files\Autodesk\Maya2025" --mayausd-location C:\mayausd --devkit-location C:\devkitBase C:\workspace
+C:\maya-hydra> python build.py --maya-location "C:\Program Files\Autodesk\Maya2025" --pxrusd-location C:\USD-Release --devkit-location C:\devkitBase C:\workspace
 ```
 
 **Notes:** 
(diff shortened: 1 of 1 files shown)


lilike-adsk:
FYI  @roopavr-adsk @ppt-adsk @debloip-adsk @vlasovi

doc/build.md
```diff
@@ -72,27 +74,28 @@ There are at least four arguments that must be passed to the script:
...
-|  --pxrusd-location    | directory where Pixar USD Core is installed.                                          |
+|  --mayausd-location   | directory where MayaUSD is installed.                                                 |
 |  --devkit-location    | directory where Maya devkit is installed.                                             |
 | workspace_location    | directory where the project use as a workspace to build and install plugin/libraries  |
 
 Optional arguments are :
 | Flags                 | Description                                                                                         |
 |--------------------   |---------------------------------------------------------------------------------------------------- |
-|   --mayausd-location  | directory where MayaUSD is installed. By providing this, you will enable more features for          |
+|   --pxrausd-location  | directory where Pixar USD Core is installed.                                                        |
```
lanierd-adsk: --pxrausd-location  is the extra "a" on purpose ? And isn't this flag supposed to br removed now with mayausd being a requirement ?

lilike-adsk: right, it's pxrusd-location, I assume this argument is still there (in build.py), to give user an opportunity in case they want to build against USD. Do you prefer not mentioning this argument at all?

lanierd-adsk: I think we need to keep it to build with maya 2024, right ?

doc/build.md
```diff
@@ -11,25 +11,27 @@ Before building the project, consult the following table to ensure you use the r
...
 
 |        Optional       | ![](images/windows.png)   |                            ![](images/mac.png)               |   ![](images/linux.png)     |
 
 ***NOTE:*** Visit the online Maya developer help document under ***Setting up your build environment*** for additional compiler requirements on different platforms.
 
-#### 2. Download and Build Pixar USD 
+#### 2. Download and Build MayaUSD 
 
-See Pixar's official github page for instructions on how to build USD: https://github.com/PixarAnimationStudios/USD. Pixar has removed support for building Maya USD libraries/plug-ins in their github repository and ```build_usd.py```.
+Started from Maya 2025, by default, the project needs to be built with MayaUSD, to build MayaUSD, see the github page https://github.com/Autodesk/maya-usd/blob/dev/doc/build.md
```
ppt-adsk: "Starting from Maya 2025"

doc/build.md
lilike-adsk: Yes, that's one reason I kept the building with Pixar USD section.

doc/build.md
```diff
@@ -11,25 +11,27 @@ Before building the project, consult the following table to ensure you use the r
...
+
+***NOTE:*** If you want to build the project with Pixar USD instead of MayaUSD or you're building the project with Maya 2024, see Pixar's official github page for instructions on how to build USD: https://github.com/PixarAnimationStudios/USD. Pixar has removed support for building Maya USD libraries/plug-ins in their github repository and ```build_usd.py```.
```
ppt-adsk: This is somewhat misleading.  mayaUsd does not necessarily supply USD libraries.  You can very well build your own USD, then build your own mayaUsd using those USD libraries, then build mayaHydra using your mayaUsd and your USD.  Also, the Pixar removal of Maya-related libraries dates from many years ago (2018 or 2019, if I remember correctly), so I'm not sure it's useful to mention this here.  I feel it will cause more confusion that anything else.

doc/build.md
lanierd-adsk: Ok, so I think you only need to fix the typo --pxr**a**usd-location should be --pxrusd-location. Thanks.

ppt-adsk: This argument cannot be removed.  When you build mayaHydra you must specify the location of the USD libraries.

doc/build.md
```diff
@@ -72,27 +74,28 @@ There are at least four arguments that must be passed to the script:
...
+|  --mayausd-location   | directory where MayaUSD is installed.                                                 |
 |  --devkit-location    | directory where Maya devkit is installed.                                             |
 | workspace_location    | directory where the project use as a workspace to build and install plugin/libraries  |
 
 Optional arguments are :
 | Flags                 | Description                                                                                         |
 |--------------------   |---------------------------------------------------------------------------------------------------- |
-|   --mayausd-location  | directory where MayaUSD is installed. By providing this, you will enable more features for          |
+|   --pxrausd-location  | directory where Pixar USD Core is installed.                                                        |
+|                       | Note if this argument instead of mayausd-location is provided, some features won't be available for |
```
ppt-adsk: This argument does not replace mayausd-location, it is in addition to it.

doc/build.md
```diff
@@ -72,27 +74,28 @@ There are at least four arguments that must be passed to the script:
...
 |                       | usd stages when using an hydra render delegate, such as : hide/delete the stage when the proxy shape|
 |                       | node is hidden/deleted, or applying a transform on the proxy shape node will apply it on the stage. |
 
 
 ```
 Linux:
-➜ maya-hydra python build.py --maya-location /usr/autodesk/maya2024 --pxrusd-location /usr/local/USD-Release --devkit-location /usr/local/devkitBase /usr/local/workspace
+➜ maya-hydra python build.py --maya-location /usr/autodesk/maya2025 --mayausd-location /usr/local/mayausd --devkit-location /usr/local/devkitBase /usr/local/workspace
```
ppt-adsk: We can't omit --pxrusd-location, configuration will fail.

doc/build.md
lilike-adsk: Sure, will fix that, @ppt-adsk are you ok with this?

doc/build.md
```diff
@@ -192,7 +195,7 @@ Unit tests can be run by setting ```--stages=test``` or by simply calling `ctest
 
 ##### Python:
 
-It is important to use the Python version shipped with Maya and not the system version when building USD on MacOS. Note that this is primarily an issue on MacOS, where Maya's version of Python is likely to conflict with the version provided by the system. 
+***NOTE:*** If you want to build the project with Pixar USD instead of MayaUSD or you're building the project with Maya 2024, it is important to use the Python version shipped with Maya and not the system version when building USD on MacOS. Note that this is primarily an issue on MacOS, where Maya's version of Python is likely to conflict with the version provided by the system.
```
ppt-adsk: Pixar USD is always required.

ppt-adsk (changes requested):
The location of USD libraries must be specified with the --pxrusd-location build.py flag, or directly given to CMake.  See
https://github.com/Autodesk/maya-hydra/blob/fb478c845a0ebfdc3d386bb398684295b04d8928/cmake/modules/FindUSD.cmake#L34

doc/build.md
lilike-adsk: I thought usd could be retreived from MayaUSD, seems I was wrong, will update

doc/build.md
lilike-adsk: good catch! thanks @ppt-adsk !

lanierd-adsk (dismissed):

doc/build.md
```diff
@@ -29,25 +29,31 @@ See Pixar's official github page for instructions on how to build USD: https://g
...
+|  CommitID/Tags | [v22.11](https://github.com/PixarAnimationStudios/OpenUSD/releases/tag/v22.11) or [v23.08](https://github.com/PixarAnimationStudios/OpenUSD/releases/tag/v23.08) or [v23.11](https://github.com/PixarAnimationStudios/OpenUSD/releases/tag/v23.11) |Maya 2024 = v22.11<br>Maya 2025 = v23.11<br>Maya PR = v23.11
 
 For additional information on building Pixar USD, see the ***Additional Build Instruction*** section below.
 
 ***NOTE:*** Make sure that you don't have an older USD locations in your ```PATH``` and ```PYTHONPATH``` environment settings. ```PATH``` and ```PYTHONPATH``` are automatically adjusted inside the project to point to the correct USD location. See ```cmake/usd.cmake```.
 
-#### 3. Universal Front End (UFE)
+#### 3. Download and Build MayaUSD 
+
+Starting from Maya 2025, the project requires MayaUSD to build, this will enable more features for usd stages when using an hydra render delegate, such as : hide/delete the stage when the proxy shape node is hidden/deleted, or applying a transform on the proxy shape node will apply it on the stage.
```
ppt-adsk: Reads better as:
```suggestion
Starting from Maya 2025, the project requires MayaUSD to build.  This enables more features for USD stages when using a Hydra render delegate, such as: hide/delete the stage when the proxy shape node is hidden/deleted, or applying a transform on the proxy shape node will apply it on the stage. 
```

ppt-adsk (changes requested):
Looks good!  Just a minor style / grammar tweak and then we're done.

ppt-adsk (approved):

PR #131, opened 2024-05-27
HYDRA-312 : Add Scene Index for Texture Toggling

roopavr-adsk:


lib/flowViewport/sceneIndex/fvpPruneTexturesSceneIndex.cpp
```diff
@@ -0,0 +1,103 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#include "flowViewport/sceneIndex/fvpPruneTexturesSceneIndex.h"
+#include "pxr/base/tf/staticTokens.h"
```
lanierd-adsk: Nitpick : I think these should have the #include <pxr/base/tf/staticTokens.h> instead of the double quote.

lib/flowViewport/sceneIndex/fvpPruneTexturesSceneIndex.cpp
```diff
@@ -0,0 +1,103 @@
...
+    const TfTokenVector nodeNames = networkInterface->GetNodeNames();
+    for (TfToken const &nodeName : nodeNames) {
+        const TfToken nodeType = networkInterface->GetNodeType(nodeName);
+        if (nodeType == _tokens->ND_standard_surface_surfaceshader ||
+            nodeType == _tokens->UsdPreviewSurface) {                
+            // Look for incoming connection(textures) to surface shader params
+            TfTokenVector inputConnections = networkInterface->GetNodeInputConnectionNames(nodeName);
+            for (TfToken const &connection : inputConnections) {
+                // Trivially  input connections.
+                networkInterface->DeleteNodeInputConnection(nodeName, connection);
```
lanierd-adsk, line 51: How do you know it's a texture node that you delete ? Could it be another connection type like another node ? I think we don't support network graph yet but if that come in the future, I would double check that the input connection is a texture before deleting it.

lib/flowViewport/sceneIndex/fvpPruneTexturesSceneIndex.h
```diff
@@ -0,0 +1,56 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#ifndef FVP_PRUNE_TEXTURE_SCENE_INDEX_NEW_H
+#define FVP_PRUNE_TEXTURE_SCENE_INDEX_NEW_H
+
+#include "flowViewport/api.h"
+#include "flowViewport/sceneIndex/fvpSceneIndexUtils.h"
+
+#include "pxr/imaging/hdsi/api.h"
```
lanierd-adsk: Same here, you could use #include <> instead of double quotes.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1088,6 +1088,15 @@ MStatus MtohRenderOverride::Render(
        _displayStyleSceneIndex->SetRefineLevel({true, delegateParams.refineLevel});
     }
 
+    // Toggle textures in the material network
+    const unsigned int currentTextureMode = drawContext.getDisplayStyle();
```
lanierd-adsk: I think the variable is wrongly named here. It's not the currentTextureMode but the currentDisplayStyle which contains a bit about the texture mode. Could you please renamed it ?

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1436,6 +1447,10 @@ void MtohRenderOverride::_CreateSceneIndicesChainAfterMergingSceneIndex(const MH
             Fvp::DisplayStyleOverrideSceneIndex::New(_inputSceneIndexOfFilteringSceneIndicesChain);
     _displayStyleSceneIndex->addExcludedSceneRoot(MAYA_NATIVE_ROOT); // Maya native prims don't use global refinement
 
+    // Add texture disabling Scene Index
+    _lastFilteringSceneIndexBeforeCustomFiltering = _pruneTexturesSceneIndex = 
+    Fvp::PruneTexturesSceneIndex::New(_lastFilteringSceneIndexBeforeCustomFiltering);
```
lanierd-adsk, line 1452: I think you should create the pruneTextureSceneIndex only when the texture are off in the viewport. And not put it in the scene index chain when textures are on.
Always creating this scene index and having it not doing much when textures are on is a performance concern. 
It will always still do things with the empty lambda or be an indirection for GetPrim and will slow down the pipeline.

lanierd-adsk (changes requested):

lib/flowViewport/sceneIndex/fvpPruneTexturesSceneIndex.cpp
roopavr-adsk: Maya's VP2 texture toggle behavior is such that it disables *any* incoming connection to the surface shader. Whether the incoming connection is a file texture node or procedural node(ramp/checker etc) or any other. So the toggling behaviour should disable any input connections based on Maya's native behavior. What do you think?

lib/mayaHydra/mayaPlugin/renderOverride.cpp
roopavr-adsk: I had it the same way earlier, but that meant the entire scene prims get removed and added on every toggle. This could be very expensive. Currently, I am only dirtying the prims on toggle which I think is less expensive. The method MarkTexturesDirty() is invoked on toggling. Also, the scene index is created only once during initialization now. I have made changes in _NeedToRecreateTheSceneIndicesChain() where texture toggling condition is not accounted for anymore. Perhaps we could handle this more elegantly when we have a new mechanism via FVP API to handle dirtying.

lanierd-adsk: The fact that the entire scene prims are removed and added again is logged in a separate defect. Let's talk about this offline.

lanierd-adsk:
@ppt-adsk @vlasovi @lilike-adsk @debloip-adsk

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lanierd-adsk: We will log a new defect to make coherent the fact we use scene indices that are always present and scene indices created when needed.

lanierd-adsk (approved):

PR #132, opened 2024-06-04
HYDRA-845 : Selection highlighting for point instancing of meshes

debloip-adsk:
This PR implements point instance selection highlighting of meshes for point instancer, instance & prototype selections.

Note that as part of this PR,  the `Fvp::PathInterface::SceneIndexPath` method was renamed to `ConvertUfePathToHydraSelections`, and now requires returning a vector of a new struct `PrimSelectionInfo`, each containing an `SdfPath` and a selection data source.

Following this, the lead object path tracker now also supports multiple lead prim paths.


debloip-adsk (commented):
Not much to add that isn't already in the Markdown doc or in code comments, except for the following two questions : 
- I've renamed `SceneIndexPath` to `ConvertUfePathToHydraSelections`, but would we prefer keeping both? Since sometimes we may not really care about the selections themselves. Note that different `Ufe::Path`s could end up mapping to the same `SdfPath`s though, for example in the case of point instancer vs. instance selection, where the `Ufe::Path`s would differ but the `SdfPath`s would both end up pointing to the instancer.
- Should we rename `Fvp::PathInterface`, since it's not strictly about paths anymore? I would think how we name it also depends on if we keep `SceneIndexPath` or not, and what other functionality we could end up putting in it.

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.cpp
```diff
@@ -65,6 +65,15 @@ class GlobalSelectionChangedObs : public Ufe::Observer
             MayaHydra::MhLeadObjectPathTracker& _leadObjectPathTracker;
     };
 
+void pushHydraSelectionsToPathVector(const Fvp::PrimSelectionInfoVector& hydraSelections, SdfPathVector& outPathVector)
```
ppt-adsk: This doesn't seem to be used to append to outPathVector, which is always empty, as far as I can tell.  Would be clearer (and allow for const safeness) if you just returned the SdfPathVector.  Return value optimization or move construction will take care of efficiency.
```suggestion
SdfPathVector hydraSelectionsPaths(const Fvp::PrimSelectionInfoVector& hydraSelections)
```

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.h
```diff
@@ -56,11 +56,11 @@ class MhLeadObjectPathTracker
     void setLeadObjectUfePath(const Ufe::Path& newLeadObjectUfePath);
 
     MAYAHYDRALIB_API
-    void updatePrimPath(); // For example : this is called after the data producer scene indices are loaded
+    void updatePrimPaths(); // For example : this is called after the data producer scene indices are loaded
 
 private:
     const Fvp::PathInterface* const _pathInterface {nullptr};
-    PXR_NS::SdfPath                 _leadObjectPrimPath;
+    PXR_NS::SdfPathVector           _leadObjectPrimPaths;
```
ppt-adsk, line 63: If we're renaming `isLeadObject()` to `isLeadPrim()`, perhaps this should be `_leadPrimPaths`?

lib/flowViewport/sceneIndex/fvpSelectionSceneIndex.h
```diff
@@ -117,7 +117,7 @@ class SelectionSceneIndex final
     //! and warns about empty return paths.
     //@{
     FVP_API
-    PXR_NS::SdfPath SceneIndexPath(const Ufe::Path& appPath) const override;
```
ppt-adsk, line 120: Feels like this method should be kept as a non-virtual concrete utility, for clients that can assert that a single scene index path will be returned.  That way, the assertion can be done inside Fvp::SelectionSceneIndex::SceneIndexPath().  I would throw some kind of custom assertion derived from `std::runtime_error` if there are multiple paths.

This would cut down on a few changes in your branch.

lib/flowViewport/sceneIndex/fvpSelectionSceneIndex.cpp
```diff
@@ -219,21 +228,22 @@ void SelectionSceneIndex::ReplaceSelection(const Ufe::Selection& selection)
...
     _selection->Clear();
 
-    SdfPathVector sceneIndexSn;
+    PrimSelectionInfoVector sceneIndexSn;
     sceneIndexSn.reserve(selection.size());
     for (const auto& snItem : selection) {
-        // Call our input scene index to convert the application path to a scene
-        // index path.
-        auto sceneIndexPath = SceneIndexPath(snItem->path());
+        // Call our input scene index to convert the application path to scene index paths and selection data sources.
```
ppt-adsk, line 234: Should standardize on naming: either use scene index path or Hydra path, but not both.  Should try to follow Pixar's convention on this.

doc/selectionHighlightingArchitecture.md
```diff
@@ -287,3 +289,183 @@ WireframeSelectionHighlightSceneIndex o-- Selection : Read
...
+|  |__NestedInstancer
+|  |  |__Prototype
+|  |  |__Prototype_SelectionHighlight
+|  |__NestedInstancer_SelectionHighlight
+|__TopInstancer_SelectionHighlight
+```
+where `TopInstancer_SelectionHighlight` would instance `NestedInstancer_SelectionHighlight`, which would in turn instance `Prototype_SelectionHighlight`.
+
+Note that in the case where a prototype is not a single prim but a sub-hierarchy, we only need to 
+create a single *explicit* selection highlight mirror for the whole prototype sub-hierarchy; the
```
ppt-adsk: For clarity:
"selection highlight mirror prim for the whole prototype subhierarchy;"

doc/selectionHighlightingArchitecture.md
```diff
@@ -287,3 +289,183 @@ WireframeSelectionHighlightSceneIndex o-- Selection : Read
...
+|  |__Prototype_SelectionHighlight
+|     |__ChildInstancer
+|__TopInstancer_SelectionHighlight
+```
+where `TopInstancer_SelectionHighlight` instances `Prototype_SelectionHighlight`, which implicitly draws the selection highlight version of `ChildInstancer`.
+
+Of note are the following selection highlighting scenarios and their corresponding behaviors :
+- Selecting a point instancer in its entirety
+  - If the instancer is a top-level instancer, all instances it draws WILL be highlighted
+  - If the instancer is a prototype, instances of itself drawn by other instancers will NOT be highlighted
```
ppt-adsk: "NOT be highlighted.  This is an intentional workflow decision from the Hydra for Maya team."

lib/flowViewport/sceneIndex/fvpPathInterface.h
```diff
@@ -20,21 +20,31 @@
...
 class Path;
 }
 
-PXR_NAMESPACE_OPEN_SCOPE
-class SdfPath;
-PXR_NAMESPACE_CLOSE_SCOPE
-
 namespace FVP_NS_DEF {
 
+struct PrimSelectionInfo
```
ppt-adsk: Naming feels longish...  SelectedPrim?  PrimSelection?  I'm O.K. with PrimSelectionInfo too.

lib/flowViewport/sceneIndex/fvpPathInterface.h
```diff
@@ -20,21 +20,31 @@
...
+{
+    PXR_NS::SdfPath primPath;
+    PXR_NS::HdDataSourceBaseHandle selectionDataSource;
+};
+
+// Using TfSmallVector to optimize for selections that map to a few prims,
+// which is likely going to be the bulk of use cases.
+using PrimSelectionInfoVector = PXR_NS::TfSmallVector<PrimSelectionInfo, 8>;
```
ppt-adsk: If we went with SelectedPrim or PrimSelection this could be SelectedPrims or PrimSelections...  And we'd avoid being specific on the container type.

lib/flowViewport/sceneIndex/fvpPathInterface.h
```diff
@@ -45,7 +55,7 @@ class PathInterface
     //! If no such path exists, an empty SdfPath should be returned.
     //! \return scene index path.
     FVP_API
-    virtual PXR_NS::SdfPath SceneIndexPath(const Ufe::Path& appPath) const = 0;
```
ppt-adsk, line 48: As per other comment I think we should keep this as a non-virtual utility method that calls the multiple selection paths one and asserts there's a single element in it.

lib/flowViewport/sceneIndex/fvpPathInterface.h
```diff
@@ -45,7 +55,7 @@ class PathInterface
...
+    virtual PrimSelectionInfoVector ConvertUfePathToHydraSelections(const Ufe::Path& appPath) const = 0;
```
ppt-adsk: Naming suggestion: I'd drop the Convert prefix, and go with UfePathToSelectedPrims() or UfePathToPrimSelections(), or similar.

test/lib/mayaUsd/render/mayaToHydra/cpp/testPointInstancingWireframeHighlight.cpp
```diff
@@ -0,0 +1,365 @@
...
+#include <gtest/gtest.h>
+
+#include <algorithm>
+
+PXR_NAMESPACE_USING_DIRECTIVE
+using namespace MayaHydra;
+
+namespace {
+
+const std::string selectionHighlightTag = "_SelectionHighlight";
```
ppt-adsk: Shouldn't we be able to recover this from the API?  E.g. from Fvp::PathInterface, or perhaps Fvp::WireframeSelectionHighlightingSceneIndex?

test/lib/mayaUsd/render/mayaToHydra/cpp/testPointInstancingWireframeHighlight.cpp
```diff
@@ -0,0 +1,365 @@
...
+    
+    testInstancerDirectHighlightFn(topInstancerItem, topInstancerPath);
+    testInstancerDirectHighlightFn(secondInstancerItem, secondInstancerPath);
+    testInstancerDirectHighlightFn(thirdInstancerItem, thirdInstancerPath);
+    testInstancerDirectHighlightFn(fourthInstancerItem, fourthInstancerPath);
+    
+    // Select point instancer ancestors
+    auto testInstancerIndirectHighlightFn = [&](const Ufe::SceneItem::Ptr& instancerItem, const Ufe::Path& instancerPath) -> void {
+        // This is not an actual selection, we use it to get the Hydra path
+        auto instancerHydraSelections = fvpMergingSceneIndex->ConvertUfePathToHydraSelections(instancerPath);
```
ppt-adsk: Is this saying that the method naming should be clearer?  Or call the SceneIndexPath() method in this case?

lib/mayaHydra/hydraExtensions/sceneIndex/registration.cpp
```diff
@@ -121,15 +127,77 @@ class PathInterfaceSceneIndex : public Fvp::PathInterfaceSceneIndexBase
...
+            if (pathComponent.string().find_first_not_of(digits) == std::string::npos) {
+                // This should only occur on the last component, when we have an instance selection
+                TF_VERIFY(pathComponent == secondSegment.components().back());
+                continue;
+            }
+            primPath = primPath.AppendChild(TfToken(pathComponent.string()));
+        }
+
+        const auto lastComponentString = secondSegment.components().back().string();
+        bool lastComponentIsNumeric = lastComponentString.find_first_not_of(digits) == std::string::npos;
```
ppt-adsk: Couldn't you just do that at line 138?  You just made that check.

lib/mayaHydra/hydraExtensions/sceneIndex/registration.cpp
```diff
@@ -121,15 +127,77 @@ class PathInterfaceSceneIndex : public Fvp::PathInterfaceSceneIndexBase
...
+            HdInstanceIndicesSchema::Builder instanceIndicesBuilder;
+            instanceIndicesBuilder.SetInstancer(HdRetainedTypedSampledDataSource<SdfPath>::New(primPath));
+            instanceIndicesBuilder.SetInstanceIndices(HdRetainedTypedSampledDataSource<VtArray<int>>::New({std::stoi(lastComponentString)}));
+            HdSelectionSchema::Builder selectionBuilder;
+            // Instancer is expected to be marked "fully selected" even if only certain instances are selected,
+            // based on USD's _AddToSelection function in selectionSceneIndexObserver.cpp :
+            // https://github.com/PixarAnimationStudios/OpenUSD/blob/f7b8a021ce3d13f91a0211acf8a64a8b780524df/pxr/imaging/hdx/selectionSceneIndexObserver.cpp#L212-L251
+            selectionBuilder.SetFullySelected(HdRetainedTypedSampledDataSource<bool>::New(true));
+            auto instanceIndicesDataSource = HdDataSourceBase::Cast(instanceIndicesBuilder.Build());
+            selectionBuilder.SetNestedInstanceIndices(HdRetainedSmallVectorDataSource::New(1, &instanceIndicesDataSource));
```
ppt-adsk: Maybe factor this out into a separate unnamed namespace function, for readability?

lib/flowViewport/sceneIndex/fvpWireframeSelectionHighlightSceneIndex.h
```diff
@@ -70,6 +72,11 @@ class WireframeSelectionHighlightSceneIndex
     FVP_API
     void addExcludedSceneRoot(const PXR_NS::SdfPath& sceneRoot);
 
+    // Returns the corresponding selection highlight mirror path, if one exists.
+    // Otherwise, returns the given path as-is.
+    FVP_API
+    PXR_NS::SdfPath GetSelectionHighlightPath(const PXR_NS::SdfPath& path) const;
```
ppt-adsk, line 82: Naming: other methods in this class have mirror in their name, feel this one should too.

lib/flowViewport/sceneIndex/fvpWireframeSelectionHighlightSceneIndex.cpp
```diff
@@ -50,6 +65,314 @@ const HdDataSourceLocator reprSelectorLocator(
...
+VtBoolArray _GetSelectionHighlightMask(const HdInstancerTopologySchema& originalInstancerTopology, const HdSelectionsSchema& selections) 
+{
+    // Schema getters were made const in USD 24.05 (specifically Hydra API version 66).
+    // We work around this for previous versions by const casting.
+    VtBoolArray originalMask = 
+#if HD_API_VERSION < 66
+    const_cast<HdInstancerTopologySchema&>(originalInstancerTopology).GetMask()->GetTypedValue(0);
+#else
+    originalInstancerTopology.GetMask()->GetTypedValue(0);
+#endif
```
ppt-adsk, line 94: Maybe factor this out in an unnamed namespace function?
```suggestion
inline HdInstancerTopologySchema& topologySchema(HdInstancerTopologySchema& t)
#if HD_API_VERSION < 66
     return const_cast<HdInstancerTopologySchema&>(t);
#else
    return t;
#endif
```

lib/flowViewport/sceneIndex/fvpWireframeSelectionHighlightSceneIndex.cpp
```diff
@@ -163,31 +552,92 @@ WireframeSelectionHighlightSceneIndex::_PrimsDirtied(
...
+        if (selectionHighlightPath != entry.primPath) {
+            dirtiedPrims.emplace_back(selectionHighlightPath, entry.dirtyLocators);
+        }
+
+        if (entry.dirtyLocators.Intersects(HdInstancerTopologySchema::GetDefaultLocator())
+            && _selectionHighlightMirrorsByInstancer.find(entry.primPath) != _selectionHighlightMirrorsByInstancer.end()) {
+            // An instancer with a selection highlight was changed; rebuild its selection highlight.
+            // We do not need to check for instancedBy dirtying : if an instancedBy data source is dirtied,
+            // then either a new instancer was added, which will be handled in _PrimsAdded, either an
+            // existing instancer's instancerTopology data source was dirtied., which is handled here.
```
ppt-adsk: Sentence seems to need a bit of work.

doc/selectionHighlightingArchitecture.md
```diff
@@ -287,3 +289,183 @@ WireframeSelectionHighlightSceneIndex o-- Selection : Read
...
+itself, and will be instance-drawn through the parent instancer. However, if the prototyped
+instancer has no parent instancer, but it is instanced by another instancer somewhere else
+in the hierarchy, then both the prototyped instancer will draw as if it were by itself, but
+also be instance-drawn by the other instancer.
+
+This nesting and composition of instancers is what leads to most of the complexity of point
+instancing selection highlighting. We can view such nesting and composition of instancers as 
+graphs, with the vertices being the instancer prims, and the edges being the paths contained 
+in the `instancerTopology/prototypes`, `instancedBy/paths` and `instancedBy/prototypeRoots` 
+data sources.
```
ppt-adsk: It would be nice to show a simple graph with this idea, I think it's a very useful mental model.

ppt-adsk (commented):
Fantastic work, pretty impressive how much additional complexity all this tracking requires...  Minor comments from me in this pull request.

As for the path interface, I'm still uncertain as to whether it's a selection-centric class, or a path mapper class?  Is there another application besides selection that we don't yet know about?  In the end though, this kind of fundamental questioning is outside the scope of this pull request, so we have time to ponder this.

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.cpp
debloip-adsk: Removed it and used the newly created `SceneIndexPaths` method instead

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.h
debloip-adsk: I went with renaming the method to `isLeadObjectPrim` instead, since everything in the code refers to a lead object (e.g. `leadObjectUfePath` instead of `leadUfePath`), which makes sense to me if the "lead object" is the underlying abstract concept, and UFE paths and prims are just materializations of that concept

lib/flowViewport/sceneIndex/fvpSelectionSceneIndex.h
debloip-adsk: Done, and also added a "multi-paths" analogue, `SceneIndexPaths`, which essentially returns `UfePathToPrimSelections`, but with just the paths and not the selection data sources.

lib/flowViewport/sceneIndex/fvpSelectionSceneIndex.cpp
debloip-adsk: I'm not sure what you mean, do you mean using scene index path vs. prim path? There are no mentions of Hydra paths, there were mentions of Hydra selections, but I've changed these to say prim selections instead.

doc/selectionHighlightingArchitecture.md
debloip-adsk: Done

doc/selectionHighlightingArchitecture.md
debloip-adsk: Done

lib/flowViewport/sceneIndex/fvpPathInterface.h
debloip-adsk: Renamed to `PrimSelection`

lib/flowViewport/sceneIndex/fvpPathInterface.h
debloip-adsk: Renamed to `PrimSelections`

lib/flowViewport/sceneIndex/fvpPathInterface.h
debloip-adsk: Done, also added `SceneIndexPaths` method as described in more detail in other comment

lib/flowViewport/sceneIndex/fvpPathInterface.h
debloip-adsk: Went with `UfePathToPrimSelections`

test/lib/mayaUsd/render/mayaToHydra/cpp/testPointInstancingWireframeHighlight.cpp
debloip-adsk: Added it to `Fvp::WireframeSelectionHighlightingSceneIndex`. I don't feel like it fits too much in `Fvp::PathInterface`, in my mind this would ideally be an implementation detail (though obviously it is visible through the scene index so it's not), as really its only purpose is to allow the creation of mirror prims at the same hierarchical level as their original counterparts (i.e. as siblings)

test/lib/mayaUsd/render/mayaToHydra/cpp/testPointInstancingWireframeHighlight.cpp
debloip-adsk: Changed to call the new `SceneIndexPaths` method

lib/mayaHydra/hydraExtensions/sceneIndex/registration.cpp
debloip-adsk: Refactored slightly to set this boolean within the loop when doing the first check

lib/mayaHydra/hydraExtensions/sceneIndex/registration.cpp
debloip-adsk: Done

lib/flowViewport/sceneIndex/fvpWireframeSelectionHighlightSceneIndex.h
debloip-adsk: I personally wouldn't, as this method does not necessarily require having mirrors in the first place. If you called this method in a plain-old basic hierarchy with no instancers, it would return you the original path, as it would indeed be the one to highlight, but it would not be a mirror path.

lib/flowViewport/sceneIndex/fvpWireframeSelectionHighlightSceneIndex.cpp
debloip-adsk: I'm not sure I see what value this brings. If the method takes in a const reference, and does whatever treatment it needs to return a non-const reference, isn't that just a more complicated way of doing a flat-out `const_cast`? And if it takes in a non-const reference, then there would be no need to use `const_cast` in the first place? To me this preprocessor bit is mostly to confine the const_casting to the old Hydra versions, and use the new ones for the current & future (which would also allow us to remove the old API usage at a future point should we wish to do so)

lib/flowViewport/sceneIndex/fvpWireframeSelectionHighlightSceneIndex.cpp
debloip-adsk: Fixed and reformatted slightly

doc/selectionHighlightingArchitecture.md
debloip-adsk: Added a graph; not sure that it is super simple though, but it does show pretty much all the possibilities.

ppt-adsk: Fantastic, love it, in my opinion this brings a lot of clarity, thanks for doing this!

lib/flowViewport/sceneIndex/fvpWireframeSelectionHighlightSceneIndex.cpp
ppt-adsk: It was just a suggestion for readability and being closer to Don't Repeat Yourself.  You have the const_cast ifdef twice, and inside the ifdef you're repeating code.  No big deal though.

lib/flowViewport/sceneIndex/fvpWireframeSelectionHighlightSceneIndex.h
ppt-adsk: Great point, agreed.

lib/flowViewport/sceneIndex/fvpSelectionSceneIndex.cpp
ppt-adsk: Sounds good to me.

ppt-adsk (approved):
Thanks for the update!

debloip-adsk:
FYI @lanierd-adsk @lilike-adsk @roopavr-adsk @vlasovi

PR #135, opened 2024-06-10
HYDRA-553 : Fix for aiSkyDomeLight bugs

roopavr-adsk:
This PR contain fixes for bugs with aiSkyDomeLight when using Arnold Render Delegate.

1. Fixes black renders when the dome light is selected: This was caused by a textureSkyDome mesh which is produced for VP2. It should not be translated to Hydra as it has it own mechanims to display dome light in the viewport.
2. Fixes light param updates not working: incorrect token were being compared.

### DIFF
--- a/lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
+++ b/lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
@@ -183,6 +183,16 @@ namespace {
         return mayaPath;
     }
 
+    bool IsTexturedSkyDomeRenderItem(const SdfPath& name)
+    {
+        static const std::string _aiTexturedSkyDome("texturedSkyDome");
+        if (name.GetString().find(_aiTexturedSkyDome) != std::string::npos) {
+            // This render item is an texturedDomeLighnt
+            return true;
+        }
+        return false;
+    }
+
     SdfPath _GetPrimPath(const SdfPath& base, const MDagPath& dg)
     {
         return base.AppendPath(GetMayaPrimPath(dg));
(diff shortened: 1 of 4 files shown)


lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -183,6 +183,16 @@ namespace {
         return mayaPath;
     }
 
+    bool IsTexturedSkyDomeRenderItem(const SdfPath& name)
+    {
+        static const std::string _aiTexturedSkyDome("texturedSkyDome");
+        if (name.GetString().find(_aiTexturedSkyDome) != std::string::npos) {
```
lanierd-adsk, line 189: I am not sure this is enough as they could be named another way that texturedSkyDome, right ?
The name probably comes from the maya node which can be named anything if I am correct ?

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -474,7 +484,10 @@ void MayaHydraSceneIndex::HandleCompleteViewportScene(const MDataServerOperation
         MayaHydraRenderItemAdapterPtr ria = nullptr;
         if (!_GetRenderItem(fastId, ria)) {
             const SdfPath slowId = _GetRenderItemPrimPath(ri);
-            if (slowId.IsEmpty()) {
+
+            // Maya/MtoA adds texturedSkyDome mesh object for VP2.
+            // We do not want that to be translated to Hydra
```
lanierd-adsk, line 489: For Arnold, we will use mtoa, that is true, but what about HdStorm ? We still want to use the skydeom under storm right ?

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
roopavr-adsk: To my understanding, this is a OGS generated render item with a hardcoded name. OGS generates RI with certain naming conventions:
For textured dome mesh render item:
/MayaHydraViewportRenderer/rprims/aiSkyDomeLight1/aiSkyDomeLightShape1/texturedSkyDome_941
For wireframe render item:
/MayaHydraViewportRenderer/rprims/aiSkyDomeLight1/aiSkyDomeLightShape1/active_wireframeSkyDome_9417

I tested renaming the domelight and suffix "../../texturedSkyDome_xxx" remains.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
roopavr-adsk: We do not need the skydome mesh for Storm or Arnold. Both have their own representation. The mesh render item was possibly created for VP2 display, which we do not need.

lanierd-adsk:
ok, thanks for verification. I will validate the code review when you refresh the request.

lanierd-adsk (approved):

PR #143, opened 2024-07-05
HYDRA-1058/HYDRA-970 : GeomSubsets picking filter + Fix multi-selection when single-picking

debloip-adsk:
This PR adds GeomSubsets picking functionality, and extends the "USD Selection Mode" menu with a GeomSubsets section. Bundled with it is a fix for the HYDRA-970 issue where doing a single click could select more than one prim (the fix was required to properly implement the following behavior from HYDRA-1058 : "A single pick selects either a GeomSubset, **_or if the picked object is not a GeomSubset, the parent prim is selected_**").


lib/mayaHydra/mayaPlugin/plugin.cpp
```diff
@@ -132,19 +212,23 @@ PLUGIN_EXPORT MStatus uninitializePlugin(MObject obj)
...
+            delete _renderOverrides[i];
         }
     }
 
-    gsRenderOverrides.clear();
+    _renderOverrides.clear();
 
     // Clear any registered callbacks
-    MGlobal::executeCommand("callbacks -cc mayaHydra;");
+    MGlobal::executeCommand("callbacks -cc -owner mayaHydra;");
```
debloip-adsk, line 231: Fixes the following error that was raised when trying to unload mayaHydra : `When clearing callbacks, the owner must be specified.` Note that mayaHydra still doesn't unload properly (and already didn't before this PR, even with this fix), but at least that's one issue down.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -255,14 +256,27 @@ UsdPointInstancesPickMode GetPointInstancesPickMode()
...
         const HdxPickHit&                pickHitArg, 
-        const MHWRender::MSelectionInfo& pickInfoArg
-    ) : pickHit(pickHitArg), pickInfo(pickInfoArg) {}
+        const MHWRender::MSelectionInfo& pickInfoArg,
+        const bool                       isSolePickHit
+    ) : pickHit(pickHitArg), pickInfo(pickInfoArg), isSolePickHit(isSolePickHit) {}
 
     const HdxPickHit&                pickHit;
     const MHWRender::MSelectionInfo& pickInfo;
+    const bool                       isSolePickHit;
```
debloip-adsk, line 279: Used to differentiate/handle the following behaviors from HYDRA-1058 : 
- Marquee select only selects GeomSubsets, never the parent prims
- A single pick selects either a GeomSubset, or if the picked object is not a GeomSubset, the parent prim is selected

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -450,15 +464,54 @@ class UsdPickHandler : public MtohRenderOverride::PickHandlerBase {
 
     UsdPickHandler(MtohRenderOverride& renderOverride) : 
         PickHandlerBase(renderOverride) {}
+    
+#if PXR_VERSION >= 2403
```
debloip-adsk, line 468: GeomSubset Hydra prims were only added in USD 24.03

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -510,56 +563,82 @@ class UsdPickHandler : public MtohRenderOverride::PickHandlerBase {
...
-                const SdfPath& pickedUsdPath, int instanceNdx) {
-
-            if (instanceNdx >= 0) {
-                // Point instance: add the instance index to the path.
-                // Appending a numeric component to the path to identify a
-                // point instance cannot be done on the picked SdfPath, as
-                // numeric path components are not allowed by SdfPath.  Do so
-                // here with Ufe::Path, which has no such restriction.
-                return pickedMayaPath + std::to_string(instanceNdx);
+        std::vector<HitPath> hitPaths;
```
debloip-adsk, line 566: We now handle multiple paths, since a given component (e.g. face) of a mesh can be in more than one GeomSubset.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1781,9 +1862,17 @@ void MtohRenderOverride::_PickByRegion(
...
     // Use the same size as selection region is enough to get all pick results.
     pickParams.resolution.Set(sel_w, sel_h);
+    pickParams.pickTarget = HdxPickTokens->pickPrimsAndInstances;
+    pickParams.resolveMode = singlePick ? HdxPickTokens->resolveNearestToCenter : HdxPickTokens->resolveUnique;
+    pickParams.doUnpickablesOcclude = false;
     pickParams.viewMatrix.Set(viewMatrix.matrix);
     pickParams.projectionMatrix.Set(adjustedProjMatrix.matrix);
-    pickParams.resolveMode = HdxPickTokens->resolveUnique;
+    pickParams.collection = _renderCollection;
+    pickParams.outHits = &outHits;
```
debloip-adsk, line 1871: - Reorganized the params configuration to explicitly set the default values first, and then override them based on special behaviors afterwards. 
- The `resolveMode` is now different if `singlePick` is true : the `resolveNearestToCenter` mode will only return the prim that is closest to the center of the selection. This fixes HYDRA-970.
- Explicitly set `doUnpickablesOcclude` to `false`, as it was the default value and our existing behavior, to avoid unforeseen surprises if the default value ever changes.

lib/mayaHydra/mayaPlugin/viewCommand.cpp
```diff
@@ -303,6 +310,8 @@ MStatus MtohViewCmd::doIt(const MArgList& args)
         setResult(MhBuildInfo::gitBranch());
     } else if (db.isFlagSet(_buildDate)) {
         setResult(MhBuildInfo::buildDate());
+    } else if (db.isFlagSet(_usdVersion)) {
+        setResult(PXR_VERSION);
```
debloip-adsk, line 314: We use `PXR_VERSION` to conditionally compile the GeomSubsets pick resolution code; since the feature can't be made available when building with earlier USD versions, we also need to not show the GeomSubsets pick filter UI in such cases. However, we had no way of determining what USD version the plugin was built with from MEL/Python; this added flag to the `mayaHydra` command does just that.

test/lib/mayaUsd/render/mayaToHydra/cpp/testUsdPointInstancePicking.cpp
```diff
@@ -65,9 +65,6 @@ TEST(TestUsdPointInstancePicking, pickPointInstance)
     mouseClick(Qt::MouseButton::LeftButton, active3dView.widget(), primMouseCoords);
     active3dView.refresh();
 
-    // When picking on the boundary of multiple objects, one Hydra pick hit per
-    // object is returned.  Therefore test that the expected selected path is
-    // in the selection.
-    ASSERT_GE(sn->size(), 1u);
+    ASSERT_EQ(sn->size(), 1u);
```
debloip-adsk, line 68: Adjusted since we should no longer have more than one prim selected when single-clicking.

test/lib/mayaUsd/render/mayaToHydra/cpp/testUsdPointInstancePicking.py
```diff
@@ -40,8 +40,8 @@ def loadUsdScene(self):
     def setUp(self):
         super(TestUsdPointInstancePicking, self).setUp()
         self.loadUsdScene()
-        cmds.setAttr('persp.translate', 19.3, 13.7, 11.4, type='float3')
-        cmds.setAttr('persp.rotate', -33.4, 63.0, 0, type='float3')
+        cmds.setAttr('persp.translate', 15, 15, 10, type='float3')
+        cmds.setAttr('persp.rotate', -40, 60, 0, type='float3')
```
debloip-adsk, line 44: One of the tests previously clicked on the boundary between two prims, but was closer to the wrong prim than the correct one, and with the fix for HYDRA-970, this made the test fail. Adjusted the camera angle slightly so that the mouse click is on the correct prim.

test/testSamples/testGeomSubsetsPicking/GeomSubsetsPickingTestScene.usda
```diff
@@ -0,0 +1,96 @@
...
+    {
+        double3 xformOp:translate = (2, 0.5, 0)
+        uniform token[] xformOpOrder = ["xformOp:translate"]
+    }
+
+    def Xform "SphereInstanceLowerHalfMarker"
+    {
+        double3 xformOp:translate = (2, -0.5, 0)
+        uniform token[] xformOpOrder = ["xformOp:translate"]
+    }
```
debloip-adsk, line 80: Markers are used for pointing the test mouse clicks at the right spots, as by default just using a prim's transform will point to its center, but we want to click at slightly different offsets from the center, and we can't use screen-space offsets, since viewport sizes can differ.

lanierd-adsk (approved):

PR #147, opened 2024-07-11
HYDRA-1103 : Update face selection shader to behave as an unlit shader

roopavr-adsk:


lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -693,9 +693,17 @@ VtValue MayaHydraSceneIndex::CreateMayaFacesSelectionMaterial()
...
     node.path = _mayaFacesSelectionMaterialPath;
+
+    // Diffuse
     node.parameters.insert(
         { _tokens->diffuseColor,
-          VtValue(GfVec3f(faceSelectioncolor[0], faceSelectioncolor[1], faceSelectioncolor[2])) });
+          VtValue(GfVec3f(faceSelectioncolor[0], faceSelectioncolor[1], faceSelectioncolor[2])*0.3f) });
+    
+    // Emissive (component selection highlighting material should be independent of scene lighting)
+    node.parameters.insert(
```
lanierd-adsk, line 703: This will make the shader be visible when there is no lights, but I think the problem was also that it should not receive lighting at all, I don't think this resolves this problem, right ?

roopavr-adsk: The face selection should be visible even when there are no lights. Isn't that the required behavior? Selection highlighting should be independent of scene lighting. Maya VP2 face sel highlight is visible even when no scene lights are used(Lighting>Use No Lights)

roopavr-adsk:
FYI : @lilike @tremblp @debloip @vlasovi @debloip-adsk

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
ppt-adsk: The 0.3 magic constant seems to appear in 3 places, lines 700, 705, and 707.  Any reason for that choice?  And why not factor it out?

roopavr-adsk: The magic constant is based on VP2's shader behavior. Good idea to factor it out.

ppt-adsk (approved):

PR #148, opened 2024-07-16
HYDRA-1057 : GeomSubsets wireframe highlighting

debloip-adsk:
This PR adds selection highlighting for GeomSubsets.

A brief overview of the thought process behind the implementation : previously, non-instanced meshes were highlighted through the `refinedWireOnSurf` display style. However, this does not allow us to create a wireframe for a subset of the mesh. So the idea is to instead use the existing architecture that maintains a selection highlight mirror hierarchy, and have a mirror prim of the mesh with a `refinedWire` display style. We can then edit this mirror mesh to only contain the topology related to the selected GeomSubsets. However, to do so, it is much simpler and more consistent to move away from `refinedWireOnSurf` entirely, and just have mesh selections also use mirror prims : this way, we can use the selection highlight mirror hierarchy for everything, and we only have one consistent (though somewhat complex) way of handling all highlights.


lib/flowViewport/sceneIndex/fvpWireframeSelectionHighlightSceneIndex.cpp
```diff
@@ -51,14 +57,6 @@ const HdRetainedContainerDataSourceHandle sRefinedWireDisplayStyleDataSource
             HdRetainedTypedSampledDataSource<VtArray<TfToken>>::New(
                 { HdReprTokens->refinedWire, TfToken(), TfToken() })));
 
-const HdRetainedContainerDataSourceHandle sRefinedWireOnSurfaceDisplayStyleDataSource
-    = HdRetainedContainerDataSource::New(
-        HdLegacyDisplayStyleSchemaTokens->displayStyle,
-        HdRetainedContainerDataSource::New(
-            HdLegacyDisplayStyleSchemaTokens->reprSelector,
-            HdRetainedTypedSampledDataSource<VtArray<TfToken>>::New(
-                { HdReprTokens->refinedWireOnSurf, TfToken(), TfToken() })));
```
debloip-adsk, line 60: Everything now uses separate prims with `refinedWire` only

lib/flowViewport/sceneIndex/fvpWireframeSelectionHighlightSceneIndex.cpp
```diff
@@ -82,103 +80,25 @@ SdfPath _GetOriginalPathFromSelectionHighlightMirror(const SdfPath& mirrorPath)
     return mirrorPath.ReplaceName(TfToken(primName.substr(0, primName.size() - selectionHighlightMirrorTag.size())));
 }
 
-// Computes the mask to use for an instancer's selection highlight mirror
-// based on the instancer's topology and its selections. This allows
-// highlighting only specific instances in the case of instance selections.
-VtBoolArray _GetSelectionHighlightMask(const HdInstancerTopologySchema& originalInstancerTopology, const HdSelectionsSchema& selections)
```
debloip-adsk, line 88: Moved to a method

lib/flowViewport/sceneIndex/fvpWireframeSelectionHighlightSceneIndex.cpp
```diff
@@ -82,103 +80,25 @@ SdfPath _GetOriginalPathFromSelectionHighlightMirror(const SdfPath& mirrorPath)
...
-            }
-        }
-    }
-
-    return selectionHighlightMask;
-}
-
-// Returns the overall data source for an instancer's selection highlight mirror.
-// This replaces the mask data source and blocks the selections data source.
-HdContainerDataSourceHandle _GetSelectionHighlightInstancerDataSource(const HdContainerDataSourceHandle& originalDataSource)
```
debloip-adsk, line 146: Moved to a method

lib/flowViewport/sceneIndex/fvpWireframeSelectionHighlightSceneIndex.cpp
```diff
@@ -410,13 +330,130 @@ WireframeSelectionHighlightSceneIndex(
...
+        return originalMask;
+    }
+    VtBoolArray selectionHighlightMask = [&]() {
+        if (!selections.IsDefined()) {
+            return originalMask.empty() ? VtBoolArray(nbInstances, true) : originalMask;
+        }
+        return VtBoolArray(nbInstances, false);
+    }();
+
+    if (!selections.IsDefined()) {
```
debloip-adsk, line 383: This is the newly handled case and why this function and `_GetSelectionHighlightInstancerDataSource` was moved to a method, in order to call `GetSelectionHighlightPath`

lib/flowViewport/sceneIndex/fvpWireframeSelectionHighlightSceneIndex.cpp
```diff
@@ -474,11 +514,18 @@ WireframeSelectionHighlightSceneIndex::GetChildPrimPaths(const SdfPath &primPath
...
-        auto childPaths = GetInputSceneIndex()->GetChildPrimPaths(primPath.ReplacePrefix(selectionHighlightMirrorAncestor, originalAncestor));
-        for (auto& childPath : childPaths) {
-            childPath = childPath.ReplacePrefix(originalAncestor, selectionHighlightMirrorAncestor);
+        auto originalChildPaths = GetInputSceneIndex()->GetChildPrimPaths(primPath.ReplacePrefix(selectionHighlightMirrorAncestor, originalAncestor));
+        SdfPathVector implicitSelectionHighlightChildPaths;
+        for (const auto& originalChildPath : originalChildPaths) {
+            SdfPath explicitSelectionHighlightChildPath = _GetSelectionHighlightMirrorPathFromOriginal(originalChildPath);
+            if (_SelectionHighlightMirrorExists(explicitSelectionHighlightChildPath)) {
+                // There already exists an explicit selection highlight mirror for this child (e.g. point instance prototypes),
+                // so don't create a duplicate implicit one.
```
debloip-adsk, line 523: This is just to avoid overdraw, as without the check everything still works, but a highlight could be drawn more than once

lib/flowViewport/sceneIndex/fvpWireframeSelectionHighlightSceneIndex.cpp
```diff
@@ -791,125 +963,173 @@ WireframeSelectionHighlightSceneIndex::_CollectSelectionHighlightMirrors(const P
...
+    SdfPathVector affectedPrototypePaths;
+    SdfPathVector affectedInstancedByPaths;
     auto operation = [&](const SdfPath& primPath, const HdSceneIndexPrim& prim) -> bool {
         outAddedPrims.push_back({primPath.ReplacePrefix(originalPrimPath, selectionHighlightPrimPath), prim.primType});
-        if (prim.primType == HdPrimTypeTokens->instancer) {
-            SdfPathVector instancingRelatedPaths = _GetInstancingRelatedPaths(prim);
-            affectedOriginalPrimPaths.insert(affectedOriginalPrimPaths.end(), instancingRelatedPaths.begin(), instancingRelatedPaths.end());
-            return false; // We have found an instancer, don't process its children (nested instancers will be processed through the instancing-related paths).
+        if (prim.primType == HdPrimTypeTokens->instancer || prim.primType == HdPrimTypeTokens->mesh) {
+            if (direction & SelectionHighlightsCollectionDirection::Prototypes) {
```
debloip-adsk, line 973: Splitting the instancing-related paths in two prevents us from always traversing the entire instancing graph, and inadvertently creating selection highlights for non-selected prims. Without this, selecting an instanced mesh would create a highlight mirror for the mesh, then a highlight mirror for the instancer, but in creating the highlight mirror for the instancer and collecting its instancing paths, we would collect its other prototype paths, and then create highlight mirrors for those as well, even if they weren't actually selected.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1894,6 +1894,7 @@ void MtohRenderOverride::_PickByRegion(
     pickParams.viewMatrix.Set(viewMatrix.matrix);
     pickParams.projectionMatrix.Set(adjustedProjMatrix.matrix);
     pickParams.collection = _renderCollection;
+    pickParams.collection.SetExcludePaths(_wireframeSelectionHighlightSceneIndex->GetSelectionHighlightMirrorPaths());
```
debloip-adsk, line 1897: We need to exclude selection highlight prims. If we don't, selecting the highlight mirror of a mesh that is trimmed by a GeomSubset selection ends up selecting the original mesh, instead of selecting the GeomSubset.

test/lib/mayaUsd/render/mayaToHydra/cpp/testUtils.cpp
```diff
@@ -437,4 +442,67 @@ QPoint getPrimMouseCoords(const HdSceneIndexPrim& prim, M3dView& view)
     return QPoint(viewportX, view.portHeight() - viewportY);
 }
 
+// This method takes in a path to a prim in a selection highlight hierarchy and ensures that 
+// the selection highlight graph is structured properly, and that the leaf mesh prims have
+// the proper display style.
+void assertSelectionHighlightCorrectness(
```
debloip-adsk, line 448: Extracted and slightly tweaked from `testPointInstancingWireframeHighlight.cpp`

ppt-adsk (approved):
Can't say I delved very deeply into the logic, but the tests pass, so that works for me.

As discussed, agree that the approach is the right one, and is the way to get the draw flexibility we need, but concerned about z-fighting and resulting decrease in quality.  Can be addressed at a later time.

PR #151, opened 2024-07-22
HYDRA-889 - Create a unit test for the shaded display mode with a texture image file

vlasovi:


vlasovi:

test/lib/mayaUsd/render/mayaToHydra/testUVandUDIM.py
```diff
@@ -39,5 +39,11 @@ def test_UDIMs(self):
         self.setHdStormRenderer()
         self.verifySnapshot("quads_with_UDIMs.png")
 
+    def test_HDR(self):
+        mayaUtils.openTestScene("testUVandUDIM", "testHDR.ma")
+        self.setHdStormRenderer()
+        cmds.refresh()
+        self.assertSnapshotClose("quads_with_HDR.png", 0.04, 0.2)
```
debloip-adsk: For consistency, I would suggest we either use and adjust the existing threshold constants, or if we use bespoke thresholds per-test as here, then we should have the other test follow the same structure as well and adjust its code to mirror this one

test/testSamples/testUVandUDIM/testHDR.ma
```diff
@@ -0,0 +1,552 @@
+//Maya ASCII 2026ff01 scene
```
debloip-adsk: Year needs to be changed, Maya 2026 is not a released product

test/testSamples/testUVandUDIM/testHDR.ma
```diff
@@ -0,0 +1,552 @@
...
+//Name: testHDR.ma
+//Last modified: Thu, Jul 18, 2024 02:05:49 PM
+//Codeset: 1252
+requires maya "2026ff01";
```
debloip-adsk: Year needs to be changed, Maya 2026 is not a released product

debloip-adsk (changes requested):

debloip-adsk (approved):

PR #154, opened 2024-07-30
HYDRA-1133 : Cmake changes pertaining to OneTBB update

roopavr-adsk:
The following changes ensure maya-hydra picks up TBB from USD first instead of that from Maya.

### DIFF
--- a/cmake/modules/FindUSD.cmake
+++ b/cmake/modules/FindUSD.cmake
@@ -164,7 +164,7 @@ find_package_handle_standard_args(USD
         USD_VERSION
         PXR_VERSION
     VERSION_VAR
-        USD_VERSION 
+        USD_VERSION
 )
 
 find_program(OIIO_idiff_BINARY
--- a/cmake/modules/FindUSD.cmake
+++ b/cmake/modules/FindUSD.cmake
@@ -54,7 +54,7 @@ if(DEFINED PXR_VERSION)
     # there is no need to extract them from the pxr/pxr.h header file anymore.
     # The only thing we need to do is assemble the USD_VERSION version string.
     set(USD_VERSION ${PXR_MAJOR_VERSION}.${PXR_MINOR_VERSION}.${PXR_PATCH_VERSION})
-elseif(USD_INCLUDE_DIR AND EXISTS "${PXR_INCLUDE_DIR}/pxr/pxr.h")
+elseif(USD_INCLUDE_DIR AND EXISTS "${USD_INCLUDE_DIR}/pxr/pxr.h")
     foreach(_usd_comp MAJOR MINOR PATCH)
         file(STRINGS
             "${USD_INCLUDE_DIR}/pxr/pxr.h"
@@ -146,7 +146,8 @@ if (USD_LIBRARY_DIR AND EXISTS "${USD_LIBRARY_DIR}/${USD_LIB_PREFIX}usdMtlx${CMA
     endif()
 endif()
 
-message(STATUS "USD include dir: ${PXR_INCLUDE_DIR}")
+message(STATUS "USD include dir: ${USD_INCLUDE_DIR}")
+message(STATUS "USD library dir: ${USD_LIBRARY_DIR}")
 message(STATUS "USD version: ${USD_VERSION}")
 if(DEFINED USD_BOOST_VERSION)
     message(STATUS "USD Boost::boost version: ${USD_BOOST_VERSION}")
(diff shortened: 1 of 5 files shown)


cmake/modules/FindUSD.cmake
```diff
@@ -159,13 +159,12 @@ find_package_handle_standard_args(USD
...
         PXR_USD_LOCATION
         USD_INCLUDE_DIR
-        USD_LIBRARY_DIR
         USD_GENSCHEMA
         USD_CONFIG_FILE
         USD_VERSION
         PXR_VERSION
     VERSION_VAR
-        USD_VERSION
+        USD_VERSION
```
ppt-adsk: Nit-pick: whitespace change

cmake/modules/FindUSD.cmake
```diff
@@ -159,13 +159,12 @@ find_package_handle_standard_args(USD
     REQUIRED_VARS
         PXR_USD_LOCATION
         USD_INCLUDE_DIR
```
seando-adsk, line 161: I think you should also replace  `USD_INCLUDE_DIR` by `PXR_INCLUDE_DIRS`. I don't think hydra even uses USD_INCLUDE_DIR so no other changes would be required.

seando-adsk (dismissed):

lib/mayaHydra/hydraExtensions/CMakeLists.txt
```diff
@@ -49,11 +49,14 @@ mayaHydra_compile_config(${TARGET_NAME})
...
+                ${PXR_INCLUDE_DIRS}
                 ${MAYA_INCLUDE_DIRS}
+)
+target_include_directories(${TARGET_NAME}
+            PUBLIC
                 ${PYTHON_INCLUDE_DIR}
-                ${PXR_INCLUDE_DIRS}
-                ${BOOST_INCLUDE_DIR}
+                ${BOOST_INCLUDE_DIR}               
                 ${CMAKE_BINARY_DIR}/include
```
ppt-adsk, line 60: Can't we fuse this into a single declaration?  Also, why are we using `SYSTEM PUBLIC`?  Is it required?

lib/mayaHydra/ufeExtensions/CMakeLists.txt
```diff
@@ -34,6 +34,11 @@ mayaHydra_compile_config(${TARGET_NAME})
 # -----------------------------------------------------------------------------
 target_include_directories(${TARGET_NAME} 
         PUBLIC
+            ${PXR_INCLUDE_DIRS}
+)
+
+target_include_directories(${TARGET_NAME} 
+        SYSTEM PUBLIC
```
ppt-adsk, line 41: Same as above, can't it be a single declaration?

plugin/mayaHydraSceneBrowserTest/CMakeLists.txt
```diff
@@ -25,8 +25,9 @@ mayaHydra_compile_config(${TARGET_NAME})
 # include directories
 # -----------------------------------------------------------------------------
 target_include_directories(${TARGET_NAME} 
-    PUBLIC
-        ${MAYA_INCLUDE_DIRS}
+            SYSTEM PUBLIC
```
ppt-adsk, line 28: Why `SYSTEM PUBLIC`?

lib/mayaHydra/hydraExtensions/CMakeLists.txt
seando-adsk: SYSTEM is required for Maya because USD are declared as SYSTEM (from the pxrConfig.cmake). And system includes **always** come last. So we must put MAYA in the same grouping and put USD first.

ppt-adsk: Would be nice to comment on this in the CMakeLists.txt file.  And can we merge this into a single declaration to avoid duplication?

seando-adsk: I don't think so. From the cmake doc it appears that SYSTEM applies to everything you list after.

roopavr-adsk: @ppt-adsk did you mean something like this? :
```
target_include_directories(${TARGET_NAME} 
            SYSTEM PUBLIC
                ${PXR_INCLUDE_DIRS}
                ${MAYA_INCLUDE_DIRS}

            PUBLIC
                ${PYTHON_INCLUDE_DIR}
                ${BOOST_INCLUDE_DIR}               
                ${CMAKE_BINARY_DIR}/include
            PRIVATE
)
```

ppt-adsk (approved):

lib/mayaHydra/hydraExtensions/CMakeLists.txt
ppt-adsk: Yes, @seando-adsk doubts this will work, but I feel this is more maintainable if it does work.

ppt-adsk: Maybe the PUBLIC section needs to go before the SYSTEM PUBLIC section.

seando-adsk: I don't think it will work.
https://cmake.org/cmake/help/v3.14/command/target_include_directories.html

Plus how will you test it? What you have above Rakesh will just put everything in SYSTEM.

PR #162, opened 2024-08-28
HYDRA-1052 : Fix when we select the maya node of a scene producer scene index it should highlight the prims

lanierd-adsk:
…ne index it should highlight the prims


lanierd-adsk:
Tests are failing on OSX, we're working on it.
FYI : @debloip-adsk @lilike-adsk

lib/flowViewport/sceneIndex/fvpSelectionSceneIndex.cpp
```diff
@@ -275,10 +275,9 @@ PrimSelections SelectionSceneIndex::UfePathToPrimSelections(const Ufe::Path& app
             TF_WARN("SelectionSceneIndex::UfePathToPrimSelections(%s) returned no path, Hydra selection will be incorrect", Ufe::PathString::string(appPath).c_str());
         };
 
-        if (!mapper) {
-            warnEmptyPath(appPath);
-        }
-        else {
+        //Don't emit a warning if there is no mapper for this path, _selectionSceneIndex->ReplaceSelection may be called 
+        // later again if there are some usd stage scene index or custom producer scene indices added to the merging scene index.
+        if (mapper) {
```
debloip-adsk: Not sure I understand the case where we don't want the warning. If we try to get the prim selections from a UFE path, but we got no prims either from the PathInterface and there is no mapper, to me that feels like an error case? Maybe the problem is that this code is/can be called too early?

lib/mayaHydra/hydraExtensions/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ target_sources(${TARGET_NAME}
         mhWireframeColorInterfaceImp.cpp
         mhLeadObjectPathTracker.cpp
         tokens.cpp
+		mhDataProducersMayaNodeToSdfPathRegistry.cpp
```
debloip-adsk: (Formatting) : inconsistent whitespace from the other files

lib/mayaHydra/hydraExtensions/CMakeLists.txt
```diff
@@ -30,6 +31,7 @@ set(HEADERS
     mhWireframeColorInterfaceImp.h
     mhLeadObjectPathTracker.h
     tokens.h
+	mhDataProducersMayaNodeToSdfPathRegistry.h
```
debloip-adsk: (Formatting) : inconsistent whitespace from the other files

lib/mayaHydra/hydraExtensions/mhDataProducersMayaNodeToSdfPathRegistry.h
```diff
@@ -0,0 +1,73 @@
...
+#include <pxr/usd/sdf/path.h>
+
+//Maya headers
+#include <maya/MObjectHandle.h>
+
+#include <unordered_map>
+
+namespace MAYAHYDRA_NS_DEF {
+
+/// MhDataProducersMayaNodeToSdfPathRegistry does a mapping between Maya nodes and USD paths.
```
debloip-adsk: Technically the paths in this class are Hydra paths, not USD

lib/mayaHydra/hydraExtensions/mhDataProducersMayaNodeToSdfPathRegistry.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+std::unique_ptr<MhDataProducersMayaNodeToSdfPathRegistry>   MhDataProducersMayaNodeToSdfPathRegistry::_instance;
+
+MhDataProducersMayaNodeToSdfPathRegistry& MhDataProducersMayaNodeToSdfPathRegistry::Get()
+{
+    if (! _instance){
+        _instance.reset(new MhDataProducersMayaNodeToSdfPathRegistry());
+    }
+    return *_instance;
+}
```
debloip-adsk: I think we could simplify management of the singleton by either declaring the instance as a static variable within the `Get()` method, or by using the TfSingleton facilities like we do for example in the `PickHandlerRegistry`. It'd also improve consistency in our code style.

lib/mayaHydra/hydraExtensions/mhDataProducersMayaNodeToSdfPathRegistry.cpp
```diff
@@ -0,0 +1,81 @@
...
+    if (0 == objectHandleHashCode) {
+        return;
+    }
+
+    for (auto it = _SdfPathByObjectHandle.begin(); it != _SdfPathByObjectHandle.end(); ++it) {
+        if (it->first.hashCode() == objectHandleHashCode) {
+            _SdfPathByObjectHandle.erase(it);
+            return;
+        }
+    }
```
debloip-adsk: Instead of using `MObjectHandle`s as a key and providing a hashing function to the `_SdfPathByObjectHandle` map, could we use the integer hash code directly as a key? That way instead of having to iterate in linear time here, we could do it in constant time, and in the other parts of this class, we'd just have to add a `.hashCode()` call on the `MObjectHandle`s, which would also keep them in constant time?

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraMayaDataProducerSceneIndexData.cpp
```diff
@@ -53,8 +54,11 @@ MayaDataProducerSceneIndexData::MayaDataProducerSceneIndexData(FVP_NS_DEF::DataP
     : FVP_NS_DEF::DataProducerSceneIndexDataBase(params)
 {
     if (_dccNode) {
-        SetupUfeObservation(_dccNode);
+        SetupDCCNode();
     }
+    //Reset the prefix member which was used only for SetupDCCNode, as for usd stages we add a prefixing scene index outside of flow viewport which holds the prefix.
+    //This is a way to have a common call to MAYAHYDRA_NS::MhDataProducersMayaNodeToSdfPathRegistry::Get().Add in this class.
+    _prefix = SdfPath::AbsoluteRootPath();
```
debloip-adsk, line 66: Could we not standardize the way we handle USD data to follow the same structure as other data producers? Looking quickly at the `HdPrefixingSceneIndex` creation for the USD data producer ([registration.cpp](https://github.com/Autodesk/maya-hydra/blob/827a42e6a11ded03e00e6a0e64d37371dc43e54e/lib/mayaHydra/hydraExtensions/sceneIndex/registration.cpp#L437-L441)), I don't see why we couldn't just rely on the prefixing scene index being added by the data producer API like [here](https://github.com/Autodesk/maya-hydra/blob/dev/827a42e6a11ded03e00e6a0e64d37371dc43e54e/flowViewport/API/perViewportSceneIndicesData/fvpDataProducerSceneIndexDataBase.cpp#L120-L131). This one is for non-USD data producers but I think we could do the same in the USD version of this method? And/or maybe use a `UsdImagingRootOverridesSceneIndex`?

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -758,7 +759,23 @@ Fvp::PrimSelections MayaHydraSceneIndex::UfePathToPrimSelections(const Ufe::Path
...
+    //The maya nodes registered in this class are used by data producers as a parent to all
+    //primitives. This class is used when the user selects one of these
+    //maya nodes to return the matching SdfPath so that all prims child of this maya node are
+    //highlighted.
+    MDagPath shapeDagPath(dagPath);
+    shapeDagPath.extendToShape();
+    const SdfPath matchingPath = MhDataProducersMayaNodeToSdfPathRegistry::Get().GetPath(MObjectHandle(shapeDagPath.node()));
+    if (! matchingPath.IsEmpty()) {
+        primPath = matchingPath;
+    }
```
debloip-adsk: Just thinking out loud, but I feel like it's a bit counterintuitive to have the `MayaHydraSceneIndex`, which AFAIK is for Maya data, handle the node selection for all other data producers. Though in a way since they're Maya nodes it also makes sense? I don't really have a clear solution in mind either, maybe we could move some stuff into the Maya implementation layer of the data producer interface? Maybe it'd be easier to address this separately though.

debloip-adsk (commented):
Sooo, I saw the FYI mention and since you had pinged me with a question regarding tests on that issue previously, I kind of hallucinated into thinking I was reviewer on this, and I just noticed it was actually Pierre as I was about to send my review, so I guess here are my comments anyways...

lib/flowViewport/sceneIndex/fvpSelectionSceneIndex.cpp
lanierd-adsk: This code is called in initHydraResource so when we switch to a Hydra viewport.
And it may be called later if there are data producer scene indices in the scene and in the second case we have all the scene setup so there should have no errors. But in the first call, the Usd stages prims are not yet in the merging scene index hence the warning.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraMayaDataProducerSceneIndexData.cpp
lanierd-adsk: I recall I had tried that already but the prefixing scene index should be above all.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
lanierd-adsk: For a data producer scene index (USD  stage or custom) you provide a maya node to be the parent of this node to move/delete/select the primitives. So I think it makes sense to have this code here since we handle maya nodes only.

lib/flowViewport/sceneIndex/fvpSelectionSceneIndex.cpp
lanierd-adsk: I found a way to keep this warning... Please have a look at the new code.

lanierd-adsk: It also calls always once replaceselection instead of potentially twice in my previous change.

lib/mayaHydra/hydraExtensions/mhDataProducersMayaNodeToSdfPathRegistry.cpp
```diff
@@ -15,64 +15,49 @@
...
+    return PXR_NS::TfSingleton<MhDataProducersMayaNodeToSdfPathRegistry>::GetInstance();
 }
 
-void MhDataProducersMayaNodeToSdfPathRegistry::Add(const MObjectHandle& objectHandle, const SdfPath& thePath)
+void MhDataProducersMayaNodeToSdfPathRegistry::Add(const unsigned long& objectHandleHashCode, const SdfPath& thePath)
 {
-    if (thePath.IsEmpty() || ! objectHandle.isValid() ) {
-        TF_CODING_WARNING("Sending an empty SdfPath or an invalid objectHandle in MhDataProducersMayaNodeToSdfPathRegistry::Add, ignoring");
+    if (thePath.IsEmpty() || 0 == objectHandleHashCode) {
+        TF_CODING_WARNING("Sending an empty SdfPath or an invalid objectHandle has code in MhDataProducersMayaNodeToSdfPathRegistry::Add, ignoring");
```
debloip-adsk: Small typo here (has code vs hash code)

lib/mayaHydra/hydraExtensions/mhDataProducersMayaNodeToSdfPathRegistry.cpp
```diff
@@ -15,64 +15,49 @@
...
+SdfPath MhDataProducersMayaNodeToSdfPathRegistry::GetPath(const unsigned long& objectHandleHashCode) const
 {
-    if(! objectHandle.isValid()){
-        return;
-    }
-    auto it = _SdfPathByObjectHandle.find(objectHandle);
-    if (it != _SdfPathByObjectHandle.end()) { 
-        _SdfPathByObjectHandle.erase(it);
-    }
-}
```
debloip-adsk: I think you could have kept this version and have it call `Remove(objectHandle.hashCode())` if you wanted, but I'll leave it up to you

debloip-adsk (changes requested):
Looks good, thanks! Just a small typo remaining

lib/flowViewport/sceneIndex/fvpSelectionSceneIndex.cpp
debloip-adsk: I see what you meant now, and yeah it is a bit awkward. But I do prefer handling this at the callsites like with the new changes than in the body of this method. Thanks for the changes!

lib/mayaHydra/hydraExtensions/mhDataProducersMayaNodeToSdfPathRegistry.cpp
lanierd-adsk: Thanks, sorry, I had forgotten to push the code with the typo fix. I had the ooportunity to move one class which was not dependent on maya any more since I use now the hash code and no longer the MObjectHandle so I moved it to flow viewport.

lib/flowViewport/selection/fvpDataProducersNodeHashCodeToSdfPathRegistry.h
```diff
@@ -0,0 +1,68 @@
...
+    DataProducersNodeHashCodeToSdfPathRegistry() = default;
+    ~DataProducersNodeHashCodeToSdfPathRegistry() = default;
+    DataProducersNodeHashCodeToSdfPathRegistry(const DataProducersNodeHashCodeToSdfPathRegistry& ) = delete;
+    DataProducersNodeHashCodeToSdfPathRegistry& operator=(const DataProducersNodeHashCodeToSdfPathRegistry& ) = delete;
+
+    friend class PXR_NS::TfSingleton<DataProducersNodeHashCodeToSdfPathRegistry>;
+
+    std::unordered_map<unsigned long, PXR_NS::SdfPath> _SdfPathByHashCode;
+
+    static std::unique_ptr<DataProducersNodeHashCodeToSdfPathRegistry>   _instance;
```
debloip-adsk: I think this can be removed since we now use `TfSingleton`

debloip-adsk (changes requested):
Sorry another small leftover thing, should probably have caught that before. Also regarding the `cmds.select(clear=True)` changes, what is the motivation behind these? Is it due to failing tests?

lanierd-adsk:
For the clearSelection, yes it's to avoid having selection highlight for primitives which don't work on OSX. It was the original behavior not to see these lines.

debloip-adsk (dismissed):

lib/flowViewport/API/interfacesImp/fvpDataProducerSceneIndexInterfaceImp.cpp
```diff
@@ -26,6 +26,13 @@
 //Std Headers
 #include <mutex>
 
+/// DataProducersNodeHashCodeToSdfPathRegistry does a mapping between DCC nodes hash code and Hydra
```
ppt-adsk, line 29: Conceptually it feels like there is considerable overlap between mapping a DCC node hash code to a Hydra path, and the Fvp::PathMapperRegistry, which maps a UFE path to a Hydra path.  Do you agree?

lib/flowViewport/API/perViewportSceneIndicesData/fvpDataProducerSceneIndexDataBase.h
```diff
@@ -113,6 +113,8 @@ TF_DECLARE_WEAK_AND_REF_PTRS(DataProducerSceneIndexDataBase);//Be able to use Re
     std::string                         _rendererNames;
     /// Is the DCC node so a MObject* DAG node for Maya
     void*                               _dccNode;
+    /// Is the DCC node hashcode (we use it to remove the node from a registry when it has been deleted)
+    unsigned long                       _dccNodeHashCode = 0;
```
ppt-adsk, line 117: Is the hash code implemented with a pointer?  The C++ standard says that this must be at least 32 bits, but unsigned long long would be the proper type for 64 bits:
https://en.cppreference.com/w/cpp/language/types
Either that or std::uintptr_t.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -784,9 +785,25 @@ Fvp::PrimSelections MayaHydraSceneIndex::UfePathToPrimSelections(const Ufe::Path
...
+  
+    //Check if this maya node has a special SdfPath associated with it, this is for custom or maya usd data producers scene indices.
+    //The class MhDataProducersMayaNodeToSdfPathRegistry does a mapping between Maya nodes and USD paths.
+    //The maya nodes registered in this class are used by data producers as a parent to all
+    //primitives. This class is used when the user selects one of these
+    //maya nodes to return the matching SdfPath so that all prims child of this maya node are
+    //highlighted.
+    MDagPath shapeDagPath(dagPath);
+    shapeDagPath.extendToShape();
+    const SdfPath matchingPath = FVP_NS::DataProducersNodeHashCodeToSdfPathRegistry::Instance().GetPath(MObjectHandle(shapeDagPath.node()).hashCode());
```
ppt-adsk, line 802: The usage of the DataProducersNodeHashCodeToSdfPathRegistry makes me feel we can unify with Fvp::PathMapperRegistry.

lib/mayaHydra/hydraExtensions/sceneIndex/registration.cpp
```diff
@@ -440,11 +436,12 @@ void MayaHydraSceneIndexRegistry::_AddSceneIndexForNode(MObject& dagNode)
     // the MayaUsdProxyShapeSceneIndex.  We don't want to automatically add an
     // additional prefixing scene index to the PathInterfaceSceneIndex (which
     // is downstream of the prefixing stream index), which would double the
-    // prefix.  Therefore, set the scene index prefix here to be 'null' (the
-    // absolute root path).
+    // prefix.  But to register a mapping from the maya node to the prefix SdfPath, we give send
```
ppt-adsk, line 439: Comment error: "give send", probably just "send".

ppt-adsk (approved):
Minor comments, and a more software architecture one for later.

PR #163, opened 2024-08-28
HYDRA-951 : Add Maya command plugin to query build info

roopavr-adsk:


lib/mayaHydra/mayaPlugin/CMakeLists.txt
```diff
@@ -48,6 +49,7 @@ target_include_directories(${TARGET_NAME}
     PRIVATE
         $<$<BOOL:${UFE_FOUND}>:${UFE_INCLUDE_DIR}>
         $<$<BOOL:${MayaUsd_FOUND}>:${MAYAUSD_INCLUDE_DIR}>
+        ${CMAKE_BINARY_DIR}
```
lilike-adsk: what's this for?

lib/mayaHydra/mayaPlugin/pluginBuildInfoCommand.h
```diff
@@ -0,0 +1,37 @@
...
+
+namespace MAYAHYDRA_NS_DEF {
+
+class MayaHydraPluginInfoCommand : public MPxCommand
+{
+public:
+    static void*   creator() { return new MayaHydraPluginInfoCommand(); }
+    static MSyntax createSyntax();
+
+    static const MString commandName;
```
lilike-adsk, line 31: where this member was used?

lib/mayaHydra/mayaPlugin/CMakeLists.txt
roopavr-adsk: unintended inclusion. removed it.

lib/mayaHydra/mayaPlugin/pluginBuildInfoCommand.h
roopavr-adsk: did you mean to ask about commandName ?

lilike-adsk: yes, commandName

ppt-adsk:
As per my comments elsewhere, if we have a new command to query mayaHydra information, we should remove it from the mayaHydra command:
https://github.com/Autodesk/maya-hydra/blob/dev/lib/mayaHydra/mayaPlugin/viewCommand.cpp
And we should change the test
https://github.com/Autodesk/maya-hydra/blob/dev/test/lib/mayaUsd/render/mayaToHydra/testMtohCommand.py
to use your new command, as well as test the new cutId flag. Thanks!

lib/mayaHydra/mayaPlugin/pluginBuildInfoCommand.h
lilike-adsk: do we have a ticket to remove it from the mayaHydra command?

roopavr-adsk: Good catch. A file was missed to be added to this PR. Added now.

lilike-adsk (dismissed):

ppt-adsk (dismissed):

lib/mayaHydra/mayaPlugin/viewCommand.cpp
ppt-adsk, line 1: Why are we removing this?  Seems like useful information.

roopavr-adsk: To print help of a mel command users are using help <command>. So this was a bit redundant. This was also part of the description in the ticket.

ppt-adsk: But help theCommand only lists the flags and their argument types.  This is information about use.  I would challenge that part of the ticket, and if it is really felt like it's a good idea, then move the information to a Markdown page in the repo.  This doesn't address the manual maintenance objection, but I still think it's worth it.  My preference would be to keep the -help flag.

roopavr-adsk: Fair point.

bjorn-siegert-adsk: I am okay with leaving the help if it provides "how to use" information but I would like to avoid to list all the args as those need to be kept in sync with the args help. Maybe add something like:
_For more information about supported arguments and their types use `help`_

roopavr-adsk: Shall we move the detailed information to a Markdown file then? Maybe a new markdown file here: https://github.com/Autodesk/maya-hydra/tree/dev/doc

lilike-adsk (approved):

PR #164, opened 2024-08-28
Isolate select scene index, with command-based implementation.

ppt-adsk:
Implement isolate select as a scene index filter. The isolate select scene index filter removes objects (and their hierarchies) that are not visible in a given viewport, allowing the user to focus on data of interest and reducing scene load. Isolate select can be different on a per-viewport basis.

The isolate selection is kept on a per-viewport basis in the Fvp::ViewportInformationAndSceneIndicesPerViewportDataManager, keyed by model panel string (used as a viewport ID). Since we have a single scene index tree for all viewports, switching viewports means switching isolate selection. Isolate selection is implemented by setting visibility to 0 on invisible object, rather than removing these objects from the scene. This matches what is done in Maya for isolate selection. A consequence of this is that we are sharing a single scene, and its objects, among all viewports. We must therefore adjust object visibility to match the isolate selection for each viewport.

When switching from one viewport to another, the isolate scene index dirties the previous isolate selection, to remove its effect on the visibility of objects as specified for the previous viewport. We then dirty visibility as appropriate for the new isolate selection.

In this pull request the isolate select scene index is not connected to the UI, and must be set through a mayaHydra testing command, which is what is done in the unit test.


lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportDataManager.cpp
```diff
@@ -151,6 +152,91 @@ ViewportInformationAndSceneIndicesPerViewportData* ViewportInformationAndSceneIn
...
+    }
+    _CheckAndSetViewport(viewportId);
+    _isolateSelectSceneIndex->ClearIsolateSelection();
+}
+
+void ViewportInformationAndSceneIndicesPerViewportDataManager::SetIsolateSelectSceneIndex(
+    const IsolateSelectSceneIndexRefPtr& sceneIndex
+)
+{
+    _isolateSelectSceneIndex = sceneIndex;
```
lanierd-adsk, line 221: Is there anything you need to clean up if there was a previous sceneIndex ?

ppt-adsk: I don't believe so.  The isolate selections exist independently of the isolate scene index.

lanierd-adsk (dismissed):

lanierd-adsk (approved):

PR #166, opened 2024-09-04
Update maya-hydra-preflight-launcher.yml

seando-adsk:
Don't use ubuntu-latest as it was updated to Python 3.12 which is not supported by MayaHydra yet.

### DIFF
--- a/.github/workflows/maya-hydra-preflight-launcher.yml
+++ b/.github/workflows/maya-hydra-preflight-launcher.yml
@@ -11,7 +11,7 @@ jobs:
 
 #  clang_format_linter:
 #      timeout-minutes: 30
-#      runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/maya-hydra-preflight-launcher.yml
```diff
@@ -11,7 +11,7 @@ jobs:
 
 #  clang_format_linter:
 #      timeout-minutes: 30
-#      runs-on: ubuntu-latest
+#      runs-on: ubuntu-22.04
```
seando-adsk, line 14: In MayaUsd our clang-format step started failing all of a sudden since the unbuntu-latest (24.04) was updated to Python 3.12.

roopavr-adsk (approved):

PR #167, opened 2024-09-05
HYDRA-794 : Support maya lighting modes for USD and dome lights

lanierd-adsk:
All kinds of lights actually


lanierd-adsk:
FYI : @debloip-adsk @lilike-adsk

test/lib/mayaUsd/render/mayaToHydra/testUsdTextureToggle.py
```diff
@@ -32,7 +32,8 @@ def setUp(self):
         # Open simple Maya scene
         testFile = mayaUtils.openTestScene(
                 "testUsdTextureToggle",
-                "testUsdTextureToggle.ma")
+                "testUsdTextureToggle.ma", useTestSettings=False)
```
ppt-adsk, line 35: I'm unsure about this change.  Is it to avoid
https://github.com/Autodesk/maya-hydra/blob/3e96b725311fc88900b977edd29410ff30dba1e7/test/testUtils/mayaUtils.py#L145
?  If so, perhaps it would be better to keep useTestSettings as True, then set the default light intensity.

lib/flowViewport/sceneIndex/fvpLightsManagementSceneIndex.cpp
```diff
@@ -0,0 +1,163 @@
...
+//ufe
+#include <ufe/globalSelection.h>
+#include <ufe/observableSelection.h>
+
+
+PXR_NAMESPACE_USING_DIRECTIVE
+
+namespace {
+
+bool _IsALight(const TfToken& primType)
```
ppt-adsk: Isn't this just
https://github.com/PixarAnimationStudios/OpenUSD/blob/59992d2178afcebd89273759f2bddfe730e59aa8/pxr/imaging/hd/tokens.cpp#L81
?

lib/flowViewport/sceneIndex/fvpLightsManagementSceneIndex.cpp
```diff
@@ -0,0 +1,163 @@
...
+{
+    static const TfTokenVector & lightsToken = HdLightPrimTypeTokens();//Get all light tokens from USD
+    return lightsToken.end() != std::find(lightsToken.begin(), lightsToken.end(), primType);
+}
+
+void _DisableLight(HdSceneIndexPrim& prim)
+{
+    //We don't set the intensity to 0 as for domelights this makes disappear the geometry
```
ppt-adsk:
```suggestion
    //We don't set the intensity to 0 as for domelights this makes the geometry disappear 
```

ppt-adsk: Is this intensity behavior the same for all dome lights, or specific to the Arnold one?

lib/flowViewport/sceneIndex/fvpLightsManagementSceneIndex.cpp
```diff
@@ -0,0 +1,163 @@
...
+        HdRetainedTypedSampledDataSource<float>::New(0.0f)
+    );
+    editor.Set(
+        HdLightSchema::GetDefaultLocator().Append(HdLightTokens->diffuse),
+        HdRetainedTypedSampledDataSource<float>::New(0.f)
+    );
+    editor.Set(
+        HdLightSchema::GetDefaultLocator().Append(HdLightTokens->specular),
+        HdRetainedTypedSampledDataSource<float>::New(0.f)
+    );
```
ppt-adsk: Minor: would the following work?
```suggestion
    for (const auto& t : {HdLightTokens->ambient, HdLightTokens->diffuse, HdLightTokens->specular}) {
        editor.Set(HdLightSchema::GetDefaultLocator().Append(t),
            HdRetainedTypedSampledDataSource<float>::New(0.0f)
        );
    }
```

lib/flowViewport/sceneIndex/fvpLightsManagementSceneIndex.cpp
```diff
@@ -0,0 +1,163 @@
...
+         } break;
+         default:
+         case LightingMode::kSceneLighting: {
+             return prim;
+         } break;
+         case LightingMode::kDefaultLighting: {
+             if (! _IsDefaultLight(primPath)){
+                 _DisableLight(prim);
+             }
+             return prim;//This is also handled in renderOverride.cpp with the _hasDefaultLighting member
```
ppt-adsk: Flow Viewport Toolkit can't really know about renderOverride.cpp.  I assume all that is needed is to adjust the comment to be more generic.

lib/flowViewport/sceneIndex/fvpLightsManagementSceneIndex.cpp
```diff
@@ -0,0 +1,163 @@
...
+         } break;
+         case LightingMode::kSelectedLightsOnly: {
+             const Ufe::GlobalSelection::Ptr& globalUfeSelection = Ufe::GlobalSelection::get();
+             const Ufe::Selection&            ufeSelection = *globalUfeSelection;
```
ppt-adsk: Minor and optional (depends on personal preference really): since you don't use globalUfeSelection again, may as well just do
```suggestion
              const Ufe::Selection&            ufeSelection = *Ufe::GlobalSelection::get();
```

lib/flowViewport/sceneIndex/fvpLightsManagementSceneIndex.cpp
```diff
@@ -0,0 +1,163 @@
...
+             if (0 == ufeSelection.size()) {
```
ppt-adsk: Minor:
```suggestion
             if (ufeSelection.empty()) {
```

lib/mayaHydra/hydraExtensions/mayaUtils.h
```diff
@@ -176,6 +176,15 @@ bool SetNodeAttribute(MObject node, std::string attrName, AttrType newValue)
...
 MObject GetShadingGroupFromShader(const MObject& shader);
 
+/**
+ * @brief Get if this MDagPath is an Arnold sky dome light.
+ *
+ * @param[in] dagPath is a MDagPath
+ *
+ * @return true if the object is a sky dome light, false otherwise
+ */
+bool IsDagPathAnArnoldSkyDomeLight(const MDagPath& dagPath);
```
ppt-adsk, line 186: This makes me uncomfortable.  Why are we treating Arnold lights in a special way, with a special conditional?  Is there no way that our adapter infrastructure can handle this?  What if another renderer has a different way of handling dome lights?

lib/mayaHydra/hydraExtensions/mayaUtils.cpp
```diff
@@ -184,4 +184,15 @@ MObject GetShadingGroupFromShader(const MObject& shader)
...
 }
 
+bool IsDagPathAnArnoldSkyDomeLight(const MDagPath& dagPath) 
+{ 
+    static const std::string _aiSkyDomeLight("aiSkyDomeLight");
+
+    if (! dagPath.isValid()) return false;
+
+    const std::string fpName = dagPath.fullPathName().asChar();
+    return (fpName.find(_aiSkyDomeLight) != std::string::npos);
```
ppt-adsk: Shouldn't we be using the type of the node, rather than the node name?  What if the user renames the dome light?

And in any case, to me this immediately brings up the design principle of trying to replace type-based conditionals with virtual functions.  Is there a way we can make this more generic so that we can easily add in the same support for other renderers?

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -784,26 +784,32 @@ Fvp::PrimSelections MayaHydraSceneIndex::UfePathToPrimSelections(const Ufe::Path
...
-        .Msg("    mapped to scene index path %s.\n", primPath.GetText());
-  
+    const bool isDomeLight = IsDagPathAnArnoldSkyDomeLight(dagPath);
+    const bool isSprim = isDomeLight;//we will look in sprims only for the sky dome light
+    MDagPath   shapeDagPath(dagPath);
+    shapeDagPath.extendToShape();
+
+    //For the sky dome light, we want to get it from the Sprims and we are using the dag path shape 
+    // as this is the dag path that gets translated to an Sprim in Hydra
+    SdfPath primPath = GetPrimPath((isDomeLight) ? shapeDagPath : dagPath, isSprim);
```
ppt-adsk: Somehow I would have hoped that we can make this more generic, and not special for Arnold dome lights.

ppt-adsk (commented):
Looks good, like the light management scene index, but a bit uncomfortable with the Arnold dome light special case.  Can we make this more generic somehow?

lib/flowViewport/sceneIndex/fvpLightsManagementSceneIndex.cpp
lanierd-adsk: Yes, thanks, I didn't know that function. I am using it now.

lib/flowViewport/sceneIndex/fvpLightsManagementSceneIndex.cpp
lanierd-adsk: It's for all dome lights as this is happening in Storm, when the dome light intensity is 0 it is removed, not only ignored.

lib/mayaHydra/hydraExtensions/mayaUtils.h
lanierd-adsk: This is specific to maya hydra, not specific to a renderer.

test/lib/mayaUsd/render/mayaToHydra/testUsdTextureToggle.py
lanierd-adsk: Yes the intensity of lights doesn't match, I would rather keep it like that as it works fine.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
lanierd-adsk: Yes, I did that.

lilike-adsk:
Good work, it would be better to manage the maya built-in lights by the new SceneIndex, lets track it in another ticket.

lilike-adsk (approved):

PR #169, opened 2024-09-13
HYDRA-1054 : Support material assignment on a usd prim

lanierd-adsk:
- Support material assignment on a usd prim
- Support default material for usd procedural prims (sphere, capsule, cylinder etc..)


lanierd-adsk:
FYI : @ppt-adsk @lilike-adsk

lib/flowViewport/sceneIndex/fvpDefaultMaterialSceneIndex.cpp
```diff
@@ -68,7 +88,7 @@ HdSceneIndexPrim DefaultMaterialSceneIndex::GetPrim(const SdfPath& primPath) con
 bool DefaultMaterialSceneIndex::_ShouldWeApplyTheDefaultMaterial(const HdSceneIndexPrim& prim)const
 {
     // Only for meshes so far
```
debloip-adsk: Comment needs to be adjusted

debloip-adsk (changes requested):
Looks good! Just one outdated comment

debloip-adsk (dismissed):

lib/flowViewport/sceneIndex/fvpDefaultMaterialSceneIndex.cpp
```diff
@@ -22,13 +22,33 @@
...
 
 PXR_NAMESPACE_USING_DIRECTIVE
 
 namespace{
     static const TfToken purposes[] = { HdMaterialBindingsSchemaTokens->allPurpose };
+
+    // Support implicit surfaces from USD as well, not only meshes
+    bool _IsDefaultMaterialCompliantPrimitive(const TfToken& primType) 
+    { 
+        static VtArray<TfToken> const compliantPrimitives = { HdPrimTypeTokens->cone,
```
ppt-adsk: Minor: would be slightly better as an std::set, with the check using either std::set::find() or std::set::count().

lib/mayaHydra/hydraExtensions/mixedUtils.cpp
```diff
@@ -78,6 +78,11 @@ SdfPath DagPathToSdfPath(
     const bool      stripNamespaces)
 {
     std::string name = dagPath.fullPathName().asChar();
+    if ( name.empty() ) {
+        MFnDependencyNode dep(dagPath.node());
+        TF_WARN("Empty fullpath name for DAG object : %s of type : %s", dep.name().asChar(), dep.typeName().asChar());
```
ppt-adsk, line 83: Do we know when this happens?  Is this for non-Dag nodes?  Does this indicate missing functionality, missing logic, or a bug?

ppt-adsk (commented):
Looks good, thanks!  Only minor comments.

lib/mayaHydra/hydraExtensions/mixedUtils.cpp
lanierd-adsk: I had this with Bifrost nodes and also when selecting the default material set. But the problem is that wer get dag nodes with no type and no name, we should try detect useful dag nodes that need to be translated as some may not need to be treated.

lib/flowViewport/sceneIndex/fvpDefaultMaterialSceneIndex.cpp
```diff
@@ -22,13 +22,33 @@
...
+                                                              HdPrimTypeTokens->cube,
+                                                              HdPrimTypeTokens->sphere,   
+                                                              HdPrimTypeTokens->capsule,
+                                                              HdPrimTypeTokens->capsule_1,
+                                                        #if HD_API_VERSION >= 67 // USD 24.05+
+                                                              UsdGeomTokens->TetMesh,
+                                                        #endif
+                                                              UsdGeomTokens->Plane,
+                                                              HdPrimTypeTokens->mesh};
+        return std::find(compliantPrimitives.cbegin(), compliantPrimitives.cend(), primType)
```
ppt-adsk, line 49: Must use std::set::find() here to get better performance, and it also makes the code more compact.  I also think std::set::count() is even more compact, but that's personal preference and std::set::find() is fine.

lib/flowViewport/sceneIndex/fvpDefaultMaterialSceneIndex.cpp
```diff
@@ -22,13 +22,33 @@
...
+            != compliantPrimitives.cend();
```
ppt-adsk: Must use std::set::find() here to get better performance, and it also makes the code more compact. I also think std::set::count() is even more compact, but that's personal preference and std::set::find() is fine
```suggestion
        return compliantPrimitives.count(primType) == 1;
```

lib/flowViewport/sceneIndex/fvpDefaultMaterialSceneIndex.cpp
ppt-adsk, line 1: Github is erroring out saying that "You need to add a comment indicating the requested change.", even though I've done exactly that.  Hoping that this additional comment will help.

ppt-adsk (changes requested):

ppt-adsk:
Here's a useful reference:
https://stackoverflow.com/questions/17256001/merits-of-stdfind

ppt-adsk (approved):

PR #171, opened 2024-09-17
HYDRA-973 - set HDST_RENDER_SELECTED_EDGE_FROM_FACE as 0 by default

lilike-adsk:
Disable RENDER_SELECTED_EDGE_FROM_FACE feature that could trigger unnecessary running of gometry shader.
This should be safe as RENDER_SELECTED_EDGE_FROM_FACE feature is not used in MayaHydra.


lanierd-adsk (approved):

lilike-adsk:
@tremblp @debloip FYI.

PR #174, opened 2024-09-19
Disable two cases to satisfy USD 24.08

lilike-adsk:
Disable two cases to satisfy USD 24.08


debloip-adsk (approved):

ppt-adsk (approved):
Might have been better to leave the test enabled in Maya 2025 by using skipUnless in the test, e.g.
https://github.com/Autodesk/maya-usd/blob/3e62584f742dc3c4c15108532b97248c7b8bf7c8/test/lib/testMayaHydraPlugin.py#L35
but since this is temporary it's fine.

PR #178, opened 2024-09-25
Revert cmake_minimum_required change change

lilike-adsk:
Per doc https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#policy-settings, the max version in call to cmake_minimum_required(VERSION <min>[...<policy_max>] [FATAL_ERROR]) is only for the polices. Change the policy will introduce other potential issues, so, revert the max version change and find an another way to solve the original issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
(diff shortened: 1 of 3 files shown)


lilike-adsk:
@lanierd-adsk @debloip-adsk

sylvain-laroche-adsk (approved):

ppt-adsk (approved):

PR #180, opened 2024-09-28
HYDRA-973 - fix putenv not working on linux/mac

lilike-adsk:
There's a legayc issue that leads the env var failed to be set, seems putenv is not working well on Linux/Mac, switch to user setenv().

### DIFF
--- a/lib/mayaHydra/mayaPlugin/plugin.cpp
+++ b/lib/mayaHydra/mayaPlugin/plugin.cpp
@@ -70,14 +70,13 @@ namespace {
 
     std::vector<MCallbackId> _pluginLoadingCallbackIds;
 
-    void setEnvVar(const char* envVarSet)
+    void setEnv(const std::string& name, const std::string& value)
     {
-        // putenv requires char* and I'm not willing to use const cast!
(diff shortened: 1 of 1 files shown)


lanierd-adsk (approved):

lib/mayaHydra/mayaPlugin/plugin.cpp
```diff
@@ -70,14 +70,13 @@ namespace {
...
 
-    void setEnvVar(const char* envVarSet)
+    void setEnv(const std::string& name, const std::string& value)
     {
-        // putenv requires char* and I'm not willing to use const cast!
-        const auto        envVarSize = strlen(envVarSet) + 1;
-        std::vector<char> envVarData;
-        envVarData.resize(envVarSize);
-        snprintf(envVarData.data(), envVarSize, "%s", envVarSet);
-        putenv(envVarData.data());
```
ppt-adsk, line 80: As per
https://stackoverflow.com/questions/5873029/questions-about-putenv-and-setenv
putenv does not make a copy of the argument string.  Therefore, in this code, as soon as the function exits and the memory allocated for envVarData is re-used, the environment will be changed / corrupted.  setenv() is the proper choice.

PR #181, opened 2024-10-02
HYDRA-1206 - Re-enable the testMaterialXOnNative case after for new lookdevX with usd 24.08

lilike-adsk:
Re-enable the testMaterialXOnNative case for new lookdevX with usd 24.08


lilike-adsk:
@lanierd-adsk @debloip-adsk

ppt-adsk (approved):

PR #182, opened 2024-10-02
HYDRA-1207 - fix and re-enable case testSceneBrowser

lilike-adsk:
Adjust test code in sceneBrowser to make the case happy on SdfPathVector value to compensate the change in USD: https://github.com/PixarAnimationStudios/OpenUSD/commit/1d19b1dbdd0daf12247f5a1cb4b858c444e3e4fa#diff-8c7415891d45a3810ea025ac58c7e025043412150b628560c12a1b8fdbe1f830R57

### DIFF
--- a/lib/adskHydraSceneBrowser/test/adskHydraSceneBrowserTestFixture.cpp
+++ b/lib/adskHydraSceneBrowser/test/adskHydraSceneBrowserTestFixture.cpp
@@ -257,7 +257,16 @@ void AdskHydraSceneBrowserTestFixture::CompareValueContent(const pxr::VtValue& v
     std::string actualValue = valueText.toStdString();
 
     std::ostringstream valueStream;
-    valueStream << value;
+    if (value.IsHolding<pxr::SdfPathVector>()) {
+        // Special case for SdfPathVector.
+        pxr::SdfPathVector paths = value.Get<pxr::SdfPathVector>();
+        for (pxr::SdfPath const& path : paths) {
+            valueStream << path << "\n";
+        }
+    }
+    else {
+        valueStream << value;
+    }
     std::string expectedValue = valueStream.str();
 
     if (!MatchesFallbackTextOutput(expectedValue)) {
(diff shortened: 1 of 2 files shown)


lilike-adsk:
@ppt-adsk @lanierd-adsk

lib/adskHydraSceneBrowser/test/adskHydraSceneBrowserTestFixture.cpp
```diff
@@ -257,7 +257,16 @@ void AdskHydraSceneBrowserTestFixture::CompareValueContent(const pxr::VtValue& v
     std::string actualValue = valueText.toStdString();
 
     std::ostringstream valueStream;
-    valueStream << value;
+    if (value.IsHolding<pxr::SdfPathVector>()) {
```
ppt-adsk, line 263: What about other vector types?  E.g. vectors of Plain Old Data (POD) types, like integers, floats?

lilike-adsk: For other types, it still keeps unchanged, the Qt widget still present the content in the way of `buffer << _value;`
It's same as what we fetch from dataSource here: `valueStream << value`,

debloip-adsk: Yeah this code's goal is to test that the UI displays the intended result, so we just mimic the behaviour from the actual scene browser code, seen in the [link Keli put in the description of this PR](https://github.com/PixarAnimationStudios/OpenUSD/commit/1d19b1dbdd0daf12247f5a1cb4b858c444e3e4fa#diff-8c7415891d45a3810ea025ac58c7e025043412150b628560c12a1b8fdbe1f830R57). I would definitely add this link in a comment though. Otherwise looks good to me.

debloip-adsk (changes requested):

ppt-adsk (approved):

debloip-adsk (approved):

PR #185, opened 2024-10-15
HYDRA-1199 : Fix crash when switching to another render delegate

lanierd-adsk:


lib/mayaHydra/hydraExtensions/sceneIndex/registration.cpp
```diff
@@ -417,10 +417,22 @@ MayaHydraSceneIndexRegistry::~MayaHydraSceneIndexRegistry()
         MSceneMessage::removeCallback(_AfterOpenCBId);
     }
     _AfterOpenCBId = 0;
+    _RemoveAllSceneIndexNodes();
```
lanierd-adsk, line 420: A ticket has been created to add a unit test once a mtoa build is ready in our automation suite.

debloip-adsk (changes requested):
The comment above `MtohRenderOverride::ClearHydraResources` should be adjusted since the scene index registry is no longer cleared only when `fullReset` is true.

Also, regarding testing, I'm not sure what you mean by when a mtoa build is ready, can't we already run unit tests with mtoa?

Otherwise, looks good, thanks!

lanierd-adsk:
>Also, regarding testing, I'm not sure what you mean by when a mtoa build is ready, can't we already run unit tests with mtoa?
I meant I logged a ticket to update mtoa to a build where they are using openusd 24.08. That is not the case with the current version we use for automation.

debloip-adsk (dismissed):

lanierd-adsk:
FYI : @ppt-adsk @lilike-adsk

lanierd-adsk:
pr-build

lanierd-adsk:
pr-build

lib/mayaHydra/hydraExtensions/sceneIndex/registration.cpp
```diff
@@ -417,10 +417,30 @@ MayaHydraSceneIndexRegistry::~MayaHydraSceneIndexRegistry()
...
     _registrationsByObjectHandle.clear();
     _registrations.clear();
 }
 
+void MayaHydraSceneIndexRegistry::_RemoveAllSceneIndexNodes()
```
lanierd-adsk, line 425: I fixed that function from my first commit, it had an issue when iterating on an array while removing its elements.

debloip-adsk: To reduce code duplication, could we still call `_RemoveSceneIndexForNode`, just keeping the new way of always removing the first element? Something like :

```
while (_registrationsByObjectHandle.begin() != _registrationsByObjectHandle.end()) {
    _RemoveSceneIndexForNode(_registrationsByObjectHandle.begin()->first);
}
```

lanierd-adsk: Ha I see. But I did it like that on purpose to avoid calling _RemoveSceneIndexForNode from a MObjectHandle as from that we try to find it in the array then we take the iterator and we already have all of this in _RemoveAllSceneIndexNodes. So it would be a waste of time.

debloip-adsk: Yeah that's why I think we should strive to re-use `_RemoveSceneIndexForNode` unless there is a functional limitation, having practically the same code in both makes me worried that in the future we might make a change in one but forget the other

lanierd-adsk: You're right in the sense that this code is not critical in term of speed, so let's do as your propose. Thanks.

debloip-adsk (approved):

PR #187, opened 2024-10-15
Add support for display of VtArray<bool> to Hydra scene browser

ppt-adsk:
Very useful to debug instancer masking.


lib/adskHydraSceneBrowser/lib/CMakeLists.txt
```diff
@@ -92,6 +92,13 @@ foreach(SOURCE IN ITEMS ${SOURCES})
     set(SET_EXPANDED_REPLACEMENT "setExpanded\(true\);")
     string(REGEX REPLACE "${SET_EXPANDED_REGEX}" "${SET_EXPANDED_REPLACEMENT}" FILE_CONTENTS "${FILE_CONTENTS}")
 
+    # Patch in support for VtArray<bool>.  Arbitrarily place it after support
```
ppt-adsk, line 95: Somewhat ugly to do this in a loop that goes over all files, but convenient.

debloip-adsk (approved):
Seems like a good candidate for a PR to Pixar as well, do we/should we have a ticket for that?

ppt-adsk:
> Seems like a good candidate for a PR to Pixar as well, do we/should we have a ticket for that?

I have just created HYDRA-1252.

PR #189, opened 2024-10-21
HYDRA-1098 : Viewport filters for Hydra data

debloip-adsk:
This PR adds support for filtering Hydra data using the Maya "Show" menu filter options.


lib/flowViewport/sceneIndex/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ target_sources(${TARGET_NAME}
     fvpWireframeSelectionHighlightSceneIndex.cpp
     fvpDisplayStyleOverrideSceneIndex.cpp
     fvpPruneTexturesSceneIndex.cpp
+    fvpPruningSceneIndex.cpp
```
debloip-adsk, line 16: Note : I went with "pruning" rather than "filtering" to avoid confusion with filtering scene indices

lib/flowViewport/sceneIndex/fvpPruningSceneIndex.cpp
```diff
@@ -0,0 +1,344 @@
...
+        { FvpPruningTokens->meshes, _MeshesFilterHandler },
+        { FvpPruningTokens->capsules, _CapsulesFilterHandler },
+        { FvpPruningTokens->cones, _ConesFilterHandler },
+        { FvpPruningTokens->cubes, _CubesFilterHandler },
+        { FvpPruningTokens->cylinders, _CylindersFilterHandler },
+        { FvpPruningTokens->spheres, _SpheresFilterHandler },
+        { FvpPruningTokens->nurbsCurves, _NurbsCurvesFilterHandler },
+        { FvpPruningTokens->nurbsPatches, _NurbsPatchesFilterHandler }
+    };
+    return filterHandlers[pruningToken](GetInputSceneIndex(), primPath, prim);
```
debloip-adsk, line 129: The basic idea of this scene index is to associate TfTokens with predicates that return whether or not to filter out a given prim, so we can separate out the filtering logic per-filter. Currently we don't really need special checks, so the handlers just use the prim type. 

This mechanism could potentially be exposed externally if there is a need (e.g. to allow plugin developers to add their own filters), though for now there is not so I've kept this internal to the class. 

The rest of the implementation of this class mostly consists of tracking which prims are pruned out or not, and sending/blocking corresponding PrimsAdded/PrimsRemoved notifications.

test/testUtils/fixturesUtils.py
```diff
@@ -20,7 +20,7 @@
 import unittest
 
 # Plugins that are bundled and loaded by default in a Maya installation
-DEFAULT_PLUGINS = ['ArubaTessellator', 'modelingToolkit']
+DEFAULT_PLUGINS = ['ArubaTessellator', 'modelingToolkit', 'drawUfe']
```
debloip-adsk, line 23: drawUfe handles drawing UFE cameras

debloip-adsk (commented):
Note : tests are not up-to-date, waiting on OpenPBR changes.

test/lib/mayaUsd/render/mayaToHydra/testViewportFilters.py
```diff
@@ -222,5 +231,99 @@ def test_Strokes(self):
...
+        from pxr import UsdGeom
+
+        stagePath = mayaUsd_createStageWithNewLayer.createStageWithNewLayer()
+        stage = mayaUsd.lib.GetPrim(stagePath).GetStage()
+
+        capsuleName = "UsdCapsule"
+        capsuleXform = UsdGeom.Xform.Define(stage, "/" + capsuleName + "Xform")
+        capsuleXform.AddTranslateOp().Set(value=(6, 0, 0))
+        UsdGeom.Capsule.Define(stage, str(capsuleXform.GetPath()) + "/" + capsuleName)
+        cmds.select(clear=True)
```
ppt-adsk: Very minor, but why do we need to clear the selection repeatedly?

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -802,6 +804,35 @@ MStatus MtohRenderOverride::Render(
...
+        static const std::map<MUint64, TfTokenVector> mayaFiltersToFvpPruningTokens = {
+            { MHWRender::MFrameContext::kExcludeMeshes, polygonFilters },
+            { MHWRender::MFrameContext::kExcludeNurbsCurves, {FvpPruningTokens->nurbsCurves} },
+            { MHWRender::MFrameContext::kExcludeNurbsSurfaces, {FvpPruningTokens->nurbsPatches} }
+        };
+
+        for (auto [mayaFilter, fvpPruningTokens] : mayaFiltersToFvpPruningTokens) {
+            for (const auto& fvpPruningToken : fvpPruningTokens) {
+                if (objectExclusions & mayaFilter) {
+                    _pruningSceneIndex->EnableFilter(fvpPruningToken);
```
ppt-adsk: Minor: if you'd had an interface with EnableFilter(bool, token) you could have save a couple of lines of code and a bit of awkwardness here:
`_pruningSceneIndex->EnableFilter(objectExclusions & mayaFilter, fvpPruningToken)`

lib/flowViewport/sceneIndex/fvpPruningSceneIndex.cpp
```diff
@@ -0,0 +1,344 @@
...
+}
+
+bool _MeshesFilterHandler(const HdSceneIndexBaseRefPtr& sceneIndex, const SdfPath& primPath, const HdSceneIndexPrim& prim)
+{
+    // Currently we just flat out remove any prim with a mesh type. If we were to add extra checks to make sure this is not
+    // a mesh prim that serves another purpose, we would add them here.
+    return prim.primType == HdPrimTypeTokens->mesh;
+}
+
+bool _CapsulesFilterHandler(const HdSceneIndexBaseRefPtr& sceneIndex, const SdfPath& primPath, const HdSceneIndexPrim& prim)
```
ppt-adsk, line 53: Minor: isn't it redundant to pass in both the prim and the primPath?  Since you're passing in the sceneIndex, you can derive the prim from the primPath.

ppt-adsk (dismissed):

test/lib/mayaUsd/render/mayaToHydra/testViewportFilters.py
debloip-adsk: No reason. Updated to keep only one before the image diffing.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
debloip-adsk: Made the Enable/Disable methods private and changed the public API to a SetFilterStatus method. Also removed the bool return value since that wasn't really useful.

lib/flowViewport/sceneIndex/fvpPruningSceneIndex.cpp
debloip-adsk: Yes, but since the calling code has already retrieved the prim anyways, not passing it would mean the filters would have to get the prim again, so it'd be redundant in that way (and depending on the upstream scene indices, might incur a performance cost?). And since I feel most filters would end up getting the prim anyways since you can't do much only off a path, it feels simpler to me to just pass it from the get-go.

debloip-adsk:
FYI @lanierd-adsk @lilike-adsk

ppt-adsk (approved):

PR #197, opened 2024-11-06
Moved isolate select scene index creation to avoid losing isolate sel…

ppt-adsk:
…ection.

### DIFF
--- a/lib/mayaHydra/mayaPlugin/renderOverride.cpp
+++ b/lib/mayaHydra/mayaPlugin/renderOverride.cpp
@@ -1096,6 +1096,22 @@ void MtohRenderOverride::_InitHydraResources(const MHWRender::MDrawContext& draw
     _dirtyLeadObjectSceneIndex = MAYAHYDRA_NS::MhDirtyLeadObjectSceneIndex::New(_inputSceneIndexOfFilteringSceneIndicesChain);
     _inputSceneIndexOfFilteringSceneIndicesChain = _dirtyLeadObjectSceneIndex;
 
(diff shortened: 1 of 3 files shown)


lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1096,6 +1096,22 @@ void MtohRenderOverride::_InitHydraResources(const MHWRender::MDrawContext& draw
     _dirtyLeadObjectSceneIndex = MAYAHYDRA_NS::MhDirtyLeadObjectSceneIndex::New(_inputSceneIndexOfFilteringSceneIndicesChain);
     _inputSceneIndexOfFilteringSceneIndicesChain = _dirtyLeadObjectSceneIndex;
 
+#ifdef MAYA_HAS_VIEW_SELECTED_OBJECT_API
```
ppt-adsk, line 1099: Moved from _CreateSceneIndicesChainAfterMergingSceneIndex().  The isolate select scene index is independent of display style and therefore does not need to be re-created on shading change (e.g. bounding box mode).

test/lib/mayaUsd/render/mayaToHydra/cpp/testUsdPointInstancingIsolateSelectBBox.py
ppt-adsk, line 1: Without the fix, the test fails.  With the fix, the test passes.

lanierd-adsk (approved):

PR #200, opened 2024-11-12
Disable mayaHydra isolate select when VP2 is the viewport renderer.

ppt-adsk:


test/lib/mayaUsd/render/mayaToHydra/cpp/testIsolateSelectSwitchToVP2.py
ppt-adsk, line 1: With the fix, the test passes.  Without the fix, the test fails.

test/lib/mayaUsd/render/mayaToHydra/cpp/testIsolateSelectSwitchToVP2.py
```diff
@@ -0,0 +1,163 @@
...
+        # Hydra view selected changed calls should remain the same.
+        self.setViewport2Renderer()
+
+        cmds.select('|pSphere1')
+        cmds.editor(modelPanel, edit=True, updateMainConnection=True)
+        cmds.isolateSelect(modelPanel, loadSelected=True)
+
+        cmds.refresh()
+
+        viewSelectedChangedCallsVP2 = cmds.mayaHydraInstruments("MtohRenderOverride:NbViewSelectedChangedCalls", q=True)
```
ppt-adsk, line 136: When switching to VP2 and performing isolate select, the mayaHydra callback should early out.

lanierd-adsk (approved):

PR #201, opened 2024-11-14
HYDRA-1291 : Fix default maya material is not picked up correctly

lanierd-adsk:


lib/mayaHydra/hydraExtensions/adapters/renderItemAdapter.cpp
```diff
@@ -33,11 +33,7 @@
 #include <pxr/usdImaging/usdImaging/tokens.h>
 
 #include <maya/MAnimControl.h>
-#include <maya/MDGContext.h>
```
lanierd-adsk, line 36: I did a bit of headers cleaning in this PR.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -773,26 +769,20 @@ VtValue MayaHydraSceneIndex::GetMaterialResource(const SdfPath& id)
...
-        }
-    }
-
-    if (! defaultMaterialSuccessfullyCreated){
-        TF_CODING_WARNING("standardSurface1 material and its shading group could not be retrieved, using a fallback material");
-        // In case we could not create the default material from the standardSurface1 material, we
-        // create a fallback material
+    bool defaultMaterialCreatedSuccessfully = false;
+
+    MObject defaultMaterialShadingGroupObj = getDefaultMaterialShadingGroupNode();
```
lilike-adsk: I recall there's already an Maya API (MMaterial::defaultMaterial()) to get the default Material, no need to query again. Just use MObject defaultMaterialShadingGroupObj = MMaterial::defaultMaterial().shadingEngine();

lilike-adsk: here's the API doc:
https://help.autodesk.com/view/MAYAUL/2024/ENU/?guid=MAYA_API_REF_cpp_ref_class_m_material_html

static MMaterial defaultMaterial()

Get the default material.
There will always be a default material in the scene and therefore the result of this function should always succeed. The default material will correspond to the initialShadingGroup node that is in the scene.

lib/mayaHydra/hydraExtensions/mayaUtils.cpp
```diff
@@ -18,46 +18,11 @@
 
 #include <maya/MDagPath.h>
 #include <maya/MFnDagNode.h>
-#include <maya/MGlobal.h>
 #include <maya/MMatrix.h>
 #include <maya/MPlug.h>
-#include <maya/MPlugArray.h>
 #include <maya/MSelectionList.h>
 #include <maya/MObjectArray.h>
-#include <maya/MFnAttribute.h>
```
lanierd-adsk, line 27: All that has been removed is not used any more

lilike-adsk (approved):
nice cleanup

PR #206, opened 2024-11-21
HYDRA-1275 Prepare code and unit test for usd 24.11

lanierd-adsk:


cmake/modules/FindUSD.cmake
```diff
@@ -107,23 +107,6 @@ find_library(USD_LIBRARY
 
 get_filename_component(USD_LIBRARY_DIR ${USD_LIBRARY} DIRECTORY)
 
-# Get the boost version from the one built with USD
```
lanierd-adsk, line 110: This is not used at all, we don't use Boost and it was removed in usd 24.11+

test/lib/mayaUsd/render/mayaToHydra/cpp/testIsolateSelect.py
```diff
@@ -53,6 +54,14 @@ class TestIsolateSelect(mtohUtils.MayaHydraBaseTestCase):
 
     _requiredPlugins = ['mayaHydraCppTests', 'mayaHydraFlowViewportAPILocator']
 
+    imageVersion = None
+
+    @classmethod
```
lanierd-adsk, line 58: For the unit tests, there are basically 2 possible workarounds to keep the same images as much as possible.
1. modifyDefaultLightIntensityIfUsdGreaterOrEqualTo_24_11() / resetDefaultLightIntensityIfUsdGreaterOrEqualTo_24_11()  to modify/reset the default light intensity which fixes most of the problematic tests.
2. When 1. doesn't work I need to do another baseline image and I am using the imageVersion from assertsnapshot where the new version of the image is in a given folder whose name is imageVersion (when it's not None).

lib/mayaHydra/hydraExtensions/adapters/lightAdapter.cpp
```diff
@@ -232,7 +237,13 @@ VtValue MayaHydraLightAdapter::GetLightParamValue(const TfToken& paramName)
         const auto color = light.color();
         return VtValue(GfVec3f(color.r, color.g, color.b));
     } else if (paramName == HdLightTokens->intensity) {
-        return VtValue(light.intensity());
+        float intensity = light.intensity();
```
debloip-adsk: Minor : auto is used in the surrounding code, could use it here as well for consistency

test/testUtils/mtohUtils.py
```diff
@@ -138,6 +139,23 @@ def setBasicCam(self, dist=DEFAULT_CAM_DIST):
...
+                cmds.setAttr('hardwareRenderingGlobals.defaultLightIntensity', 3.1415)
+
+    def resetDefaultLightIntensityIfUsdGreaterOrEqualTo_24_11(self):
+        #For Usd 24.11+ reset the default light intensity to 1 instead of PI to counter balance the changes in usd 24.11
+        if Usd.GetVersion() >= (0, 24, 11):
+            if maya.mel.eval("optionVar -exists defaultLightIntensity"):
+                maya.mel.eval("optionVar -fv defaultLightIntensity 1.0")
+            if cmds.attributeQuery('defaultLightIntensity', node='hardwareRenderingGlobals', exists=True):
+                cmds.setAttr('hardwareRenderingGlobals.defaultLightIntensity', 1.0)
+
```
debloip-adsk: These are pretty much copy-paste, so I'd suggest extracting out a common method to set the default light intensity

test/lib/mayaUsd/render/mayaToHydra/cpp/testIsolateSelect.py
debloip-adsk: For 1. : Could we change the modify/reset methods to be more generic/modifiable over time than them being fixed to a single version, since I presume these things could change again? i.e. have something like setDefaultLightIntensityByUsdVersion and resetDefaultLightIntensity, in which we could handle all the different USD versions 

For 2. : Could we move the imageVersion class attribute and setup to the MayaHydraBaseTestCase in mtohUtils.py? That way the usd version would always be set automatically, can be used as needed and the code will not have to be copy-pasted in each test

debloip-adsk (changes requested):

test/lib/mayaUsd/render/mayaToHydra/cpp/testIsolateSelect.py
```diff
@@ -58,9 +57,9 @@ class TestIsolateSelect(mtohUtils.MayaHydraBaseTestCase):
 
     @classmethod
     def setUpClass(cls):
-        if Usd.GetVersion() >= (0, 24, 11):
-            cls.imageVersion = 'usd_2411+'
         super(TestIsolateSelect, cls).setUpClass()
+        if cls._usdVersion >= (0, 24, 11):
+            cls.imageVersion = 'usd_2411+'
```
debloip-adsk, line 62: What I had in mind was actually to move the imageVersion attribute itself to `MayaHydraBaseTestCase`, but I can see how not all tests might want to use that as their image version(s)

debloip-adsk (approved):

PR #211, opened 2024-11-26
HYDRA-909 - remove lighted prefix from hydra scene hierarchy

lilike-adsk:
To remove "lighted" prefix from hydra scene hierarchy, we collect all prims that need lighting/shadowing and put them in a SdfPathVector, and set to shadowCollection.

Impact: when shadow is on, modification to the scene might be slow a bit.
Peformance testing (adding prim into a scene with 10000 prims when shadow is on and tumbling scene with 10000 prims when shadow is on) shows the impact is trivial.


lib/mayaHydra/hydraExtensions/adapters/lightAdapter.cpp
```diff
@@ -154,12 +154,7 @@ VtValue MayaHydraLightAdapter::Get(const TfToken& key)
         MFnLight       mayaLight(GetDagPath());
         GlfSimpleLight light;
         const auto     color = mayaLight.color();
-        auto     intensity   = mayaLight.intensity();
```
lanierd-adsk: You probably don't have the up to date file as this shouldn't be removed, unless you moved it somewhere else ?

lilike-adsk: yes, need merge with latest

lilike-adsk: @lanierd-adsk you can wait until it's ready, still need to update tests and polish, and do performance tests. :) will let you know when it's ready

lanierd-adsk: Sounds good. Thanks.

lib/flowViewport/sceneIndex/fvpWireframeSelectionHighlightSceneIndex.cpp
```diff
@@ -333,6 +333,11 @@ WireframeSelectionHighlightSceneIndex(
     TF_AXIOM(_wireframeColorInterface);
 
     auto operation = [this](const SdfPath& primPath, const HdSceneIndexPrim& prim) -> bool {
+
+        if (_IsExcluded(primPath)) {
```
lilike-adsk, line 337: Grab the fix from Pierre, as removing Lighted hierarcky will expose this potential issue

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -1424,6 +1422,24 @@ void MayaHydraSceneIndex::_RemoveRenderItem(const MayaHydraRenderItemAdapterPtr&
...
+void MayaHydraSceneIndex::GetLightedPrimPaths(SdfPathVector& lightedPrimPaths)
+{
+    _MapAdapter<MayaHydraRenderItemAdapter>(
+        [&](MayaHydraRenderItemAdapter* a) {
+            if (a->GetPrimitive() == MHWRender::MGeometry::Primitive::kTriangles) {
+                lightedPrimPaths.emplace_back(a->GetID());
+            }
+        },
+        _renderItemsAdapters);
+    _MapAdapter<MayaHydraDagAdapter>(
```
lilike-adsk: This is for meshAdapter

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.h
```diff
@@ -305,16 +312,6 @@ class MAYAHYDRALIB_API MayaHydraSceneIndex : public HdRetainedSceneIndex, public
     void _Destroy();
 
 private:
-    // ------------------------------------------------------------------------
```
lilike-adsk, line 308: remove dead code

lilike-adsk:
@lanierd-adsk the change is updated, you can start review now, thanks!

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -1424,6 +1422,24 @@ void MayaHydraSceneIndex::_RemoveRenderItem(const MayaHydraRenderItemAdapterPtr&
...
+        [&](MayaHydraDagAdapter* a) {
+            if (a->HasType(HdPrimTypeTokens->mesh)) {
```
lanierd-adsk: I think this is too restrictive, we should include all Renderable Prims with something like :
// This includes implicit primitive types.
static const VtArray<TfToken> GEOMETRY_PRIM_TYPES = {
    HdRprimTypeTokens->allTokens.begin(),
    HdRprimTypeTokens->allTokens.end()
};

For example, volumes should receive the lighting, I think also curves and points if they support it.
I would put this in a function so we can update it if we want like 
 const VtArray<TfToken> GetLightedPrimitivesType()

lanierd-adsk:
This is only for maya native data right ? What about custom data producers prims and usd prims ?
If we collect the data to be lighted by prim types, could we use a filtering scene index to traverse the whole scene prims (usd+maya+custom) and collect the lighted prims ?

lanierd-adsk (changes requested):

lilike-adsk:
> This is only for maya native data right ? What about custom data producers prims and usd prims ? If we collect the data to be lighted by prim types, could we use a filtering scene index to traverse the whole scene prims (usd+maya+custom) and collect the lighted prims ?

Yes, the original "Lighted" hierarchy is only for native data.  It could be possible to use a SI to support shadows for all data, put the idea in USD shadow ticket: https://jira.autodesk.com/browse/HYDRA-1269

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
lilike-adsk: Good idea, I'll try it

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -237,25 +235,6 @@ namespace {
         return sdfDagPath.AppendPath(inPath);
     }
 
-    ///Returns false if this object should not be lighted, true if it should be lighted
```
lilike-adsk, line 240: The logics are moved to adapter::WantBeLighted() function

lib/mayaHydra/hydraExtensions/adapters/adapter.h
```diff
@@ -117,6 +117,8 @@ class MayaHydraAdapter
     MAYAHYDRALIB_API
     virtual GfVec4f GetDisplayColor() const { return {1.f,1.f,1.f,1.f}; }
 
+    virtual bool WantBeLighted() const { return false; }
```
lanierd-adsk: Nice to have this change, but I would rather call it Illuminated()

lanierd-adsk (changes requested):

lanierd-adsk (approved):

PR #216, opened 2024-12-02
Loading mayaHydra should not set scene as modified.

ppt-adsk:


test/lib/mayaUsd/render/mayaToHydra/testSceneModified.py
ppt-adsk, line 1: Trivial test for modified scene.

test/lib/mayaUsd/render/mayaToHydra/testPolygonPrimitives.py
```diff
@@ -26,6 +26,8 @@ class TestPolygonPrimitives(mtohUtils.MayaHydraBaseTestCase):
     IMAGE_DIFF_FAIL_THRESHOLD = 0.05
     IMAGE_DIFF_FAIL_PERCENT = 1.5
 
+    _requiredPlugins = ['modelingToolkit']
```
ppt-adsk, line 29: This plugin was being loaded for all tests, in fixturesUtils.py, but only this test was using it.

test/testUtils/fixturesUtils.py
```diff
@@ -19,9 +19,6 @@
 import sys
 import unittest
 
-# Plugins that are bundled and loaded by default in a Maya installation
```
ppt-adsk, line 22: ArubaTesselator plugin is unused by our tests, and modelingToolkit was only used by a single test.

test/testUtils/fixturesUtils.py
```diff
@@ -32,10 +29,10 @@ def _setUpClass(modulePathName, pluginName, initializeStandalone):
 
     if pluginName:
         import maya.cmds as cmds
+        wasModified = cmds.file(query=True, modified=True)
         cmds.loadPlugin(pluginName, quiet=True)
-
-    for defaultPlugin in DEFAULT_PLUGINS:
-        cmds.loadPlugin(defaultPlugin, quiet=True)
+        isModified = cmds.file(query=True, modified=True)
+        assert isModified == wasModified, ('Loading plugin %s modified the scene' % pluginName)
```
ppt-adsk, line 35: Assert that loading the test fixture plugin does not mark the scene as modified.

test/testUtils/mayaUtils.py
```diff
@@ -43,61 +43,6 @@
 HD_STORM = "HdStormRendererPlugin"
 HD_STORM_OVERRIDE = "mayaHydraRenderOverride_" + HD_STORM
 
-def loadPlugin(pluginName):
```
ppt-adsk, line 46: Removed dead code.

test/testUtils/mtohUtils.py
```diff
@@ -35,13 +35,6 @@
 HD_STORM_OVERRIDE = "mayaHydraRenderOverride_" + HD_STORM
 MAYAUSD_PLUGIN_NAME = 'mayaUsdPlugin'
 
-def checkForPlugin(pluginName: str):
```
ppt-adsk, line 38: Removed dead code.

lib/mayaHydra/mayaPlugin/plugin.cpp
```diff
@@ -78,6 +79,44 @@ namespace {
     setenv(name.c_str(), value.c_str(), 1);
     #endif
     }
+
+class SceneModifiedGuard
```
ppt-adsk, line 83: Core change in the pull request.  Create an RAII guard class to set the scene as unmodified after loading the mayaHydra plugin.  Plugins often add default nodes on being loaded.  This conceptually does not change the modified state of the scene.

test/testUtils/mtohUtils.py
```diff
@@ -106,8 +101,15 @@ def setUp(self):
...
         # from previous tests.
         mayaUtils.openNewScene()
+        modified = cmds.file(query=True, modified=True)
+        assert not modified, 'Internal test framework error: scene left as modified by mayaUtils.openNewScene()'
+
         self.setHdStormRenderer()
 
+        # We've just opened a new scene, so we should not be modified.  Setting
+        # Storm should conceptually not change that status.
+        cmds.file(modified=False)
```
debloip-adsk: Should this be an assert? The comment makes it sound like this is checking that setting Storm does not change the modified state, but this just forces it to false

ppt-adsk: It is setting it to unmodified, because the line above (setHdStormRenderer()) modifies the file.  I can update the comment if you want.

debloip-adsk: What confuses me is the contradiction between : 
> [...] we should not be modified. Setting Storm should conceptually not change that status

and
> (setHdStormRenderer()) modifies the file

It's unclear to me whether Storm should mark the file as modified. If it should, then why do force the flag to false, and if it shouldn't, then should this not be an assert, or otherwise point to a problem where setting Storm does change the modified flag without us wanting it to do so?

debloip-adsk (dismissed):

debloip-adsk (approved):

PR #231, opened 2024-12-12
HYDRA-1325 : Fix when no extent attribute is provided in a prim in th…

lanierd-adsk:
…e bounding box scene index


laferra-adsk (approved):
Looks good! I will adapt my code with your improvements. Txs

PR #232, opened 2024-12-13
HYDRA-1336 : Change specular Color to have an update when changing th…

lanierd-adsk:
…e FVP color parameter of the custom GLSL shader when using only indirect lighting


lilike-adsk (dismissed):

lilike-adsk:
Seems we need to rebaseline the images

lilike-adsk (approved):

PR #239, opened 2025-01-20
Reduced dirty notifications for isolate select.

ppt-adsk:
Makes isolate select notifications easier to understand, and potentially improves performance by avoiding pulling on data when not necessary.  Performance has not been measured, only correctness.

### DIFF
--- a/lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportDataManager.cpp
+++ b/lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportDataManager.cpp
@@ -174,6 +174,7 @@ void ViewportInformationAndSceneIndicesPerViewportDataManager::DisableIsolateSel
 )
 {
     _isolateSelection[viewportId] = nullptr;
+    _isolateSelectSceneIndex->SetViewport(viewportId, nullptr);
 }
 
 SelectionPtr
--- a/lib/flowViewport/sceneIndex/fvpIsolateSelectSceneIndex.cpp
+++ b/lib/flowViewport/sceneIndex/fvpIsolateSelectSceneIndex.cpp
@@ -108,6 +108,18 @@ bool isGeomSubset(const HdSceneIndexPrim& prim) {
 #endif
 }
 
+HdContainerDataSourceHandle setInstancerMaskDataSource(
+    const HdSceneIndexPrim& inputPrim,
+    const VtArray<bool>&    instanceMask
+)
+{
+    auto maskDs = HdRetainedTypedSampledDataSource<VtArray<bool>>::New(instanceMask);
+
+    return HdContainerDataSourceEditor(inputPrim.dataSource)
+        .Set(instancerMaskLocator, maskDs)
+        .Finish();
+}
+
 }
 
 namespace FVP_NS_DEF {
@@ -159,11 +171,8 @@ HdSceneIndexPrim IsolateSelectSceneIndex::GetPrim(const SdfPath& primPath) const
 
     auto instancerMask = _instancerMasks.find(primPath);
     if (instancerMask != _instancerMasks.end()) {
-        auto maskDs = HdRetainedTypedSampledDataSource<VtArray<bool>>::New(instancerMask->second);
-
-        inputPrim.dataSource = HdContainerDataSourceEditor(inputPrim.dataSource)
-            .Set(instancerMaskLocator, maskDs)
-            .Finish();
+        inputPrim.dataSource = 
+            setInstancerMaskDataSource(inputPrim, instancerMask->second);
 
         return inputPrim;
     }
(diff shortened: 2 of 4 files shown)


lib/flowViewport/API/perViewportSceneIndicesData/fvpViewportInformationAndSceneIndicesPerViewportDataManager.cpp
```diff
@@ -174,6 +174,7 @@ void ViewportInformationAndSceneIndicesPerViewportDataManager::DisableIsolateSel
 )
 {
     _isolateSelection[viewportId] = nullptr;
+    _isolateSelectSceneIndex->SetViewport(viewportId, nullptr);
```
ppt-adsk, line 177: Fixed latent bug that was being compensated for in renderOverride.cpp.  The Flow Viewport Toolkit should not rely on clients to manage the isolate select scene index properly when it can do so itself.

lib/flowViewport/sceneIndex/fvpIsolateSelectSceneIndex.cpp
ppt-adsk, line 1: Core file in the review.  Instancers that were completely excluded were being excluded by virtue of their prototype visibility being set false, which is wasteful.  Now, we simply use the instancer mask to make all instances invisible.

lib/flowViewport/sceneIndex/fvpIsolateSelectSceneIndex.cpp
```diff
@@ -159,11 +171,8 @@ HdSceneIndexPrim IsolateSelectSceneIndex::GetPrim(const SdfPath& primPath) const
 
     auto instancerMask = _instancerMasks.find(primPath);
     if (instancerMask != _instancerMasks.end()) {
-        auto maskDs = HdRetainedTypedSampledDataSource<VtArray<bool>>::New(instancerMask->second);
-
-        inputPrim.dataSource = HdContainerDataSourceEditor(inputPrim.dataSource)
-            .Set(instancerMaskLocator, maskDs)
-            .Finish();
+        inputPrim.dataSource =
```
ppt-adsk, line 174: Factored out to unnamed namespace.

lib/flowViewport/sceneIndex/fvpIsolateSelectSceneIndex.cpp
```diff
@@ -179,6 +188,20 @@ HdSceneIndexPrim IsolateSelectSceneIndex::GetPrim(const SdfPath& primPath) const
     // HYDRA-1242: setting visibility on GeomSubset prim causes hang in Hydra
     // Storm.
     if (!included && !isGeomSubset(inputPrim)) {
+        // If an instancer is not included, none of its instances are, so set
```
ppt-adsk, line 191: Mark all instances as invisible by using the instancer mask.  We still set the instancer itself as invisible, even though for Hydra Storm this does not seem to have any effect.

lib/flowViewport/sceneIndex/fvpIsolateSelectSceneIndex.cpp
```diff
@@ -437,21 +460,43 @@ void IsolateSelectSceneIndex::_DirtyVisibilityRecursive(
     HdSceneIndexObserver::DirtiedPrimEntries* dirtiedEntries
 ) const
 {
-    TF_DEBUG(FVP_ISOLATE_SELECT_SCENE_INDEX)
-        .Msg("            %s: marking %s visibility locator dirty.\n", _viewportId.c_str(), primPath.GetText());
-
+    // If the prim is a material, early out: visibility is not
+    // relevant for materials
     auto prim = GetInputSceneIndex()->GetPrim(primPath);
+    if (prim.primType == HdPrimTypeTokens->material) {
```
ppt-adsk, line 466: Don't set materials as invisible, and don't recurse.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
ppt-adsk, line 1: Deal with awkward Maya notification system that uses a boolean to distinguish between isolate select state changes and isolate select object changes, sending two notifications on first transition from on to off.  We account for this and only dirty once.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1831,15 +1831,50 @@ void MtohRenderOverride::_ViewSelectedChangedCb(
...
+    // pendingObjects: message true --> notify, go to on.
+    // 
+    // on: message false --> go to off, notify.
+    // on: message true --> stay on, notify.
+
+    if (found->second == IsolateSelectState::IsolateSelectOff) {
+        if (TF_VERIFY(!viewSelectedObjectsChanged)) {
+            found->second = IsolateSelectState::IsolateSelectPendingObjects;
+        }
+        return;
```
ppt-adsk, line 1867: Early out and don't notify, wait for the second notification.

lib/mayaHydra/mayaPlugin/renderOverride.h
```diff
@@ -338,6 +339,21 @@ class MtohRenderOverride : public MHWRender::MRenderOverride,
...
+    // isolate select off to on, two notifications are therefore sent,
+    // first false (state change), then true (objects set).  To avoid
+    // double dirtying in Hydra, we track the following isolate select
+    // states per viewport:
+    // 
+    enum class IsolateSelectState {IsolateSelectOff, IsolateSelectPendingObjects,
+				   IsolateSelectOn};
+
+    using VpIsolateSelectStates = std::map<std::string, IsolateSelectState>;
+    VpIsolateSelectStates _isolateSelectState;
```
ppt-adsk, line 356: Per-viewport isolate select notification state tracking.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1831,15 +1831,50 @@ void MtohRenderOverride::_ViewSelectedChangedCb(
...
+    //
+    // State transitions:
+    // 
+    // off: message false --> go to pendingObjects, do not notify.
+    // off: message true --> illegal, warn, do not notify.
+    // 
+    // pendingObjects: message false --> illegal, warn, do not notify.
+    // pendingObjects: message true --> notify, go to on.
+    // 
+    // on: message false --> go to off, notify.
```
debloip-adsk, line 1860: Just to be sure, is line 1845 handling this case? Asking since "message" in this comment corresponds to `viewSelectedObjectsChanged`, but when IsolateSelectState is already on, I don't see any code that sets it to off based on the value of `viewSelectedObjectsChanged`, only based on `view.viewSelected()`. Maybe one implies the other but just want to make sure.

debloip-adsk (commented):
One quick question, otherwise looking good, thanks!

lib/mayaHydra/mayaPlugin/renderOverride.cpp
ppt-adsk: Correct, I am assuming that view.viewSelected() false is consistent with a transition from isolate select on to a message with viewSelectedObjectsChanged false.  Empirical validation says this is the case, so line 1845 is handling it.

debloip-adsk (approved):

PR #240, opened 2025-01-21
HYDRA-1271 : Selection highlighting rework

debloip-adsk:
Rework of the selection highlighting infrastructure.

Instead of having a single scene index handling all cases, the logic is now split up across multiple scene indices, each dedicated to handling highlights for a specific type of prim. The common logic between them is extracted out in a base class, which makes each scene index independent of each other. See the BaseWhSi.h doc comment for an explanation on how selection highlights are structured.

Note that sets & maps keyed by SdfPaths are heavily used in order to accelerate finding the children and/or parents of paths among a given collection. Generally speaking, lower_bound will be used to find children of a given path (or the path itself), while upper_bound followed by an iterator decrement will be used to find a parent of a path.

Existing highlighting tests were replaced by image-based tests, and new tests were added for native instancing highlighting. Reasoning behind going with image-based tests is that we don't care all that much about how the highlight prims are structured as opposed to how the result looks. And since there are a lot of cases to handle, it is much faster & simpler to go with image-based tests (and it also allows us to easily see if a case actually works or not).

Possible improvements/alternatives for the future :
- Currently, to handle wireframe color changes, the `MhDirtyLeadObjectSceneIndex` also forwards the dirty notifications to instancers' prototypes. However, this could be handled by the instancing-related highlight scene indices instead in order to reduce the amount of notifications sent.
- Extract out the common logic between the point instancer scene index and point instance prototype scene index.
- The mesh scene index currently includes all of the mesh's parents and children in its sub-hierarchy, but we could potentially do like the geomSubset scene index and only return the mesh itself.
- Extract out the shared logic found in the `BaseWhSi` class into a single driver class, which would control the specialized scene indices. This would avoid having each scene index repeat the processing of the `BaseWhSi` and maintain separate copies of the fully selected paths.
- Instead of tracking fully selected paths in a standalone manner and re-implementing `HasFullySelectedAncestorInclusive`, hook the scene indices to the Flow Viewport selection and use the `Fvp::Selection`'s version of the method. This would also require the `Fvp::Selection` to send out notifications whenever a path starts or stops being fully selected to accomodate `ProcessFullySelectedChange`.
- HYDRA-1407 : When both a mesh and a geomSubset highlight coexist, the mesh highlight will always show up on top of the geomSubset highlight, which can lead to incorrect colors.


lib/flowViewport/fvpWireframeColorInterface.h
```diff
@@ -33,6 +34,7 @@ class WireframeColorInterface
 public:
     //Get the wireframe color of a primitive for selection highlighting
     virtual PXR_NS::GfVec4f getWireframeColor(const PXR_NS::SdfPath& primPath) const = 0;
+    virtual PXR_NS::GfVec4f getWireframeColor(const Fvp::PrimSelection& primSelection) const = 0;
```
debloip-adsk, line 37: Needed to distinguish lead vs. active colors for different instances in the same point instancer.

lib/flowViewport/sceneIndex/wireframeHighlights/baseWhSi.cpp
```diff
@@ -0,0 +1,764 @@
...
+#include <pxr/imaging/hd/sceneIndexPrimView.h>
+#include <pxr/imaging/hd/selectionSchema.h>
+#include <pxr/imaging/hd/selectionsSchema.h>
+#include <pxr/imaging/hd/tokens.h>
+
+#include <unordered_set>
+
+PXR_NAMESPACE_USING_DIRECTIVE
+
+namespace {
```
debloip-adsk: Contents of the anonymous namespace were copied over as-is from the previous `fvpWireframeSelectionHighlightSceneIndex`. The only exception is the data source classes, which have been renamed. The `RepathingContainerDataSource`, which was previously `_SelectionHighlightRepathingContainerDataSource`, has been moved out to be re-used across files.

lib/flowViewport/sceneIndex/wireframeHighlights/baseWhSi.cpp
```diff
@@ -0,0 +1,764 @@
...
+                childSource)) {
+        return _RerootingSceneIndexPathArrayDataSource::New(
+            _srcPrefix, _dstPrefix, childPathArrayDataSource);
+    }
+
+    return childSource;
+}
+
+//We want to set the displayStyle of the selected prim to refinedWireOnSurf only if the displayStyle of the prim is refined (meaning shaded)
+HdContainerDataSourceHandle SetWireframeRepr(const HdContainerDataSourceHandle& dataSource, const GfVec4f& color)
```
debloip-adsk: Copied as-is. Was previously `_HighlightSelectedPrim`

lib/mayaHydra/hydraExtensions/mhLeadObjectPathTracker.h
```diff
@@ -52,10 +58,10 @@ class MhLeadObjectPathTracker
     void setLeadObjectUfePath(const Ufe::Path& newLeadObjectUfePath);
 
     MAYAHYDRALIB_API
-    void updatePrimPaths(); // For example : this is called after the data producer scene indices are loaded
+    void updatePrimSelections(); // For example : this is called after the data producer scene indices are loaded
 
 private:
-    PXR_NS::SdfPathVector           _leadObjectPrimPaths;
+    Fvp::PrimSelections             _leadObjectPrimSelections;
```
debloip-adsk, line 64: All changes related to this class are to track lead selections by actual selections and not just paths, in order to differentiate between different instances of a common point instancer.

lib/mayaHydra/hydraExtensions/mhWireframeColorInterfaceImp.h
```diff
@@ -50,10 +50,14 @@ class MhWireframeColorInterfaceImp : public Fvp::WireframeColorInterface
...
 
+    MAYAHYDRALIB_API
+    PXR_NS::GfVec4f getWireframeColor(const Fvp::PrimSelection& primSelection) const override;
+
 private:
     enum SelectionState {kLead, kActive, kDormant};
 
-    SelectionState _getSelectionState(const PXR_NS::SdfPath& primPath)const;
+    SelectionState _getSelectionState(const PXR_NS::SdfPath& primPath) const;
+    SelectionState _getSelectionState(const Fvp::PrimSelection& primSelection) const;
```
debloip-adsk: Needed to distinguish lead vs. active colors for different instances in the same point instancer.

lib/mayaHydra/hydraExtensions/sceneIndex/mhDirtyLeadObjectSceneIndex.cpp
```diff
@@ -40,30 +44,43 @@ namespace {
...
+            pathsToDirty.push(childPath);
+        }
+        
+        HdSceneIndexPrim currPrim = GetInputSceneIndex()->GetPrim(currPathToDirty);
+        if (currPrim.primType == HdPrimTypeTokens->instancer) {
+            HdInstancerTopologySchema instancerTopology = HdInstancerTopologySchema::GetFromParent(currPrim.dataSource);
+            for (const auto& prototypePath : instancerTopology.GetPrototypes()->GetTypedValue(0)) {
+                pathsToDirty.push(prototypePath);
+            }
+        }
```
debloip-adsk, line 83: Potential performance improvement : this could be handled by the instancing-related highlight scene indices to reduce the number of notifications sent.

lib/flowViewport/sceneIndex/wireframeHighlights/piInstancerWhSi.h
```diff
@@ -0,0 +1,117 @@
...
+    void ProcessFullySelectedChange(const PXR_NS::SdfPath& primPath, bool isFullySelected) override;
+
+private:
+    struct SelectionData {
+        PrimSelection _primSelection;
+        PXR_NS::SdfPathSet _instancerPaths;
+        PXR_NS::SdfPathSet _prototypePaths;
+    };
+
+    std::set<PXR_NS::SdfPath> _pointInstancerPaths;
```
debloip-adsk: piInstancerWhSi and piPrototypeWhSi are eerily similar; the main difference between the two is that piInstancerWhSi needs to create highlights when a parent of an instancer is selected (which is why this set exists), which is not the case for piPrototypeWhSi who only cares if prototype prims are directly selected. piPrototypeWhSi also needs to handle geomSubsets (as they can be selected prototypes), but not piInstancerWhSi.

test/lib/mayaUsd/render/mayaToHydra/cpp/testUtils.cpp
```diff
@@ -463,69 +463,6 @@ QPoint getPrimMouseCoords(const HdSceneIndexPrim& prim, M3dView& view)
     return QPoint(viewportX, view.portHeight() - viewportY);
 }
 
-// This method takes in a path to a prim in a selection highlight hierarchy and ensures that 
-// the selection highlight graph is structured properly, and that the leaf mesh prims have
-// the proper display style.
-void assertSelectionHighlightCorrectness(
```
debloip-adsk, line 469: Old utility method for the previous version of selection highlighting.

test/lib/mayaUsd/render/mayaToHydra/cpp/testGeomSubsetsWireframeHighlight.py
debloip-adsk, line 1: Moved to an image-diff test.

test/lib/mayaUsd/render/mayaToHydra/cpp/testPointInstancingWireframeHighlight.py
debloip-adsk, line 1: Moved to an image-diff test.

debloip-adsk:
@ppt-adsk I fully expect you'll have questions, so don't hesitate to reach out, it'll allow us to maybe simplify or at least add more clarifying comments to the code.

lib/mayaHydra/hydraExtensions/mhWireframeColorInterfaceImp.cpp
```diff
@@ -57,6 +57,20 @@ MhWireframeColorInterfaceImp::SelectionState MhWireframeColorInterfaceImp::_getS
     return kDormant;
 }
 
+MhWireframeColorInterfaceImp::SelectionState MhWireframeColorInterfaceImp::_getSelectionState(const Fvp::PrimSelection& primSelection) const
```
ppt-adsk, line 60: Is the following overkill?
```
template<class T>
struct SelectionWrapper {
      SelectionWrapper(const T& t) : _t(t);
      PXR_NS::SdfPath path() const;
      bool isLeadObject(const std::shared_ptr<MhLeadObjectPathTracker>&) const;

      T _t;
};

template<>
bool SelectionWrapper<Fvp::PrimSelection>::isLeadObject(const std::shared_ptr<MhLeadObjectPathTracker>& pt) const
{
    return pt->isLeadObjectSelection(_t);
}

template<>
PXR_NS::SdfPath SelectionWrapper<Fvp::PrimSelection>::path() const
{
    return _t.primPath;
}

template<>
bool SelectionWrapper<PXR_NS::SdfPath>::isLeadObject(const std::shared_ptr<MhLeadObjectPathTracker>& pt) const
{
    return pt->isLeadObjectPrim(_t);
}

template<>
PXR_NS::SdfPath SelectionWrapper<PXR_NS::SdfPath>::path() const
{
    return _t;
}
```
If MhLeadObjectPathTracker simply had overloads for isLeadObject(), with an SdfPath argument or a PrimSelection argument, it would make templating simpler.  The name of the method should not encode the type of the argument, in some sense.

lib/flowViewport/sceneIndex/wireframeHighlights/meshWhSi.cpp
```diff
@@ -0,0 +1,197 @@
...
+#include "meshWhSi.h"
+
+#include <pxr/imaging/hd/instancedBySchema.h>
+#include <pxr/imaging/hd/tokens.h>
+
+PXR_NAMESPACE_USING_DIRECTIVE
+
+namespace {
+
+}
```
ppt-adsk: Nit-pick: empty.

lib/flowViewport/sceneIndex/wireframeHighlights/piInstancerWhSi.h
ppt-adsk: Could a version of this pull request comment become a code comment somewhere in one of the headers?

ppt-adsk (dismissed):
Fantastic work, now the complexity is partitioned off into more manageable chunks!

Only minor comments for documentation.

Feels like an adapted version of the pull request comments could be a
wireframeHighlights/README.md file.

Add requirements section to describe e.g.
- Different selection highlight colors.
- Each selection has its selection highlight sub-hierarchy under a
  highlightHierarchyPrefix, which separates selection highlight prims from
  scene prims.

doc/selectionHighlightingArchitecture.md
```diff
@@ -284,213 +277,3 @@ WireframeSelectionHighlightSceneIndex o-- Selection : Read
   propagate across scene index inputs, so that if a Maya Dag ancestor is
   selected, a USD descendant's appearance can change.  This is the same
   situation as global transformation and visibility.
-
-## Mesh point instancing wireframe selection highlighting
```
debloip-adsk, line 288: Edited and moved to new README.md in wireframeHighlights folder

lib/flowViewport/fvpUtils.h
```diff
@@ -92,6 +92,39 @@ class PrimvarDataSource final : public PXR_NS::HdContainerDataSource
...
+
+template<typename MapValueType>
+auto FindSelfOrFirstChild(const PXR_NS::SdfPath& path, const std::map<PXR_NS::SdfPath, MapValueType>& pathMap) -> decltype(pathMap.cend()) {
+    auto itPath = pathMap.lower_bound(path);
+    if (itPath != pathMap.cend() && itPath->first.HasPrefix(path)) {
+        return itPath;
+    }
+    return pathMap.cend();
+}
+
```
debloip-adsk, line 127: There was an issue where if two paths were in a set/map where one was the parent of another, lookup using the upper_bound + decrement trick could fail (e.g. if a set contained /topPath & /topPath/child1, upper_bound + decrement when called with /topPath/child2 would return /topPath/child1, which is not a prefix of /topPath/child2, even though /topPath is in the set).

I believe for now this was only an issue in BaseWhSi::HasFullySelectedAncestorInclusive (as the other sets/maps all had independent paths), but for safety, I've changed all of the lower_bound/upper_bound usages to find children/parents to use these new utilities, which handle both cases correctly. These can also be re-used elsewhere, e.g. I've re-used it in fvpPruningSceneIndex which had essentially the same bit of code.

lib/flowViewport/sceneIndex/wireframeHighlights/CMakeLists.txt
```diff
@@ -3,23 +3,23 @@
...
-    niPrototypeWhSi.h
-    piInstancerWhSi.h
-    piPrototypeWhSi.h
+    fvpBaseWhSi.h
+    fvpGeomSubsetWhSi.h
+    fvpMeshWhSi.h
+    fvpNiInstanceWhSi.h
+    fvpNiPrototypeWhSi.h
+    fvpPiInstancerWhSi.h
+    fvpPiPrototypeWhSi.h
```
debloip-adsk: Renamed files to conform to the fvp nomenclature

lib/flowViewport/sceneIndex/wireframeHighlights/fvpPiPrototypeWhSi.cpp
```diff
@@ -27,8 +27,18 @@ PXR_NAMESPACE_USING_DIRECTIVE
...
+#if PXR_VERSION >= 2403
+    HdPrimTypeTokens->geomSubset,
+#endif
+};
+
+bool _IsSupportedPointInstancePrototype(const HdSceneIndexBaseRefPtr& sceneIndex, const SdfPath& primPath) {
     HdSceneIndexPrim prim = sceneIndex->GetPrim(primPath);
+    if (kSupportedPrimTypes.find(prim.primType) == kSupportedPrimTypes.cend()) {
+        return false;
+    }
```
debloip-adsk: Adjusts behaviour to match the original documented and intended behavior described at the bottom of the old README.md (now moved to the new README.md). Test snapshots were also updated accordingly.

lib/mayaHydra/hydraExtensions/mhWireframeColorInterfaceImp.cpp
debloip-adsk: I do find the templating a bit overkill at least for now, that being said I did make some slight tweaks to MhWireframeColorInterfaceImp (extracting out the common code of getWireframeColor) and MhLeadObjectPathTracker (renaming the isLeadObjectXYZ methods to just isLeadObject)

lib/flowViewport/sceneIndex/wireframeHighlights/piInstancerWhSi.h
debloip-adsk: Added comments at the top of fvpPiInstancerWhSi.h and fvpPiPrototypeWhSi.h

ppt-adsk (approved):

PR #242, opened 2025-01-28
HYDRA-1286 : Support displacement on GeomSubset wireframe highlights

debloip-adsk:


lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
```diff
@@ -268,45 +298,234 @@ class _RerootingSceneIndexPathArrayDataSource : public HdPathArrayDataSource
     HdPathArrayDataSourceHandle const _inputDataSource;
 };
 
+// Copied over from USD's rerootingSceneIndex.cpp
+class _RerootingSceneIndexContainerDataSource : public HdContainerDataSource
```
debloip-adsk, line 287: This class was previously re-implemented as `RepathingContainerDataSource` so that it could be externally exposed, but now we need no longer need it to be externally accessible due to the new utility method `RepathInstancingDataSources`, so we just go back to using the raw version copied over from USD.

lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
```diff
@@ -268,45 +298,234 @@ class _RerootingSceneIndexPathArrayDataSource : public HdPathArrayDataSource
...
+            {
+                instanceIndicesSchema.GetInstancer()->GetTypedValue(0),
+                instanceIndicesSchema.GetPrototypeIndex()->GetTypedValue(0),
+                std::vector<int>(instanceIndices.begin(), instanceIndices.end())
+            }
+        );
+    }
+
+    return primSelection;
+}
```
debloip-adsk: These did not actually change, just reordered.

lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
```diff
@@ -268,45 +298,234 @@ class _RerootingSceneIndexPathArrayDataSource : public HdPathArrayDataSource
...
+
+PXR_NS::HdContainerDataSourceHandle RepathInstancingDataSources(
```
debloip-adsk, line 387: Replaces `RepathingContainerDataSource` for usage by the wireframe highlight scene indices

test/lib/mayaUsd/render/mayaToHydra/testGeomSubsetsWireframeHighlight.py
debloip-adsk, line 1: Existing tests did not change, just refactored things to allow the displacement test to use a different scene.

debloip-adsk, line 1: Existing tests did not change, just refactored things to allow the displacement test to use a different scene.

lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
lanierd-adsk: I am not sure I understand why you are doing a copy of the usd class and not using it directly from usd  ?

lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
```diff
@@ -337,126 +558,178 @@ _TrimMeshForGeomSubset(const HdContainerDataSourceHandle& meshRootDataSource, co
...
     return dataSourceEditor.Finish();
 }
 #endif
 
-}
-
-namespace FVP_NS_DEF {
-
-TfTokenVector RepathingContainerDataSource::GetNames()
+PXR_NS::HdContainerDataSourceHandle ComputeSmoothNormals(const PXR_NS::HdContainerDataSourceHandle& meshPrimDataSource)
```
lanierd-adsk: I think it would be good to set this in one of the utility class we have so it gets shared by all projects from maya-hydra in case we need it somewhere else. I would also rename the function to addSmoothNormals (I know it's me who named it ComputeSmoothNormals :-) )

lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
```diff
@@ -337,126 +558,178 @@ _TrimMeshForGeomSubset(const HdContainerDataSourceHandle& meshRootDataSource, co
...
-    if (auto childPathDataSource =
-            HdTypedSampledDataSource<SdfPath>::Cast(childSource)) {
-        return _RerootingSceneIndexPathDataSource::New(
-            _srcPrefix, _dstPrefix, childPathDataSource);
+    // Scale normals
+    auto normalsValueDataSource = HdTypedSampledDataSource<VtArray<GfVec3f>>::Cast(HdContainerDataSource::Get(meshPrimDataSource, normalsValueLocator));
+    if (normalsValueDataSource) {
+        auto normals = normalsValueDataSource->GetTypedValue(0);
+        for (size_t iNormal = 0; iNormal < normals.size(); iNormal++) {
+            normals[iNormal] = GfVec3f(xformMatrix.GetInverse().GetTranspose().TransformDir(normals[iNormal]));
```
lanierd-adsk: Could you please compute the transpose of the inverse in a materix out of the loop and applyTransformDir from it ?
I am not sure this gets optimized correctly by the different compilers we use when being inside the loop.

lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
```diff
@@ -337,126 +558,178 @@ _TrimMeshForGeomSubset(const HdContainerDataSourceHandle& meshRootDataSource, co
...
+    // Set xform to the identity to prevent scaling from being re-applied
+    dataSourceEditor.Set(HdXformSchema::GetDefaultLocator().Append(HdXformSchemaTokens->matrix), HdRetainedTypedSampledDataSource<GfMatrix4d>::New(GfMatrix4d(1)));
 
-    return childSource;
+    return dataSourceEditor.Finish();
 }
 
-//We want to set the displayStyle of the selected prim to refinedWireOnSurf only if the displayStyle of the prim is refined (meaning shaded)
-HdContainerDataSourceHandle SetWireframeRepr(const HdContainerDataSourceHandle& dataSource, const GfVec4f& color)
+PXR_NS::HdContainerDataSourceHandle ForceDisplacement(const PXR_NS::HdContainerDataSourceHandle& meshPrimDataSource, float displacement)
```
lanierd-adsk: I would add this to a utility class as well and call it ApplyDisplacement instead of ForceDisplacement

lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
```diff
@@ -337,126 +558,178 @@ _TrimMeshForGeomSubset(const HdContainerDataSourceHandle& meshRootDataSource, co
...
-        HdInstanceIndicesSchema instanceIndicesSchema = nestedInstanceIndicesSchema.GetElement(iNestedInstanceIndices);
-        auto instanceIndices = instanceIndicesSchema.GetInstanceIndices()->GetTypedValue(0);
-        primSelection.nestedInstanceIndices.push_back(
-            {
-                instanceIndicesSchema.GetInstancer()->GetTypedValue(0),
-                instanceIndicesSchema.GetPrototypeIndex()->GetTypedValue(0),
-                std::vector<int>(instanceIndices.begin(), instanceIndices.end())
-            }
-        );
+PXR_NS::HdContainerDataSourceHandle AddDependency(
```
lanierd-adsk: It could also be part of the utility functions

lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
```diff
@@ -756,4 +1029,81 @@ BaseWhSi::CollectInstancingPaths(const PXR_NS::SdfPath& primPath, InstancingPath
     }
 }
 
+VtValue 
+BaseWhSi::GetMaterialDisplacementValue(const PXR_NS::HdContainerDataSourceHandle& primDataSource) const
```
lanierd-adsk: As well, I think it should be part of a utility function

lanierd-adsk (changes requested):
Thanks for adding this. A few comments overall.

lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
debloip-adsk: Those classes are not exposed, they are located in an anonymous namespace in a .cpp file (see USD's [rerootingSceneIndex.cpp](https://github.com/PixarAnimationStudios/OpenUSD/blob/ba8aaf1f9e3d72822ce9c9041a6dd75e4eb53906/pxr/usdImaging/usdImaging/rerootingSceneIndex.cpp), so we can't use them directly. May be worth to make a PR to USD to expose these, I've logged HYDRA-1422 so we can discuss this.

lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
debloip-adsk: Agree, moved it to fvpUtils and updated the name

lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
debloip-adsk: Good point, moved it out of the loop, thanks

lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
debloip-adsk: For both this one and ForceScale, I went to with "Force" to really emphasize that this is not something that should be naturally occurring, and is more of a workaround more than anything. This is also why I don't think we should move these over to common utilities, as we should be using this as little as possible. I've left both of these where they are for now, let me know if you have other thoughts on this

lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
debloip-adsk: Agree, moved it to fvpUtils

lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
debloip-adsk: Moved it (and GetMaterialPath) to fvpUtils

debloip-adsk:
I also moved ConvertHydraToFvpSelection to fvpUtils

lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
lanierd-adsk: Ok, I've got your point, let's keep it like that.

lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
```diff
@@ -756,4 +887,38 @@ BaseWhSi::CollectInstancingPaths(const PXR_NS::SdfPath& primPath, InstancingPath
...
 
+#if PXR_VERSION >= 2403
+PXR_NS::HdContainerDataSourceHandle
+BaseWhSi::MakeGeomSubsetHighlight(
+    const PXR_NS::HdContainerDataSourceHandle& meshPrimDataSource,
+    const PXR_NS::HdContainerDataSourceHandle& geomSubsetPrimDataSource) const
+{
+    HdContainerDataSourceHandle editedMeshPrimDataSource = meshPrimDataSource;
+
+    VtValue displacementValue = GetMaterialDisplacementValue(geomSubsetPrimDataSource, *GetInputSceneIndex());
```
lanierd-adsk: I recall there were 2 possible cases for the scale when the parent mesh of the geomsubset has a material and the geomsubset has a material as well :
- the scale can be on the parent mesh 's material
- the scale can be on the geomsubset material
Are these 2 cases handled in your code ? It seems to me, I may be wrong, that if the scale is on the parent mesh 's material, we'll miss it.
And I am not sure you can have scale on both materials, I think in this case we apply only one if I am correct ?

lanierd-adsk (commented):
Thanks for the reworking. Only 1 remaining question and we're done.

lanierd-adsk (approved):

PR #243, opened 2025-02-03
Tremblp/hydra 1413/isolate select keep lighting

ppt-adsk:


test/lib/mayaUsd/render/mayaToHydra/testIsolateSelectWithUsdLighting.py
ppt-adsk, line 1: As desired, the USD light is excluded from the isolate select set, but still contributes its lighting to the scene.

lanierd-adsk (dismissed):

lanierd-adsk (approved):

PR #249, opened 2025-02-13
HYDRA-1175 : Use LightsManagement scene index instead of removing the…

lanierd-adsk:
… light prims for maya native data, update the unit tests


lib/flowViewport/sceneIndex/fvpLightsManagementSceneIndex.cpp
```diff
@@ -91,60 +156,54 @@ bool LightsManagementSceneIndex::_IsDefaultLight(const SdfPath& primPath)const
...
+    }
+
+    return prim;
+}
+
+void LightsManagementSceneIndex::SetDisabledLightsPrims(
+    const std::set<PXR_NS::SdfPath>& activeLightsPrims)
+{
+    if (_disabledLightsPrims != activeLightsPrims) {
+        _disabledLightsPrims = activeLightsPrims;
```
debloip-adsk: Bit confused on the nomenclature here, shouldn't the parameter name be disabledLightPrims and not activeLightsPrims? The prims passed here are the non-active ones, correct?

test/testSamples/testMayaLightingModes/AllKindsOfLights.ma
```diff
@@ -151,7 +196,7 @@ createNode aiImagerDenoiserOidn -s -n "defaultArnoldDenoiser";
 	rename -uid "F3377A5A-4A07-7C62-07C5-D68E104F2CB2";
 createNode file -n "file1";
 	rename -uid "6E5F111E-40E1-79BC-D1C7-2588F65A88E1";
-	setAttr ".ftn" -type "string" "./UVChecker.png";
+	setAttr ".ftn" -type "string" "D:/GIT/maya-hydra-opensource/test/testSamples/testMayaLightingModes/UVChecker.png";
```
debloip-adsk, line 199: Not sure if it's going to cause issues, but can this path be made relative?

test/lib/mayaUsd/render/mayaToHydra/testMayaLights.py
```diff
@@ -33,8 +33,11 @@ def IMAGE_DIFF_FAIL_PERCENT(self):
         # Only the "all lights with shadows" check fails, and shadows with point light only is skipped.
         if platform.system() == "Darwin":
             return 3
+        if mayaUtils.mayaMajorVersion() == 2025:
+            return 2 #There are more pixels different in maya 2025 for this test
```
debloip-adsk, line 37: Just making sure, but I'm guessing the difference is small enough that it doesn't warrant separate reference images?

debloip-adsk:
Few questions, overall looking good, thanks!

lib/flowViewport/sceneIndex/fvpLightsManagementSceneIndex.cpp
lanierd-adsk: Good catch, I started by naming this function SetActiveLightsPrims, this is a leftover from that naming, I'll fix it.

test/lib/mayaUsd/render/mayaToHydra/testMayaLights.py
lanierd-adsk: No I had an error of about 2440 pixels (~1.5%).
I did an image compare to see the differences and it was sparse pixels here and there, nothing important and only in maya 2025 under Windows and Linux...

lib/flowViewport/sceneIndex/fvpLightsManagementSceneIndex.h
```diff
@@ -71,9 +71,13 @@ class LightsManagementSceneIndex : public PXR_NS::HdSingleInputFilteringSceneInd
     FVP_API
     LightingMode GetLightingMode()const {return _lightingMode;}
     
-protected:
+    FVP_API
+    void SetDisabledLightsPrims(const std::set<PXR_NS::SdfPath>& activeLightsPrims);
```
debloip-adsk: The parameter should also be renamed here

debloip-adsk (dismissed):

lanierd-adsk:
@debloip-adsk Can you please do the code review again ? Thanks.

debloip-adsk (approved):

PR #255, opened 2025-02-24
Image file writing, with test.

ppt-adsk:
Added the capability to write a Hydra-generated image to a file, in the Flow Viewport Toolkit library.


lib/flowViewport/imageWriter/fvpImageBufferWriterFactory.cpp
```diff
@@ -0,0 +1,39 @@
...
+namespace FVP_NS_DEF {
+
+/* static */ 
+ImageBufferWriter::Ptr ImageBufferWriter::Create(
+    const PXR_NS::VtDictionary& args,
+    const PXR_NS::TfToken&      aov
+)
+{
+    // Can't use ternary operator because of differing std::shared_ptr types.
+    if (args.find("hgi") == args.end()) {
```
lanierd-adsk, line 31: I think it would be worth explaining more why you're looking for hgi andthe differences it will do unless it is explained somewhere else ?

lanierd-adsk (commented):
Thanks for doing all this. I think it would be worth explaining in the 2 classes about image and texture buffer when one r the other has to be used.

lanierd-adsk (approved):

PR #257, opened 2025-02-27
Added USD inherits / specializes test.

ppt-adsk:


test/lib/mayaUsd/render/mayaToHydra/testUsdInheritsSpecializes.py
```diff
@@ -0,0 +1,65 @@
...
+    _requiredPlugins = ['mayaHydraCppTests']
+
+    IMAGE_DIFF_FAIL_THRESHOLD = 0.1
+    IMAGE_DIFF_FAIL_PERCENT = 2
+
+    @property
+    def imageFileName(self):
+        # Same image snapshot for both inherits and specializes test.
+        # Fallback material change in USD 24.11, Maya 2025 uses USD 23.11.
+        return 'inherits_2025.png' if mayaUtils.mayaMajorVersion() == 2025 \
```
ppt-adsk: @lanierd-adsk is this the right thing to do?  Or is there a way to have the same reference image for both 2025 and for later versions?

lanierd-adsk: There is another way where you keep the same filename but use an image version parameter (as a string) as a parameter of the assertSnapshotClose. This reads the filename in a subfolder with the imageVersion parameter name. Please see :
https://github.com/Autodesk/maya-hydra/blob/74678d6be7c169674e7bfad48382d6b9b65ada6f/test/lib/mayaUsd/render/mayaToHydra/testBasicRender.py#L70
And we have subfolder of this test named "lambert" and "standardSurface" which correspond to the image version parameter name used.

lanierd-adsk: In this test, we have a folder for usd 24.11+ images using that imageVersion parameter :
https://github.com/Autodesk/maya-hydra/blob/74678d6be7c169674e7bfad48382d6b9b65ada6f/test/lib/mayaUsd/render/mayaToHydra/testFootPrintNode.py#L45

lanierd-adsk (commented):
Thanks for doing this, I think it would be better to use a subfolder with the imageVerdsion parameter as I described in the comments.

test/lib/mayaUsd/render/mayaToHydra/testUsdInheritsSpecializes.py
ppt-adsk: Used `modifyDefaultLightIntensityByUsdVersion()`, which unifies the default lighting.

test/lib/mayaUsd/render/mayaToHydra/testUsdInheritsSpecializes.py
```diff
@@ -0,0 +1,66 @@
...
+    IMAGE_DIFF_FAIL_PERCENT = 2
+
+    @property
+    def imageFileName(self):
+        # Same image snapshot for both inherits and specializes test.
+        return 'inherits.png'
+
+    def setUp(self):
+        super(TestUsdInheritsSpecializes, self).setUp()
+        self.modifyDefaultLightIntensityByUsdVersion()
```
lanierd-adsk, line 38: Sorry, I didn't get you were having this kind of problems... I did that function :-)

lanierd-adsk (approved):

PR #266, opened 2025-04-18
HYDRA-1552 : Bifrost USD attribute changes don't refresh in the Hydra…

lanierd-adsk:
… Viewport. To fix it we check when rendeing a frame if some updates need to be done in the MayaUsd proxy shape nodes


ppt-adsk (commented):
Looks good, but is there any way for us to test this automatically?

ppt-adsk (dismissed):

ppt-adsk (approved):

PR #271, opened 2025-04-30
Do not register pick handler twice.

ppt-adsk:
I was unable to write an automated test for this.  In the automated test, the renderer switch on the second viewport does not occur, even if I force refresh, so the duplicate pick handler registration does not happen.

Manually tested the fix interactively.

### DIFF
--- a/lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
+++ b/lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
@@ -466,9 +466,13 @@ MayaHydraSceneIndex::MayaHydraSceneIndex(
         _fallbackMaterial = SdfPath::EmptyPath(); // Empty path for hydra fallback material
     });
 
-    // Add our pick handler to the pick handler registry.
-    auto pickHandler = std::make_shared<MayaPickHandler>(*this);
-    TF_AXIOM(MayaHydra::PickHandlerRegistry::Instance().Register(_rprimPath, pickHandler));
+    // Add our pick handler to the pick handler registry if there is none.
+    auto& phr = MayaHydra::PickHandlerRegistry::Instance();
+    _unregisterPickHandler = (phr.RegisteredHandler(_rprimPath) == nullptr);
+    if (_unregisterPickHandler) {
+      auto pickHandler = std::make_shared<MayaPickHandler>(*this);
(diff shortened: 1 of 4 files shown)


lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -466,9 +466,13 @@ MayaHydraSceneIndex::MayaHydraSceneIndex(
         _fallbackMaterial = SdfPath::EmptyPath(); // Empty path for hydra fallback material
     });
 
-    // Add our pick handler to the pick handler registry.
-    auto pickHandler = std::make_shared<MayaPickHandler>(*this);
-    TF_AXIOM(MayaHydra::PickHandlerRegistry::Instance().Register(_rprimPath, pickHandler));
+    // Add our pick handler to the pick handler registry if there is none.
+    auto& phr = MayaHydra::PickHandlerRegistry::Instance();
+    _unregisterPickHandler = (phr.RegisteredHandler(_rprimPath) == nullptr);
+    if (_unregisterPickHandler) {
```
ppt-adsk, line 472: If we switch renderers on a viewport in multi-viewport mode, the scene index tree is re-built, and we try to register a pick handler again, with the same _rprimPath.  Make sure we register a pick handler only once.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -525,7 +529,9 @@ void MayaHydraSceneIndex::_Destroy()
     Fvp::PathMapperRegistry::Instance().SetFallbackMapper(nullptr);
 
     // Remove our pick handler from the pick handler registry.
-    TF_AXIOM(MayaHydra::PickHandlerRegistry::Instance().Unregister(_rprimPath));
+    if (_unregisterPickHandler) {
+      TF_AXIOM(MayaHydra::PickHandlerRegistry::Instance().Unregister(_rprimPath));
```
ppt-adsk, line 533: If this object registered the pick handler, it unregisters it.

debloip-adsk (approved):

ppt-adsk:
I believe the pre-flight failures are for the macOS licensing issue and the Linux shared library load issue, nothing to do with my pull request, so this fix can be merged.

PR #273, opened 2025-05-06
HYDRA-1497 : Fix geomsubset sel . mode being very slow

lanierd-adsk:


lib/mayaHydra/hydraExtensions/pick/mhUsdPickHandler.cpp
```diff
@@ -236,7 +236,7 @@ std::vector<UsdPickHandler::HitPath> resolveGeomSubsetsPicking(
     const TfToken& geomSubsetType, 
     int componentIndex)
 {
-    if (componentIndex < 0 || sceneIndex->GetPrim(basePrimPath).primType != HdPrimTypeTokens->mesh) {
```
lanierd-adsk, line 239: I am creating a JIRa to add a unit test later.

lanierd-adsk: We actually check the prim type at line 247 in this function so checking it here was more to save time but it produced a slowdown.

debloip-adsk (approved):

PR #277, opened 2025-05-15
HYDRA-1551 - Fix selection highlights on USD objects when enabling wireframe

cossetr-adsk:
Add a new notification to dirty the `overrideWireframeColor` of objects that are selected, but that are not the Lead selection.


lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -877,6 +877,7 @@ MStatus MtohRenderOverride::Render(
     // Hydra supports Wireframe and WireframeOnSurfaceRefined repr for wireframe on shaded mode.
     // Refinement level for Hydra is set in Hydra Render Globals    
     const MFrameContext::WireOnShadedMode wireOnShadedMode = MFrameContext::wireOnShadedMode();//Get the user preference
+    _meshWhSi->SetWireframeOnShadedEnabled(wireOnShadedMode != MFrameContext::WireOnShadedMode::kWireFrameOnShadedNone);
```
lilike-adsk: It's not only WireframeOnShaded, also Wireframe mode, I'd rename this flag as  _nonShadedEnabled.

ppt-adsk:
Needs an automated test.

lib/flowViewport/sceneIndex/wireframeHighlights/fvpMeshWhSi.cpp
```diff
@@ -22,6 +22,18 @@ PXR_NAMESPACE_USING_DIRECTIVE
 
 namespace FVP_NS_DEF {
 
+// Copied from fvpBaseWhSi.cpp
+namespace {
+    TF_DEFINE_PRIVATE_TOKENS(
```
ppt-adsk: Should this be TF_DEFINE_PUBLIC_TOKENS then and avoid the copy?

lib/flowViewport/sceneIndex/wireframeHighlights/fvpMeshWhSi.cpp
```diff
@@ -22,6 +22,18 @@ PXR_NAMESPACE_USING_DIRECTIVE
...
+        _tokens,
+        // Handle primsvars:overrideWireframeColor in Storm for wireframe selection highlighting color
+        (overrideWireframeColor)    // Works in HdStorm to override the wireframe color
```
ppt-adsk: Is this really Storm-specific?  Feels quite unfortunate to have renderer-specific code in such a basic class.

ppt-adsk (commented):
Great work!

lib/flowViewport/sceneIndex/wireframeHighlights/fvpMeshWhSi.cpp
lilike-adsk: Yes, we used overrideWireframeColor primvar (Storm specific) for wire color, I'm not sure why we didn't use wireframeColor primvar (Hydra specific) directly, we can log a ticket for this.

lanierd-adsk: This "overrideWireframeColor" variable is used in a Storm shader, it's not supposed to work in other renderers.
IIRC, I tried the wireframeColor primvar but we can give it another try, it was long ago.

lib/flowViewport/sceneIndex/wireframeHighlights/fvpMeshWhSi.cpp
cossetr-adsk: This token is already defined privately in three places:
- https://github.com/Autodesk/maya-hydra/blob/2eee6ff2b4f1f3eb76f2c6629b05caf77a2b48b9/lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp#L52
- https://github.com/Autodesk/maya-hydra/blob/2eee6ff2b4f1f3eb76f2c6629b05caf77a2b48b9/lib/mayaHydra/hydraExtensions/sceneIndex/mhDirtyLeadObjectSceneIndex.cpp#L39
- https://github.com/Autodesk/maya-hydra/blob/2eee6ff2b4f1f3eb76f2c6629b05caf77a2b48b9/lib/flowViewport/sceneIndex/fvpReprSelectorSceneIndex.cpp#L41

I think it would be good to refactor it and define it in a single place, but it wasn't obvious to me where that should be.

ppt-adsk: lib/flowViewport/tokens.h?  Not very specific, though.

lib/flowViewport/sceneIndex/wireframeHighlights/fvpMeshWhSi.cpp
likeli007: @cossetr-adsk could you try to replace all `overrideWireframeColor `with `wireframeColor ` in our code base and see if the highlighting color is still working?

lilike-adsk: oops, the above is me, wrong account

test/testUtils/imageUtils.py
```diff
@@ -211,8 +212,11 @@ def assertImagesEqual(self, imagePath1, imagePath2):
         self.assertImagesClose(imagePath1, imagePath2, fail=None, failpercent=None)
     
     def assertSnapshotClose(self, refImagePath, fail, failpercent, hardfail=None, 
-                warn=None, warnpercent=None, hardwarn=None, perceptual=False):
-        snapImagePath = os.path.join(self.getSnapshotDir(), os.path.basename(refImagePath))
+                warn=None, warnpercent=None, hardwarn=None, perceptual=False, *, imageVersion=None):
+        if imageVersion is not None:
+            snapImagePath = os.path.join(self.getSnapshotDir(), imageVersion, os.path.basename(refImagePath))
```
cossetr-adsk, line 217: Before this fix, snapshots from different versions would overwrite each other. The test would run correctly, but when looking in the output folder, only 1 set of snapshots is visible instead of 3.
With this fix, snapshots for each version are stored in a separate subfolder.

test/lib/mayaUsd/render/mayaToHydra/testDataProducerSelHighlight.py
```diff
@@ -121,39 +121,55 @@ def test_UsdKeepSelectionHighlighting(self):
...
         #Remove a non lead object
         ufeGlobalSel.remove(self.pSphere1UfeItem)
-        self.assertSnapshotClose("Storm_S2AndTSelPAndS1Removed.png", self.IMAGE_DIFF_FAIL_THRESHOLD, self.IMAGE_DIFF_FAIL_PERCENT)
+        assertSnapshotCloseImpl("Storm_6_S1Removed.png")
+
+        # De-select everything
+        ufeGlobalSel.clear()
+        assertSnapshotCloseImpl("Storm_0_NoSelected.png")
+
+    def test_LeadAndActiveColorsSelectionHighlighting(self):
```
cossetr-adsk, line 153: I decided to run a separate selection test for smoothShaded, wireframe, and wireframeOnShaded. The transition between display modes is already tested in test_UsdKeepSelectionHighlighting

test/lib/mayaUsd/render/mayaToHydra/testDataProducerSelHighlight.py
```diff
@@ -121,39 +121,55 @@ def test_UsdKeepSelectionHighlighting(self):
...
 
         #Remove the lead object
         ufeGlobalSel.remove(self.pPlaneUfeItem)
-        self.assertSnapshotClose("Storm_S1And2AndTSelectedPRemoved.png", self.IMAGE_DIFF_FAIL_THRESHOLD, self.IMAGE_DIFF_FAIL_PERCENT)
+        assertSnapshotCloseImpl("Storm_5_PRemoved.png")
 
         #Remove a non lead object
         ufeGlobalSel.remove(self.pSphere1UfeItem)
-        self.assertSnapshotClose("Storm_S2AndTSelPAndS1Removed.png", self.IMAGE_DIFF_FAIL_THRESHOLD, self.IMAGE_DIFF_FAIL_PERCENT)
+        assertSnapshotCloseImpl("Storm_6_S1Removed.png")
```
cossetr-adsk, line 147: Without the fix in `lib/flowViewport/sceneIndex/wireframeHighlights/fvpMeshWhSi.cpp`, the test fails here in wireframe and wireframeOnShaded because the viewport is not updated correctly when de-selecting a non lead object.

lib/flowViewport/sceneIndex/wireframeHighlights/fvpMeshWhSi.cpp
cossetr-adsk: I put something in `lib/flowViewport/tokens.h`. I think it is not great, but it seems like a step in the right direction to avoid code duplication. Let me know what you think.

lib/flowViewport/sceneIndex/wireframeHighlights/fvpMeshWhSi.cpp
cossetr-adsk: I tried using the existing `HdShaderTokens->wireframeColor` instead of the current `_tokens->overrideWireframeColor`, but I get a bunch of shader errors and USD objects are not rendered: `error C1013: function "HdGet_wireframeColor" is already defined at 0(424)`

lilike-adsk: Thanks for checking, let's use `overrideWireframeColor `for now

lilike-adsk (approved):

test/testUtils/imageUtils.py
```diff
@@ -211,8 +212,11 @@ def assertImagesEqual(self, imagePath1, imagePath2):
         self.assertImagesClose(imagePath1, imagePath2, fail=None, failpercent=None)
     
     def assertSnapshotClose(self, refImagePath, fail, failpercent, hardfail=None, 
-                warn=None, warnpercent=None, hardwarn=None, perceptual=False):
-        snapImagePath = os.path.join(self.getSnapshotDir(), os.path.basename(refImagePath))
+                warn=None, warnpercent=None, hardwarn=None, perceptual=False, *, imageVersion=None):
```
cossetr-adsk, line 215: Add `, *,` to force using named parameters for what comes after the *. This way, the new `imageVersion` parameter can't be used accidentally

lib/flowViewport/sceneIndex/wireframeHighlights/fvpMeshWhSi.cpp
ppt-adsk: Yes, I agree, not specific enough, but a step in the right direction, good enough for now.

ppt-adsk (approved):
LGTM!

PR #279, opened 2025-05-20
HYDRA-1578 : Use default bounding box if no extents are specified

debloip-adsk:


test/lib/mayaUsd/render/mayaToHydra/testBoundingBox.py
```diff
@@ -0,0 +1,53 @@
+# Copyright 2024 Autodesk
```
lanierd-adsk: Please change this to 2025 as this file is new. Thanks.

lanierd-adsk (approved):

PR #284, opened 2025-06-05
HYDRA-1664 : Prevent crash in native instance highlighting

debloip-adsk:


lib/flowViewport/sceneIndex/wireframeHighlights/fvpNiInstanceWhSi.cpp
```diff
@@ -55,6 +55,11 @@ HdSceneIndexPrim NiInstanceWhSi::GetHighlightPrim(const SdfPath &selectionPath,
 
     auto originalPath = fullPrimPath.ReplacePrefix(selectionPath, _selectionPathsToPrototypePrefixes.at(selectionPath));
     HdSceneIndexPrim prim = GetInputSceneIndex()->GetPrim(originalPath);
+    if (!prim.dataSource) {
+        // If there is no data source, return. Trying to get child
+        // data sources or schemas from a null data source will crash.
+        return prim;
+    }
```
debloip-adsk, line 62: In one of our test scenes, there could be an empty prim in a native instance hierarchy, which lead to a crash. I tried to see if I could write a unit test for this, but I'm not sure it's worth the effort. I tried to see what kind of setup in the test scene was leading to the empty prim in the prototype hierarchy, but the scene setup's is quite complicated, so it's difficult to tell what exactly was leading to this situation. Writing a custom data producer just for a unit test sounds like a bit too much as well, so I think just merging the fix alone and testing it manually is a better use of our time

lilike-adsk (approved):

PR #294, opened 2025-07-09
HYDRA-1626: Hydra Primitive Scene Stats 

huangi-adsk:
* Adding a scene stat data dump function to be called by the performance automation framework that works for non-Mesh geometry types in Maya surface, USD or custom meshes
* Unit tests added for empty scene and scene with basic prims

### DIFF
--- a/lib/mayaHydra/mayaPlugin/renderOverride.cpp
+++ b/lib/mayaHydra/mayaPlugin/renderOverride.cpp
@@ -86,15 +86,12 @@
 #include <pxr/imaging/hd/rprim.h>
 #include <pxr/imaging/hd/sceneIndexPluginRegistry.h>
 #include <pxr/imaging/hd/dataSource.h>
-#include <pxr/imaging/hd/basisCurvesTopology.h>
-#include <pxr/imaging/hd/meshTopology.h>
 #include <pxr/imaging/hd/meshSchema.h>
 #include <pxr/imaging/hd/basisCurvesSchema.h>
 #include <pxr/imaging/hd/sceneIndexPrimView.h>
 #include <pxr/imaging/hd/mesh.h>
 #include <pxr/imaging/hd/basisCurves.h>
 #include <pxr/imaging/hd/points.h>
-#include <pxr/imaging/hd/tokens.h>
 #include <pxr/imaging/hdx/selectionTask.h>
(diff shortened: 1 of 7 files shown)


lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -440,6 +444,77 @@ int MtohRenderOverride::GetUsedGPUMemory()
     return totalGPUMemory / (1024*1024); 
 }
 
+std::map<std::string, int> MtohRenderOverride::GetSceneStatistics()
+{
+    std::map<std::string, int> stats = {
```
lilike-adsk, line 452: Nice, I'm thinking about the interface, I'd suggest to use a data structure (class) to hold the hydra scene stats. 
Like below, and we can have a function to output the HydraSceneStat to your map<string, int> , the benefit is we can extend the hydra scene stats easily.

HydraSceneStat
|--Prims
|----RPrim
|------Mesh
|--------Faces
|--------Points
|----SPrim
|------Camera
|------Light
|------Material
|----BPrim
|------Textures
...

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -440,6 +444,77 @@ int MtohRenderOverride::GetUsedGPUMemory()
...
+        auto* rprim = renderIndex->GetRprim(primId);
+        if (!rprim) {
+            continue;
+        }
+
+        stats["primitives"]++;
+
+        HdSceneIndexPrim prim;
+        TfToken primType;
+        if (instance->_mayaHydraSceneIndex) {
```
lilike-adsk: Need to find a common way to query the primtype, it could come from any SceneIndex

lilike-adsk: One way might be check the type of auto* rprim = renderIndex->GetRprim(primId);, if it could be dynamically casted to HdMesh, we will assume it's a mesh prim.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
huangi-adsk: After discussion yesterday, simpler approach taken due to resource management

lilike-adsk (approved):

PR #296, opened 2025-07-21
HYDRA-1438: Supporting Transform Selection Highlighting on Native Instances

huangi-adsk:
* Added transform primvar dirty detection to the wireframe highlighting rebuild logic to allow wireframe highlights to follow transformed native USD instances with field `instanceable = true` (scale, translation, rotation).


ppt-adsk (changes requested):
Looks good, but needs a test.

ppt-adsk (dismissed):
Nice!

ppt-adsk (approved):

PR #300, opened 2025-07-23
Change erroneous USD version code switch from 2403 to 2405

jjeziorski:
Code switch for geom subsets is incorrect it should be 24.05+ as described here: https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/CHANGELOG.md?plain=1#L3001

One of our variants uses USD 24.03 and it cannot be built with current state.


lilike-adsk:
Good catch, thanks for fixing this

lilike-adsk (dismissed):

lilike-adsk (approved):

PR #302, opened 2025-07-25
Fix camera parameter translations and simplify code

jjeziorski:
Fix the translation of focus distance and simplify the handling of camera lens and aperture parameters. Enhance the retrieval of camera attributes by searching for requested parameters directly.

Value for focus distance was incorrectly assumed in inches, which is not consistent with [docs](https://help.autodesk.com/view/MAYAUL/2024/ENU/?guid=GUID-C3EBB008-7DBE-4B9D-B9AC-1DA974CEFE15)

I'm not sure where logic for computing lens focal and aperture came from, but it was expensive, 
 units were way off and gave wrong values for depth of field calculations. With proposed changes values are consistent with values exported by `maya-usd`.

I've also added proposed change to support custom attributes by checking camera plugs.

### DIFF
--- a/lib/mayaHydra/hydraExtensions/adapters/cameraAdapter.cpp
+++ b/lib/mayaHydra/hydraExtensions/adapters/cameraAdapter.cpp
@@ -130,7 +130,6 @@ VtValue MayaHydraCameraAdapter::Get(const TfToken& key) { return MayaHydraShapeA
 
 VtValue MayaHydraCameraAdapter::GetCameraParamValue(const TfToken& paramName)
 {
-    constexpr double mayaInchToHydraCentimeter = 0.254;
     constexpr double mayaInchToHydraMillimeter = 0.0254;
     constexpr double mayaFocaLenToHydra = 0.01;
 
@@ -220,7 +219,7 @@ VtValue MayaHydraCameraAdapter::GetCameraParamValue(const TfToken& paramName)
         auto focusDistance = camera.focusDistance(&status);
         if (hadError(status))
             return {};
-        return VtValue(float(focusDistance * mayaInchToHydraCentimeter));
+        return VtValue(float(focusDistance));
     }
     if (paramName == HdCameraTokens->focalLength) {
         const double aspectRatio = _viewport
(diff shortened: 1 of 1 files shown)


lilike-adsk:
@jjeziorski have you had a chance to try the change with custom attributes support?

jjeziorski:
@lilike-adsk Yes, I had a look at custom attributes support and I've reverted suggested changes in this PR.
It would be great if remaining changes could be reviewed. Thank you

lilike-adsk:
> @lilike-adsk Yes, I had a look at custom attributes support and I've reverted suggested changes in this PR. It would be great if remaining changes could be reviewed. Thank you

Sure, will take a look your left changes

lilike-adsk:
starting a preflight run

lib/mayaHydra/hydraExtensions/adapters/cameraAdapter.cpp
```diff
@@ -220,23 +199,13 @@ VtValue MayaHydraCameraAdapter::GetCameraParamValue(const TfToken& paramName)
         auto focusDistance = camera.focusDistance(&status);
         if (hadError(status))
             return {};
-        return VtValue(float(focusDistance * mayaInchToHydraCentimeter));
+        return VtValue(float(focusDistance));
```
lilike-adsk, line 202: Good catch

lib/mayaHydra/hydraExtensions/adapters/cameraAdapter.cpp
```diff
@@ -220,23 +199,13 @@ VtValue MayaHydraCameraAdapter::GetCameraParamValue(const TfToken& paramName)
...
-        status = camera.getViewingFrustum(aspectRatio, left, right, bottom, top, true, false, true);
-
-        const double cameraNear = camera.nearClippingPlane();
-
-        const double focalLen
-            = (convertFit(camera) == CameraUtilConformWindowPolicy::CameraUtilMatchVertically)
-            ? (2.0 * cameraNear) / (top - bottom)
-            : (2.0 * cameraNear) / (right - left);
-        return VtValue(float(focalLen * mayaFocaLenToHydra));
+        auto focalLength = camera.focalLength(&status);
```
lilike-adsk, line 205: I'll check why the parameters were depending on the viewport size before

jjeziorski: The bottom line for us is that scene exported to USD and rendered directly with `mayaHydra` should render the same regardless of viewport size. With current code it would only match sometimes.

lilike-adsk (approved):

PR #309, opened 2025-09-08
HYDRA-709 : Extend transform tests

debloip-adsk:
Add transform tests for cameras, lights, and mesh adapter.


test/lib/mayaUsd/render/mayaToHydra/CMakeLists.txt
```diff
@@ -130,6 +130,7 @@ endif()
 # both versions of the test in parallel will conflict.
 set(INTERACTIVE_TEST_SCRIPT_FILES_MESH_ADAPTER
     testMeshes.py
+    testTransformMeshAdapter.py
     cpp/testMeshAdapterTransform.py
```
debloip-adsk, line 134: Difference between the two is that the cpp test checks notifications, but not the visual result

lilike-adsk (dismissed):

lilike-adsk (approved):

PR #312, opened 2025-09-12
HYDRA-1203 : Update minimum CMake version

debloip-adsk:
This PR bumps the minimum CMake version to 3.21, as it is the first CMake version to support VS2022, which is our requirement on Windows.

There is also a slight improvement to an existing CMake snippet.


CMakeLists.txt
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-cmake_minimum_required(VERSION 3.13...3.17)
+cmake_minimum_required(VERSION 3.21)
```
lilike-adsk: also update the build.md to keep consistent?

debloip-adsk: Ah yes good catch. Updated

lilike-adsk (dismissed):
Nice cleanup

lilike-adsk (approved):

PR #320, opened 2025-09-22
HYDRA-1682 : Integrate Hydra Viewport Toolbox in MayaHydra

debloip-adsk:
This PR adds support for Autodesk's [Hydra Viewport Toolbox](https://github.com/Autodesk/hydra-viewport-toolbox) inside MayaHydra.

With this change come the following features : 
- Ability to use selection highlighting with any render delegate
- Ability to query and display specific AOVs in the viewport
- Ability to define a crop render region for more efficient rendering

Note that this is a first iteration and is still WIP.


debloip-adsk:
A teeny tiny PR to start the week :)

lib/flowViewport/fvpUtils.h
```diff
@@ -26,6 +26,7 @@
 #include <pxr/imaging/hd/retainedDataSource.h>
 #include <pxr/imaging/hd/selectionSchema.h>
 #include <pxr/imaging/hd/primvarsSchema.h>
+#include <pxr/imaging/hd/sceneIndex.h>
```
lanierd-adsk, line 29: I don't think this is needed, it seems to be a leftover ?

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraMayaDataProducerSceneIndexData.cpp
```diff
@@ -61,9 +61,6 @@ MayaDataProducerSceneIndexData::MayaDataProducerSceneIndexData(FVP_NS_DEF::DataP
     if (_dccNode) {
         SetupDCCNode();
     }
-    //Reset the prefix member which was used only for SetupDCCNode, as for usd stages we add a prefixing scene index outside of flow viewport which holds the prefix.
-    //This is a way to have a common call to MAYAHYDRA_NS::MhDataProducersMayaNodeToSdfPathRegistry::Get().Add in this class.
-    _prefix = SdfPath::AbsoluteRootPath();
```
lanierd-adsk, line 66: We remove this, but is it still used ?

lilike-adsk (approved):
Take a glance, LGTM.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1201,54 +1675,62 @@ void MtohRenderOverride::_InitHydraResources(const MHWRender::MDrawContext& draw
...
         TfToken("MayaHydraSceneIndex"),
-        _engine,
-        *_renderIndex,
+        *renderIndex(),
         MAYA_NATIVE_ROOT,
         _isUsingHdSt
     );
 
-    // Render index proxy sets up the Flow Viewport merging scene index, must
+    // Data producer mering scene index sets up the Flow Viewport merging scene index, must
```
lanierd-adsk, line 1697: There is a typo on "merging"

test/lib/mayaUsd/render/mayaToHydra/testMayaShadingModes.py
```diff
@@ -38,6 +38,12 @@ def test_MayaShadingModes(self):
 
         panel = mayaUtils.activeModelPanel()        
         
+        #Deal with 2 render passes
```
lanierd-adsk, line 41: Actually I think there is a bit of cleanup to do here and in some other unit tests files. 
As we should rather use self._imageVersionFor2Passes which is defined in mtohUtils.py

test/testUtils/mtohUtils.py
```diff
@@ -96,6 +96,9 @@ def setUpClass(cls):
         #Set the usd version
         cls._usdVersion = Usd.GetVersion()
         
+        # Set the image version for 2 passes once per test file
+        cls._imageVersionFor2Passes = None
```
lanierd-adsk, line 100: This is defined for all our unit tests that subclass this class.

lanierd-adsk (changes requested):
Thanks you for doing all this work. A few cleanup tasks to do only.

debloip-adsk:
Hi @jjeziorski @hamedsabri, heads up that with this change, MayaHydra will now build with Hydra Viewport Toolbox (HVT) enabled, which is a separate external dependency (https://github.com/Autodesk/hydra-viewport-toolbox). If you have internal policies that require external dependencies to be handled in a specific way, then this might affect you. For the time being, you can still build without HVT by setting `ENABLE_VIEWPORT_TOOLBOX=OFF`, however we have not tested it extensively, so some additional fixes might be required to get that type of build fully working again. Let us know if that is a concern to you. Thanks!

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraMayaDataProducerSceneIndexData.cpp
debloip-adsk: I think the _prefix data member is still used and relevant for data producers in general, just not in that specific case. This change was done in https://git.autodesk.com/maya3d/maya-hydra/pull/94, maybe @lilike-adsk knows better?

test/lib/mayaUsd/render/mayaToHydra/testMayaShadingModes.py
ppt-adsk: We should enter this as a JIRA ticket then.

ppt-adsk (approved):
Briefly looked at all files (except images), looks good to me.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraMayaDataProducerSceneIndexData.cpp
lilike-adsk: _prefix is still used but the reset call (_prefix = SdfPath::AbsoluteRootPath();)  should be removed, probably a conflict resolving glitch?

test/lib/mayaUsd/render/mayaToHydra/testMayaShadingModes.py
debloip-adsk: Yeah I started working on it but the changes seem a bit more substantial than I expected, so since it's not a functional thing I think it's better to do this separately. Logged HYDRA-1891.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraMayaDataProducerSceneIndexData.cpp
debloip-adsk: Well the `_prefix = SdfPath::AbsoluteRootPath`  call is indeed removed, this is what's intended right?

lanierd-adsk (approved):
A few typo and leftovers, but nothing blocking from merging this.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraMayaDataProducerSceneIndexData.cpp
lilike-adsk: Right

debloip-adsk:
Merging this PR as-is to keep things simple. Test cleanup issue is logged, and I will open a separate PR for the typos once things are settled.

PR #321, opened 2025-09-24
Fix typo and surplus include

debloip-adsk:


lanierd-adsk (approved):
Thanks for these fixes !

PR #323, opened 2025-09-25
HYDRA-1844 - Implementation to translate custom attributes

lilike-adsk:
**Add support for translating custom attributes on Maya nodes, here's how this translation works:**
1. When Hydra is asking for PrimvarDescriptors on a prim, MayaHydra will get all custom attributes with non-default value from the corresponding Maya node and add them as custom primvars to prim.
2. Then later, a HdRenderDelegate can get all the custom primvars on a prim and intepret them,  in some cases, a HdRenderDelegate may want to re-map the custom primvars to what it expects, if that's the case, see the added customPrimvarSceneIndexPlugin example on how to ahieve that.

**Known Issues:**
1. Currently only certain types of Maya custom attributes are supported, will add support for new types in the future if needed


lilike-adsk:
@jjeziorski As mentioned earlier, here's the PR for custom attributes translation on Maya nodes (mesh, light, camera, etc.).  In your case, those custom attributes(with non-default value) in Maya camera node will be added to the camera prim automatically under Hydra Viewport, a render delegate can then fetch those custom primvars. Let me know if any comments.

lib/mayaHydra/hydraExtensions/adapters/adapter.cpp
```diff
@@ -121,4 +129,42 @@ MStatus MayaHydraAdapter::Initialize()
     return status;
 }
 
+HdPrimvarDescriptorVector MayaHydraAdapter::GetPrimvarDescriptors(HdInterpolation interpolation)
+{
+    // All extension attributes as custom primvars
+    if (interpolation == HdInterpolationConstant) {
+        if (_extAttrMapNeedUpdate) {
+            // Apply a global lock to avoid race condition while doing parallel DG node evaluation.
+            //std::lock_guard<LockType> lock(dg_access_mutex);
```
ppt-adsk: Commented out?

ppt-adsk (commented):
Just the minor question about the commented-out lock guard.

lib/mayaHydra/hydraExtensions/adapters/adapter.cpp
lilike-adsk: Good catch, this is unintentional, will fix it

ppt-adsk (dismissed):

ppt-adsk (dismissed):

jjeziorski:
> @jjeziorski As mentioned earlier, here's the PR for custom attributes translation on Maya nodes (mesh, light, camera, etc.). In your case, those custom attributes(with non-default value) in Maya camera node will be added to the camera prim automatically under Hydra Viewport, a render delegate can then fetch those custom primvars. Let me know if any comments.

Thanks for heads up. I'll have a look and let you know if I run into any issues.

ppt-adsk (approved):

lilike-adsk:
> > @jjeziorski As mentioned earlier, here's the PR for custom attributes translation on Maya nodes (mesh, light, camera, etc.). In your case, those custom attributes(with non-default value) in Maya camera node will be added to the camera prim automatically under Hydra Viewport, a render delegate can then fetch those custom primvars. Let me know if any comments.
> 
> Thanks for heads up. I'll have a look and let you know if I run into any issues.

Thanks, I'm merging it, we can revisit this if any issue was found.

PR #324, opened 2025-10-01
Lanierd/hydra 1802 : Render purpose broken

lanierd-adsk:
Render purpose broken with HVT


lib/flowViewport/sceneIndex/fvpPassFilteringSceneIndex.cpp
```diff
@@ -361,6 +349,29 @@ SdfPathVector PassFilteringSceneIndex::GetChildPrimPaths(const SdfPath& primPath
     return filteredChildPaths;
 }
 
+void PassFilteringSceneIndex::DirtyPrimsFromPurposeRenderTag(const TfToken purposeRenderTag)
+{
+    HdSceneIndexObserver::RemovedPrimEntries removedEntries;
+    HdSceneIndexObserver::AddedPrimEntries   addedEntries;
+    auto& inputSceneIndex = GetInputSceneIndex();
+    if (inputSceneIndex) { 
+        for (const SdfPath& path : HdSceneIndexPrimView(inputSceneIndex)) {
```
ppt-adsk, line 359: This traverses the whole hierarchy.  Is this what we want?  Sounds expensive.

lib/flowViewport/sceneIndex/fvpPassFilteringSceneIndex.cpp
```diff
@@ -361,6 +349,29 @@ SdfPathVector PassFilteringSceneIndex::GetChildPrimPaths(const SdfPath& primPath
...
+            HdSceneIndexPrim prim = inputSceneIndex->GetPrim(path);
+            const TfToken    purposeRenderTagFromPrim = GetPurposeRenderTag(prim.dataSource);//From fvpUtils
+            if (purposeRenderTag == purposeRenderTagFromPrim){
+                removedEntries.emplace_back(path);
+                addedEntries.emplace_back(path, prim.primType);
```
ppt-adsk: As per
https://openusd.org/release/api/_page__hydra__getting__started__guide.html
I think just adding the prim is equivalent to a re-sync, so you don't need to remove it.

In any case, as per the above, if removal is necessary (which I don't believe it is), removal of the root of the hierarchy is sufficient.

lib/flowViewport/fvpFramePassData.h
```diff
@@ -72,11 +77,22 @@ struct FramePassData
     //! render prims that do not have a purpose render tag
     bool _supportPrimsWithNoPurposeRenderTag = false;
 
+    //! The filtering scene index for this pass, we cannot use the actual pass filtering scene index class as it also includes this class definition
```
ppt-adsk:
```suggestion
    //! The filtering scene index for this pass, we cannot use Fvp::PassFilteringSceneIndex as it also includes this class declaration
```

lib/flowViewport/fvpFramePassData.h
```diff
@@ -72,11 +77,22 @@ struct FramePassData
...
+    PXR_NS::HdSingleInputFilteringSceneIndexBaseRefPtr _passFilteringSceneIndex;
```
ppt-adsk, line 79: If Fvp::PassFilteringSceneIndex and Fvp::FramePassData are circularly dependent, perhaps FrameDataData should be a nested class of PassFilteringSceneIndex?

Or perhaps FramePassData should not store the PassFilteringSceneIndex and it should be passed to those functions that require it?  Unsure.

lib/flowViewport/fvpFramePassData.h
```diff
@@ -72,11 +77,22 @@ struct FramePassData
...
+
     //! Helper methods to safely access the frame pass
     bool IsValid() const { return _framePass != nullptr; }
     bool HasRenderIndexProxy() const { return _renderIndexProxy != nullptr; }
     const hvt::FramePassPtr& GetFramePass() const { return _framePass; }
     hvt::RenderIndexProxyPtr GetRenderIndexProxy() const { return _renderIndexProxy; }
+    void
+    SetPassFilteringSceneIndex(const PXR_NS::HdSingleInputFilteringSceneIndexBaseRefPtr& sceneIndex)
```
ppt-adsk, line 87: Do we really need this modifier?  Could this data member be set on construction (if we need it)?

lib/flowViewport/fvpFramePassData.cpp
```diff
@@ -0,0 +1,36 @@
...
+PXR_NAMESPACE_USING_DIRECTIVE
+
+namespace FVP_NS_DEF {
+
+void FramePassData::DirtyPrimsFromPurposeRenderTag(const PXR_NS::TfToken purposeRenderTag)
+{    
+    auto passFilteringSceneIndex
+        = TfDynamic_cast<PassFilteringSceneIndexRefPtr>(_passFilteringSceneIndex);
+    if (passFilteringSceneIndex) {
+        passFilteringSceneIndex->DirtyPrimsFromPurposeRenderTag(purposeRenderTag);
```
ppt-adsk, line 31: It seems that the only call to this method is in renderOverride.cpp, which should be able to access PassFilteringSceneIndex and call this method itself, because otherwise there is no use of any of the other FramePassData data members.  Therefore, I question the need for this `FramePassData::DirtyPrimsFromPurposeRenderTag()` method.

lib/flowViewport/fvpFramePassData.h
```diff
@@ -20,9 +20,14 @@
 
 #ifdef VIEWPORT_TOOLBOX
 
+#include <pxr/pxr.h>
 #include <pxr/base/tf/token.h>
 #include <pxr/usd/sdf/path.h>
-#include <pxr/pxr.h>
+#include <pxr/imaging/hd/filteringSceneIndex.h>
+#include <pxr/imaging/hd/sceneIndexObserver.h>
+#include <pxr/imaging/hd/sceneIndexPrimView.h>
```
ppt-adsk: Seem unused.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -155,6 +156,20 @@ PXR_NAMESPACE_USING_DIRECTIVE
...
+    if (attrName == TfToken("mayaHydraRenderPurpose")) {
+        return HdRenderTagTokens->render;
+    } else if (attrName == TfToken("mayaHydraProxyPurpose")) {
+        return HdRenderTagTokens->proxy;
+    } else if (attrName == TfToken("mayaHydraGuidePurpose")) {
+        return HdRenderTagTokens->guide;
+    } else {
+        TF_CODING_ERROR("Unknown purpose attribute name '%s'", attrName.GetText());
+        return {};
+    }
```
ppt-adsk: Style suggestion, do if you feel like it.
```suggestion
    {
        static std::map<TfToken, TfToken> attrToTag{
            {TfToken("mayaHydraRenderPurpose"), HdRenderTagTokens->render},
            {TfToken("mayaHydraProxyPurpose"),  HdRenderTagTokens->proxy},
            {TfToken("mayaHydraGuidePurpose"),  HdRenderTagTokens->guide}};
        auto found = attrToTag.find(attrName);
        if (found != attrToTag.end()) {
            return found->second;
        }
        TF_CODING_ERROR("Unknown purpose attribute name '%s'", attrName.GetText());
        return {};
    }
```

ppt-adsk (commented):
Looks good, just a few questions and minor points.

lib/flowViewport/fvpFramePassData.cpp
lanierd-adsk: Since we usually have several passes, there are multiple passfilteringsceneindices and I thought it would be better to store the passFilteringSCeneIndex into framepass data instead of a separate set as this is where we store every data for a pass. Please contact me directly if you have some more questions about this.

lib/flowViewport/fvpFramePassData.h
lanierd-adsk: Good catch!

lib/flowViewport/fvpFramePassData.h
lanierd-adsk: Fvp::PassFilteringSceneIndex uses Fvp::FramePassData to filter out primitives. But I don't think it would make sense to nest one of this class into another.
Fvp::FramePassData stores the data for a pass, including the pass filtering scene index. I would keep it like that.

lib/flowViewport/fvpFramePassData.h
lanierd-adsk: Yes, in renderoverride.cpp, we store the pass data and only the pass data knows the pass filtering scene index so I think it makes sense this way. We cannot have a constructor that initializes all parameters from frame pass data as some members are initialized in different places at different time.

lib/flowViewport/sceneIndex/fvpPassFilteringSceneIndex.cpp
lanierd-adsk: I don't see any other way of doing this. We need to invalidate in the whole scene the prims that have the purpose render tag.

lib/flowViewport/sceneIndex/fvpPassFilteringSceneIndex.cpp
lanierd-adsk: ok, thanks I didn't know this :  "A PrimsAdded notice can be sent for a prim already in the scene, in which case it should be interpreted as a resync"
I don't find this logical, but this is how it is in OpenUSD.

ppt-adsk (approved):

PR #327, opened 2025-10-06
HYDRA-1930 : Point lights don't cast shadows with Storm

lanierd-adsk:


lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -1913,5 +1914,14 @@ void MayaHydraSceneIndex::SetLightsManagementSceneIndex(const Fvp::LightsManagem
     _lightsManagementSceneIndex = lightsManagementSceneIndex;
 }
 
+GfBBox3d MayaHydraSceneIndex::GetBoundingBox()const
+{
+    GfBBox3d bbox;
+    _MapAdapter<MayaHydraAdapter>(
+        [&](MayaHydraAdapter* a) { bbox = GfBBox3d::Combine(a->GetBoundingBox(), bbox); },
```
lilike-adsk, line 1921: I recall the boundingbox for renderItemAdapter is not correctly setup. see the changes at https://git.autodesk.com/maya3d/maya-hydra/blob/production_rendering_poc/lib/mayaHydra/hydraExtensions/adapters/renderItemAdapter.cpp#L200

lilike-adsk (commented):
Nice work, see my comments. BTW, why the plane in the test scene is not visible in some new baseline images?

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
lanierd-adsk: Thanks for the fix !

lanierd-adsk:
> BTW, why the plane in the test scene is not visible in some new baseline images?

Because I enlarged it so the shadows are casted on it. When it's too small, shadows are not casted on it.

lilike-adsk (approved):

lilike-adsk (approved):

PR #328, opened 2025-10-06
HYDRA-1918 : Minimize shader compilations for selection highlights

debloip-adsk:


lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
```diff
@@ -352,56 +376,134 @@ class _RerootingSceneIndexContainerDataSource : public HdContainerDataSource
     HdContainerDataSourceHandle const _inputDataSource;
 };
 
+class _MaterialBlockingContainerDataSource : public HdContainerDataSource {
```
debloip-adsk, line 379: Since materials can be bound in different ways, having these material blocking data sources allows for a generic way to easily block material path data sources wherever they are located in the hierarchy. They essentially act like overlays on top of the original data sources, with extra processing to figure out if a material path data source is about to be returned, and block it if need be.

lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
```diff
@@ -824,6 +926,79 @@ BaseWhSi::CollectInstancingPaths(const PXR_NS::SdfPath& primPath, InstancingPath
     }
 }
 
+//We want to set the displayStyle of the highlighting prim with following overrides:
+// 1. A "reprSelector" with HdReprTokens->wire or HdReprTokens->refinedWire, means only draw wires.
+// 2. An "overrideWireframeColor" primvar with the color of the highlighting.
+HdContainerDataSourceHandle
+BaseWhSi::SetWireframeRepr(const HdContainerDataSourceHandle& dataSource, const GfVec4f& color) const
```
debloip-adsk, line 933: Moved this into the BaseWhSi class since we need access to the input scene index to get the materials and check if they have displacement

lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
```diff
@@ -824,6 +926,79 @@ BaseWhSi::CollectInstancingPaths(const PXR_NS::SdfPath& primPath, InstancingPath
...
+    }
+#else
+    auto usdMaterialBindings = UsdImagingDirectMaterialBindingsSchema::GetFromParent(dataSource);
+    if (usdMaterialBindings.IsDefined()) {
+        edited.Set(
+            UsdImagingDirectMaterialBindingsSchema::GetDefaultLocator(), 
+            _MaterialBlockingContainerDataSource::New(usdMaterialBindings.GetContainer(), GetInputSceneIndex(), TfToken("materialPath"))
+        );
+    }
+#endif
```
debloip-adsk, line 974: This is the new code of SetWireframeRepr

test/lib/mayaUsd/render/mayaToHydra/cpp/testMaterialBlocking.cpp
```diff
@@ -0,0 +1,88 @@
...
+PXR_NAMESPACE_USING_DIRECTIVE
+
+TEST(MaterialBlocking, testMaterialBlocking)
+{
+    // Get the selection highlight scene index
+    const auto& sceneIndices = GetTerminalSceneIndices();
+    ASSERT_GT(sceneIndices.size(), 0u);
+    size_t iSceneIndex = 0;
+    for (;iSceneIndex < sceneIndices.size(); iSceneIndex++) {
+        if (!sceneIndices[iSceneIndex]->GetChildPrimPaths(SdfPath("/Pass1")).empty()) {
```
lanierd-adsk: Could you please do a function in some utils file ? I think it would be good to share this in the code and have a single place to change it if we want not to rely on the naming which seems fragile. 
Probably not only in cpp tests but having it in the usual code in a util* file would be good.

lanierd-adsk (changes requested):
Thank you for doing this, just one change to share a function.

test/lib/mayaUsd/render/mayaToHydra/cpp/testMaterialBlocking.cpp
debloip-adsk: I extracted it out in the test utils, since getting the scene index for a pass in this way is pretty much exclusive to the unit tests. If we need to give access to the pass scene indices in the main code, there are probably better ways than this since we'd have access to the source data

lanierd-adsk (approved):

PR #330, opened 2025-10-14
HYDRA-1572 : Fix z-fighting for selection highlighting

debloip-adsk:


debloip-adsk:
Note regarding the updated reference images : since the z-fighting fix is only available in 25.08+, we need alternative versions (yet again) of reference images. Paired with the two_passes version, this is getting a bit of a mess. On top of that, it seems like some of our tests' thresholds might be a bit too low, because I would have expected a lot more tests to fail or give different results without the z-fighting. I think what we should do is that once we also have the fix for MSAA, we should go over our tests and cleanup to remove all the extra reference images, and also make sure that the thresholds are meaningful and detect changes like z-fighting.

lanierd-adsk (approved):

PR #331, opened 2025-10-20
HYDRA-1925 : Fix the viewport constantly refreshing when using HVT

debloip-adsk:
The issue was that the method we previously used to replace the default selection tasks did not actually replace them in the HVT task manager, so they were never executed and thus never converged, but were still considered when calling IsConverged. Since we have a callback running on a timer to refresh the viewport when the render has not converged (for progressive rendering), this lead to constant viewport refreshing.

### DIFF
--- a/lib/mayaHydra/mayaPlugin/renderOverride.cpp
+++ b/lib/mayaHydra/mayaPlugin/renderOverride.cpp
@@ -171,11 +171,6 @@ TfToken _GetPurposeRenderTagFromAttrName(const TfToken& attrName)
     return {};
 }
 
-inline bool areDifferentForOneOfTheseBits(unsigned int val1, unsigned int val2, unsigned int bitsToTest)
-{
-    return ((val1 & bitsToTest) != (val2 & bitsToTest));
-}
-
 inline bool isInComponentsPickingMode(const MHWRender::MSelectionInfo& selectInfo)
 {
     return selectInfo.selectable(MSelectionMask::kSelectMeshVerts)
@@ -192,6 +187,7 @@ inline bool isInComponentsPickingMode(const MHWRender::MSelectionInfo& selectInf
 // configurable and cannot be replaced by plugin behavior.  Currently, the Flow
 // Viewport selection task is a no-op.  PPT, 2-Oct-2023.
 
+#ifndef VIEWPORT_TOOLBOX
 void replaceSelectionTask(HdTaskSharedPtrVector* tasks)
 {
     // For TF_WARN and TF_AXIOM macros.
@@ -213,6 +209,7 @@ void replaceSelectionTask(HdTaskSharedPtrVector* tasks)
 
     *found = HdTaskSharedPtr(new Fvp::SelectionTask);
 }
+#endif
 
 #ifdef MAYA_HAS_VIEW_SELECTED_OBJECT_API
 std::string getRenderingDestination(
@@ -908,7 +905,6 @@ MStatus MtohRenderOverride::Render(
                 TF_WARN("HdxProgressiveTask not found");
             }
         }
-#endif
 
         auto editTasks = [](HdTaskSharedPtrVector&  tasksToEdit,
                             MayaHydraGLBackup&      backup) -> void {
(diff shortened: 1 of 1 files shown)


lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -171,11 +171,6 @@ TfToken _GetPurposeRenderTagFromAttrName(const TfToken& attrName)
     return {};
 }
 
-inline bool areDifferentForOneOfTheseBits(unsigned int val1, unsigned int val2, unsigned int bitsToTest)
-{
-    return ((val1 & bitsToTest) != (val2 & bitsToTest));
-}
-
```
debloip-adsk, line 178: This is not related to the fix, but is dead code that we've been carrying for a while. Might as well remove it now.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -192,6 +187,7 @@ inline bool isInComponentsPickingMode(const MHWRender::MSelectionInfo& selectInf
 // configurable and cannot be replaced by plugin behavior.  Currently, the Flow
 // Viewport selection task is a no-op.  PPT, 2-Oct-2023.
 
+#ifndef VIEWPORT_TOOLBOX
 void replaceSelectionTask(HdTaskSharedPtrVector* tasks)
```
debloip-adsk, line 191: Old method for replacing the default selection tasks which did not work with HVT. Made it non-HVT only.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -2771,6 +2762,10 @@ void MtohRenderOverride::_CreateFramePass(
     framePassDescriptor.uid         = passId;
     auto framePass                  = hvt::ViewportEngine::CreateFramePass(framePassDescriptor);
 
+    // Remove the default selection tasks as we do not use them.
+    framePass->GetTaskManager()->RemoveTask(HdxPrimitiveTokens->colorizeSelectionTask);
+    framePass->GetTaskManager()->RemoveTask(TfToken("selectionTask"));
```
debloip-adsk, line 2767: Unfortunately there is no token for the HdxSelectionTask

debloip-adsk (commented):
Note : previously we replaced the default selection tasks with our custom one. However for now I just removed the default ones and did not add back our custom one, as it doesn't do anything yet. Adding it might require a bit of tweaking to integrate it with the HVT infrastructure, and I feel it's better to cross that bridge when we actually get there.

ppt-adsk:
Is there a way to test this, e.g. using a counter in Fvp::Instruments, with a non-progressive (i.e. Storm) renderer?  If in the test we wait for the timer callback period (or twice that long), and change nothing in the scene, the number of redraws should not have changed.

debloip-adsk:
We can't really unit test this as the timer callback is called when Maya executes the idle events, and these are not run with our current unit test setup. See the comment on HYDRA-816 for more info about the idle events not being run

debloip-adsk:
Here's a rough thought : we could maybe expose the timer callback in MtohRenderOverride, and add a MEL/Python command that would trigger it manually, and use that to replicate the workflow in unit tests.

There are probably other ways to go about it though. Ultimately, the goal here would be to detect when Hydra (or Storm) renders don't converge, which would lead to this viewport refreshing issue. Let me know what you think

ppt-adsk:
> We can't really unit test this as the timer callback is called when Maya executes the idle events, and these are not run with our current unit test setup. See the comment on HYDRA-816 for more info about the idle events not being run

Ah yes, forgot about that, quite right, there is no idle loop.

ppt-adsk:
> Here's a rough thought : we could maybe expose the timer callback in MtohRenderOverride, and add a MEL/Python command that would trigger it manually, and use that to replicate the workflow in unit tests.
> 
> There are probably other ways to go about it though. Ultimately, the goal here would be to detect when Hydra (or Storm) renders don't converge, which would lead to this viewport refreshing issue. Let me know what you think

I like where you're going.  With a progressive renderer like Arnold, it seems difficult to test for convergence, as this is scene dependent and not predictable.  However, if we can reproduce the problem with Storm, then I'd vote for a test.

ppt-adsk (approved):

PR #332, opened 2025-10-21
HYDRA-1925 : Add unit test to check for rendering convergence with Storm

debloip-adsk:
Follow-up from the conversation in #331

### DIFF
--- a/lib/mayaHydra/mayaPlugin/testingCommand.h
+++ b/lib/mayaHydra/mayaPlugin/testingCommand.h
@@ -1,0 +1,39 @@
+//
+// Copyright 2025 Autodesk
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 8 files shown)


lib/mayaHydra/mayaPlugin/testingCommand.h
ppt-adsk, line 1: Could this have been wedged into the existing mayaHydraCppTest command, or would we be stretching the purpose of that command beyond its purpose?  I'm fine if separation of concerns leads us to another testing command.

debloip-adsk: I think into the command itself doesn't make much sense, but if you're talking about the mayaHydraCppTestsCmd.cpp file, we potentially could as it also contains the mayaHydraInstruments command. Personally I would prefer keeping them separate, as if we add things into this testing command in the future, we'd start to have quite a bit of unrelated code in the same file.

ppt-adsk (approved):

PR #335, opened 2025-10-27
HYDRA-1909 - Fix picking crash on instance with guide purpose

lilike-adsk:
With multi FramePass enabled, when resolving HdxPickHit data returned from Hydra renderers, we need to know which FramePass (RenderIndex) the hit prim belongs to. To archieve that, we add a _passIndex_ data on top of HdxPickHit data structure.

### DIFF
--- a/lib/mayaHydra/hydraExtensions/pick/mhPickHit.h
+++ b/lib/mayaHydra/hydraExtensions/pick/mhPickHit.h
@@ -36,8 +36,8 @@ struct PickHit
         : passIndex(passIdx)
         , hdxPickHit(pickHit) {}
 
-    int        passIndex; // FramePass index.
-    HdxPickHit hdxPickHit;
+    const int        passIndex; // FramePass index.
+    const HdxPickHit hdxPickHit;
 };
 
 }
--- a/lib/mayaHydra/hydraExtensions/pick/mhPickHit.h
+++ b/lib/mayaHydra/hydraExtensions/pick/mhPickHit.h
@@ -1,0 +1,45 @@
+//
+// Copyright 2025 Autodesk
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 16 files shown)


lib/mayaHydra/hydraExtensions/pick/mhUsdPickHandler.cpp
```diff
@@ -280,7 +281,7 @@ UsdPickHandler::HitPath resolveInstancePicking(HdRenderIndex& renderIndex, const
 {
     auto primOrigin = HdxPrimOriginInfo::FromPickHit(&renderIndex, pickHit);
 
-    if (pickHit.instancerId.IsEmpty()) {
+    if (pickHit.instancerId.IsEmpty() || primOrigin.instancerContexts.empty()) {
```
lilike-adsk, line 284: Here's the crash point, we were attempting to access primOrigin.instancerContexts.front() directly without checking the size of vector. Add a check to avoid any potential crash in the future.

lilike-adsk: With the passIndex added, we can get the instrancer populated in the primOrigin.instancerContexts now.

lib/mayaHydra/hydraExtensions/pick/mhPickHit.h
```diff
@@ -0,0 +1,45 @@
...
+/// needed to implement picking.
+///
+struct PickHit
+{
+    PickHit(int passIdx, HdxPickHit pickHit)
+        : passIndex(passIdx)
+        , hdxPickHit(pickHit) {}
+
+    int        passIndex; // FramePass index.
+    HdxPickHit hdxPickHit;
```
ppt-adsk: Could these both be const?

ppt-adsk (commented):
Nice work!  Is there any way we can write a test for this?

lib/mayaHydra/hydraExtensions/pick/mhPickHit.h
lilike-adsk: yes, they could

lilike-adsk:
> Nice work! Is there any way we can write a test for this?

added one test

ppt-adsk (approved):

PR #341, opened 2025-11-13
HYDRA-1943/HYDRA-1963 : Improve pass filtering scene index

debloip-adsk:
This PR makes the pass filtering scene index fully implement notification handling, and also exclude materials by default unless required by rprims of the pass.

As a result of this change, some other fixes also had to be made.
- Fix bounding box scene index to not require extents (should have been part of #279 but was missed)
- Make MayaHydraSceneIndex::_CreateAdapter add the adapter to its adapter map before populating
- Added a new scene index (PrimRemovalEnforcingSceneIndex) to work around a bug in OpenUSD code


lib/flowViewport/fvpFramePassData.h
```diff
@@ -67,9 +66,6 @@ struct FramePassData
     //! Paths to force exclude from this frame pass
     PXR_NS::SdfPathVector _excludePaths;
 
-    //! Whether to keep lights in this frame pass
-    bool _keepLights = false;
```
debloip-adsk, line 71: Removed this flag, it was set to true in both passes. Easy to re-implement later if needed.

lib/flowViewport/fvpUtils.cpp
```diff
@@ -110,6 +114,24 @@ SdfPath GetMaterialPath(const PXR_NS::HdContainerDataSourceHandle& primDataSourc
     return bindingPathDataSource->GetTypedValue(0);
 }
 
+bool MaterialHasDisplacement(const PXR_NS::HdSceneIndexPrim& materialPrim) {
```
debloip-adsk, line 117: Moved from fvpBaseWhSi

lib/flowViewport/sceneIndex/fvpBBoxSceneIndex.cpp
```diff
@@ -486,12 +486,6 @@ bool BboxSceneIndex::_ShouldConvertToBoundingBox(
         return false;
     }
 
-    // Check if the prim has an extent (required for bounding box creation)
-    HdExtentSchema extentSchema = HdExtentSchema::GetFromParent(prim.dataSource);
-    if (!extentSchema.IsDefined()) {
-        return false;
-    }
```
debloip-adsk, line 493: Bugfix. We no longer require extents for bounding box creation as per HYDRA-1578 and its PR #279

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -1629,9 +1629,15 @@ AdapterPtr MayaHydraSceneIndex::_CreateAdapter(
     if (adapter == nullptr || !adapter->IsSupported()) {
         return {};
     }
+    // Order is important here. We need to add the adapter to the map before populating.
+    // Why : once we send PrimsAdded notifications, a downstream scene index might query 
+    // a prim that uses a MayaHydraDataSource and try to get its material. Since 
+    // MayaHydraDataSource::_GetMaterialBindingDataSource() calls MayaHydraSceneIndex::GetMaterialId(), 
+    // which in turn does lookups into the adapter maps, the adapter map needs to contain 
+    // the desired prim's adapter.
+    adapterMap.insert({ id, adapter });
```
debloip-adsk, line 1638: This feels a bit brittle. Should we log a tech debt ticket?

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -770,7 +772,6 @@ SdfPathVector MtohRenderOverride::RendererRprims(TfToken rendererName, bool visi
             primIds.insert(
                 primIds.end(), tempPrimIds.begin(), tempPrimIds.end()); // Insert all elements
         }
-        break; // We found the right frame pass and its render index, no need to continue, data has been taken from the renderindex
```
debloip-adsk, line 773: We now cumulate the prims from all frame passes using the given renderer instead of just the first one that matches, to preserve previous behavior. Prims are usually going to only be rendered in one pass anyways.

test/lib/mayaUsd/render/mayaToHydra/cpp/testPassFiltering.cpp
```diff
@@ -0,0 +1,120 @@
...
+    auto sphereWireframePrims = SceneIndexInspector(passSceneIndices[1]).FindPrims(findSphereWireframePredicate);
+    ASSERT_EQ(sphereWireframePrims.size(), 1u);
+    auto sphereWireframePath = sphereWireframePrims.front().primPath;
+
+    // Maya mesh prim
+    testPrim(
+        SdfPath("/MayaHydraViewportRenderer/materials/openPBRSurface1SG"), 
+        HdPrimTypeTokens->material,
+        {0}
+    );
```
debloip-adsk, line 80: We use hardcoded paths by default and query the scene for prims whose names change (e.g. instancers with a bunch of numbers at the end). That way we test both methods : direct prim query vs first querying the scene and then the prim.

lib/flowViewport/sceneIndex/fvpPrimRemovalEnforcingSceneIndex.h
```diff
@@ -0,0 +1,137 @@
...
+#include <pxr/usd/sdf/path.h>
+#include <pxr/usd/sdf/pathTable.h>
+
+namespace FVP_NS_DEF {
+
+class PrimRemovalEnforcingSceneIndex;
+typedef PXR_NS::TfRefPtr<PrimRemovalEnforcingSceneIndex> PrimRemovalEnforcingSceneIndexRefPtr;
+typedef PXR_NS::TfRefPtr<const PrimRemovalEnforcingSceneIndex> PrimRemovalEnforcingSceneIndexConstRefPtr;
+
+/// \class PrimRemovalEnforcingSceneIndex
```
debloip-adsk, line 32: A "fun" read and a worthy challenger to https://github.com/Autodesk/maya-hydra/blob/dev/cmake/compiler_config.cmake#L116-L152 :)

lib/flowViewport/sceneIndex/fvpPassFilteringSceneIndex.cpp
```diff
@@ -296,31 +113,123 @@ bool PassFilteringSceneIndex::_IsFilteredOut(const SdfPath& primPath) const
...
         }
     } else {
         // No data source
     }
 
     return false; // will be rendered.
 }
-#endif
+
+HdSceneIndexObserver::AddedPrimEntries PassFilteringSceneIndex::_UpdateFilteringStatus(const PXR_NS::SdfPath& primPath, bool dirtied, bool resync)
```
debloip-adsk, line 142: We make _UpdateFilteringStatus and _RemoveMaterialEntry return notifications rather than send them immediately so we can avoid a bunch of individual _SendPrimsXYZ() calls and have finer control over when we send them.

lib/mayaHydra/hydraExtensions/sceneIndex/registration.cpp
```diff
@@ -238,7 +239,13 @@ void MayaHydraSceneIndexRegistry::_AddSceneIndexForNode(MObject& dagNode)
         registration->pluginSceneIndex,
         registration->sceneIndexPathPrefix);
 
-    registration->rootSceneIndex = pfsi;
+    // Add a scene index that enforces a coherent prim removal + prim retrieval behavior.
+    // This is to work around a bug in OpenUSD with the UsdImagingDrawModeSceneIndex where 
+    // it can send PrimsRemoved notifications, but still return valid prims and prim paths 
+    // when calling GetPrim and GetChildPrimPaths afterwards.
+    auto primRemovalSi = Fvp::PrimRemovalEnforcingSceneIndex::New(pfsi);
```
debloip-adsk, line 250: Forgot to mention, adding the hacky scene index only on USD stages to avoid potentially hiding too many bugs

test/lib/mayaUsd/render/mayaToHydra/cpp/testPassFiltering.cpp
ppt-adsk, line 1: Nice.

lib/mayaHydra/hydraExtensions/sceneIndex/registration.cpp
ppt-adsk: Can you add a comment as to which USD version this started appearing in, and a link to the code?  I think it would be useful.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
ppt-adsk: Agree.  It feels like this should be the responsibility of the adapterCreator(), as creating the adapter and inserting the adapter into the adapter map should be a single operation.

lib/flowViewport/sceneIndex/fvpPrimRemovalEnforcingSceneIndex.h
ppt-adsk: Thanks for the thorough description.

lib/flowViewport/sceneIndex/fvpPassFilteringSceneIndex.h
```diff
@@ -68,51 +77,44 @@ class PassFilteringSceneIndex
...
+    PXR_NS::HdSceneIndexObserver::AddedPrimEntries _UpdateMaterialEntry(const PXR_NS::SdfPath& primPath);
+
     Fvp::FramePassConstDataPtr _framePassData;
 
 private:
-    // Helper function to determine if a prim should be included in all passes
-    bool _ShouldIncludeInAllPasses(const PXR_NS::HdSceneIndexPrim& prim) const;
+    PXR_NS::SdfPathSet _filteredPrims;
+
+    // Used to track the materials required by the Rprims that will actually be rendered in the pass
```
ppt-adsk: Isn't rprim Hydra 1.0 terminology?  The interpretation is a renderable prim, right?  Rprim might still be the best way to describe this.

ppt-adsk (dismissed):
Terrific work, and great catch for the merging scene index behavior change.  Minor comments only.

lib/mayaHydra/hydraExtensions/sceneIndex/registration.cpp
debloip-adsk: Hard to say when the issue in UsdImagingDrawModeSceneIndex itself was introduced, as it only became a problem with the change in HdMergingSceneIndex. For all I know it might've always been a thing. Still mentioned the version and the change that uncovered the issue, and to refer to the hacky scene index file for more details.

lib/flowViewport/sceneIndex/fvpPassFilteringSceneIndex.h
debloip-adsk: Yeah I'm not too sure either. I think there is no longer a hard distinction between Rprim/Sprim/Bprim in a Hydra 2.0 scene, but the terminology is still used often as it's still useful to reason about things conceptually. But yes in this case, we don't actually need to be explicit about Rprims, it could be any type of prim and it would be handled the same. Adjusted the naming a bit.

ppt-adsk (approved):

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
debloip-adsk: Logged HYDRA-1992

PR #344, opened 2025-11-17
HYDRA-1849 : instance selection highlight color is wrong

lanierd-adsk:
update the comment and image but the bug has been resolved by another PR


debloip-adsk (approved):
I'm curious, do you know what is the change that fixed it?

lanierd-adsk:
@debloip-adsk no I have no idea which change fixed it.

PR #345, opened 2025-11-17
HYDRA-1975 - fix aov sharing crash

lilike-adsk:
Fix the AOV buffer sharing issues:
1. Consume latest HVT which includes updated render buffer management changes
2. Remove the call to UsdImagingGLEngine::GetRenderAovs() as it may interfere with the current renderer
3. Remove the call to GetRenderBuffer(aovName) as the AOV may not have been created yet for first frame
4. To match what HVT expects, if a non-color AOV was visualized previously, we only share that one


lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -292,6 +292,30 @@ int GetNearestHitIndex(
 
     return nearestHitIndex;
 }
+
+// Can't rely on UsdImagingGLEngine::GetRenderAovs() as it may interfere with the current renderer
+TfTokenVector _GetRenderAovs(HdRenderIndex* renderIndex)
```
lilike-adsk: Copy the implementation of UsdImagingGLEngine::GetRenderAovs() here, will ask HVT to see if it can provide one utility to do this query.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -689,9 +713,7 @@ TfTokenVector MtohRenderOverride::GetAvailableFramePassAovs(int passIndex)
     for (auto* instance : _allInstances) {
         if (instance->_initializationSucceeded
             && passIndex < static_cast<int>(instance->_framePassesData.size())) {
-            UsdImagingGLEngine engine(
```
lilike-adsk, line 682: Reuse same hgi/hgiDriver in a new engine may interfere with the existing engine

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -998,8 +1017,13 @@ MStatus MtohRenderOverride::Render(
             currentPass->params().enablePresentation = isLastVisiblePass;
             
             // Set the AOV to visualize for the current pass if it exists
-            const TfToken aovName       = TfToken(visibleAOVNames[visibleIdx].asChar());
-            const bool    aovNameExists = currentPass->GetRenderBuffer(aovName) != nullptr;
+            const TfToken aovName = TfToken(visibleAOVNames[visibleIdx].asChar());
```
lanierd-adsk, line 1020: This code is wrong in the sense that you only have hardcoded AOVs :
HdAovTokens->primId,
HdAovTokens->depth,
HdAovTokens->normal,
HdAovTokens->Neye,
HdAovTokensMakePrimvar(TfToken("st"))

But the render delegate , like Arnold, has some other AOVs available that the user may want to use for display. I don't think we should check in the visibleAOVs name if it exists sibnce this list is hardcoded.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -998,8 +1017,13 @@ MStatus MtohRenderOverride::Render(
...
+            // Can't rely on GetRenderBuffer(aovName) here as the AOV may not have been created yet
+            const auto renderDelegate = _GetRenderDelegate(actualPassIndex);
+            const bool aovNameExists
```
lanierd-adsk, line 1023: Here we are doing the same test as in visibleAOVNames, if it is in AOVNames then this will always be true since we tested for the same thing in MtohRenderOverride::GetAvailableFramePassAovs. But I don't think we should use this hardcoded list.

lanierd-adsk (changes requested):
I don't think using a hardcoded list is correct, the render delegate are proposing custom AOVs that are not part of the hardcoded list we use.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lilike-adsk: This is the same code used in UsdImagingGLEngine::GetRenderAovs() , yes, a better way is required in the future, before that, we have to copy that piece of code here as using UsdImagingGLEngine::GetRenderAovs() will affect the current rendering.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lilike-adsk: We still need to validate if the AOV is supported by a renderer here, as user may specify a non-supported AOV by the commad and we didn't validate the user input.

lanierd-adsk: ok thanks. I was confused and thought we needed to have the diaplsy aov be part of the hardcoded list which is not the case.

lanierd-adsk (approved):

PR #349, opened 2025-11-20
HYDRA-1821 : Fix MSAA with 2 passes

debloip-adsk:
In this PR : 

- Update to latest HVT & adjust our rendering code very slightly to fix MSAA
- Remove all the "two_passes" images that were added to deal with the MSAA not working
- Re-baseline new & old reference images for MSAA (& USD 25.08 in some cases)
- Fix testFramePasses reference images (looks like the previous reference images had incorrect results and it works now)


test/lib/mayaUsd/render/mayaToHydra/testFramePasses.py
```diff
@@ -94,7 +88,7 @@ def test_FootPrintNodeForSecondPass(self):
             cmds.refresh()
 
             self.setBasicCam(1)
-            self.EnablePassesOneByOneAndDoSnapshots(name="FootPrintAsScndGraphics", failures=failures, useDynamicVersion=True)#Is dependent on the number of frame passes
```
lilike-adsk, line 97: I think the new baselines for this case look different is because the old baselines were located at "two_passes" folder.

lilike-adsk (approved):
Great clean up!

test/lib/mayaUsd/render/mayaToHydra/testFramePasses.py
debloip-adsk: Ah yes that makes sense, good catch thanks

PR #350, opened 2025-11-21
Fixed logic, added proper initialization, and added OpenPBR support.

ppt-adsk:


lib/flowViewport/sceneIndex/fvpPruneTexturesSceneIndex.cpp
```diff
@@ -44,6 +43,7 @@ _PruneTexturesFromMatNetwork(
     for (TfToken const &nodeName : nodeNames) {
         const TfToken nodeType = networkInterface->GetNodeType(nodeName);
         if (nodeType == _tokens->ND_standard_surface_surfaceshader ||
+            nodeType == _tokens->ND_open_pbr_surface_surfaceshader ||
```
ppt-adsk, line 46: This is the core fix, and it feels fragile.  Perhaps we won't be adding support for shaders very often, but how do we know we haven't forgotten any?  Could a plugin data model have its own surface shader, and we wouldn't know about it?  Perhaps in that latter case it would be the responsibility of the data model plugin to add a texture pruning scene index, but it still feels like we're missing a level of abstraction, or perhaps some kind of registration mechanism.

lib/flowViewport/sceneIndex/fvpPruneTexturesSceneIndex.cpp
```diff
@@ -60,16 +60,17 @@ _PruneTexturesFromMatNetwork(
 // static
 PruneTexturesSceneIndexRefPtr
 PruneTexturesSceneIndex::New(
-    const HdSceneIndexBaseRefPtr &inputSceneIndex)
+    const HdSceneIndexBaseRefPtr &inputSceneIndex,
+    bool pruneTextures)
```
ppt-adsk, line 64: Initial setting was don't prune (keep textures), which is wrong, so allow passing in setting on scene index tree creation.

lib/flowViewport/sceneIndex/fvpPruneTexturesSceneIndex.cpp
```diff
@@ -60,16 +60,17 @@ _PruneTexturesFromMatNetwork(
...
-        new PruneTexturesSceneIndex(inputSceneIndex));
+        new PruneTexturesSceneIndex(inputSceneIndex, pruneTextures));
 }
 
 void
-PruneTexturesSceneIndex::MarkTexturesDirty(bool isTextured)
+PruneTexturesSceneIndex::MarkTexturesDirty(bool pruneTextures)
 {
-    _needsTexturesPruned = isTextured;
+    _pruneTextures = pruneTextures;
```
ppt-adsk, line 73: Argument was called isTextured, but data member is the opposite logic, pruneTextures.  Made everything consistent.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1291,7 +1291,7 @@ MStatus MtohRenderOverride::Render(
     bool isTextured = currentDisplayMode & MHWRender::MFrameContext::kTextured;
     if (_pruneTexturesSceneIndex &&
         _currentlyTextured != isTextured) {
-        _pruneTexturesSceneIndex->MarkTexturesDirty(isTextured);
+        _pruneTexturesSceneIndex->MarkTexturesDirty(!isTextured);
```
ppt-adsk: Now that the prune textures scene index uses prune textures consistently in its code (opposite of is textured), must negate the argument here.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1984,8 +1984,9 @@ void MtohRenderOverride::_CreateSceneIndicesChainAfterMergingSceneIndex(const MH
     _displayStyleSceneIndex->addExcludedSceneRoot(MAYA_NATIVE_ROOT); // Maya native prims don't use global refinement
 
     // Add texture disabling Scene Index
+    bool isTextured = drawContext.getDisplayStyle() & MHWRender::MFrameContext::kTextured;
     _lastFilteringSceneIndexBeforeCustomFiltering = _pruneTexturesSceneIndex =
-    Fvp::PruneTexturesSceneIndex::New(_lastFilteringSceneIndexBeforeCustomFiltering);
+        Fvp::PruneTexturesSceneIndex::New(_lastFilteringSceneIndexBeforeCustomFiltering, !isTextured);
```
ppt-adsk: On creation pass in the correct initial state, rather than always prune textures as false.

test/testSamples/testTexturedMode/texturedUsdSphere.ma
ppt-adsk, line 1: Will sanitize this file in a second pass.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lanierd-adsk: For clarity, could you please add a constexpr bool pruneTextures = !isTextured;

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lanierd-adsk: Same here, please

test/lib/mayaUsd/render/mayaToHydra/TexturedModeTest/textured.png
lanierd-adsk, line 1: Could you please place the camera closer to the sphere so the pixels are meaningful ?
Also if more light can be added that would be better if it's not the default lighting.

lanierd-adsk (changes requested):
Thank you for doing these change sonly a few changes for more clarity

ppt-adsk (commented):
Comments on how to make more robust are welcome!

lanierd-adsk (dismissed):

lanierd-adsk (approved):

PR #352, opened 2025-11-27
HYDRA-1993 : Fix Maya lights crash when using HdArnold

debloip-adsk:
Why the crash happened : 
1. HdStorm supports shadows only for simple and distant lights. 
2. In our Maya lights adapters, we compute shadow data based on whether the selected renderer is Storm or not. If it's Storm, we don't generate shadowing data, but we do for all other renderers.
3. Our secondary pass is always using Storm.

So if we select Arnold, then shadow data is generated, and the secondary pass with Storm receives this shadow data but does not support it : crash

This fix adds the option in pass filtering to remove lights from a pass, which we do in the secondary pass, whilst keeping only the Maya default light (which is a simple light, so no issues there)


lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -1917,4 +1915,10 @@ GfBBox3d MayaHydraSceneIndex::GetBoundingBox() const
     return bbox;
 }
 
+const SdfPath& MayaHydraSceneIndex::MayaDefaultLightPath()
```
debloip-adsk, line 1918: We need to access this before the first MayaHydraSceneIndex is created, and this is essentially a static constant anyways, so I extracted the data field into a static variable inside the static method, so we don't need to have created a MayaHydraSceneIndex to use the value.

test/testSamples/testPassFiltering/passFilteringScene.ma
debloip-adsk, line 1: This adds a Maya area light and a USD distant light

lanierd-adsk (approved):

PR #359, opened 2025-12-15
HYDRA-2000 : Split MayaHydraSceneIndex viewport features into separate scene index

debloip-adsk:
First stab at breaking down the mayaHydraSceneIndex monolith.

Main architectural changes : 
- MayaHydraSceneIndex's target responsibility is now strictly to translate Maya data to Hydra data. IMO this class could and should be renamed to something like MayaDataSceneIndex, but I didn't do it yet because that would pollute the PR's changes. I believe it could be broken down even further somewhere down the line, see next point.
- A new scene index, MayaViewportSceneIndex, was created to contain viewport-related code that previously lived in MayaHydraSceneIndex. Since handling these features still requires some access to the adapter classes stored by MayaHydraSceneIndex, it takes in a reference to it (but it is not necessarily its input scene index in the scene index chain). I think at some point we could separate the storage of the adapters from the MayaHydraSceneIndex so that MayaViewportSceneIndex does not depend on it.

Features handled by the MayaViewportSceneIndex
  - Picking
  - Maya faces selection highlighting
  - Default material
  - Default light
  - Active viewport lights
  - Xray
  - Some playback checks

Other changes :
- Dead code removal
- Maya faces selection behaviour is slightly different and now more closely matches VP2 (though it should have worked before as well)


lib/mayaHydra/hydraExtensions/adapters/renderItemAdapter.cpp
```diff
@@ -71,11 +71,15 @@ MayaHydraRenderItemAdapter::MayaHydraRenderItemAdapter(
...
-    _InsertRprim(this);
 }
 
 MayaHydraRenderItemAdapter::~MayaHydraRenderItemAdapter() { _RemoveRprim(); }
 
+void MayaHydraRenderItemAdapter::Populate()
+{
+    _InsertRprim(this);
+}
+
```
debloip-adsk, line 82: Needed to extract this for a fix in mayaHydraSceneIndex related to HYDRA-1992

lib/mayaHydra/hydraExtensions/adapters/renderItemAdapter.cpp
```diff
@@ -553,22 +557,22 @@ bool MayaHydraRenderItemAdapter::GetVisible()
 {
     // Assuming that, if the playback is in the active view only
     // (MAnimControl::kPlaybackViewActive), we are called because we are in the active view
-    if (_isHideOnPlayback) {
-        // MAYA-127216: Remove dependency on parent class MayaHydraAdapter. This will let us use
-        // MayaHydraSceneDelegate directly
-        auto mayaHydraSceneIndex = static_cast<MayaHydraSceneIndex*>(GetMayaHydraSceneIndex());
-        return !mayaHydraSceneIndex->GetPlaybackRunning();
```
debloip-adsk, line 560: Store playback state instead of redirecting to mayaHydraSceneIndex. The playback state will be updated by mayaViewportSceneIndex

lib/mayaHydra/hydraExtensions/adapters/shapeAdapter.cpp
```diff
@@ -131,24 +131,4 @@ const GfRange3d& MayaHydraShapeAdapter::GetExtent()
     return _extent;
 }
 
-void MayaHydraShapeAdapter::PopulateSelectedPaths(
```
debloip-adsk, line 120: Dead code

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -462,7 +389,7 @@ class MayaPathMapper : public Fvp::PathMapper
 
 } // namespace
 
-MayaHydraSceneIndex::MayaHydraSceneIndex(MayaHydraInitData& initData, bool lightEnabled)
```
debloip-adsk, line 465: lightEnabled was unused

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -542,30 +436,10 @@ void MayaHydraSceneIndex::_Destroy()
...
-    // Remove our pick handler from the pick handler registry.
-    if (_unregisterPickHandler) {
-        TF_AXIOM(MayaHydra::PickHandlerRegistry::Instance().Unregister(_rprimPath));
-    }
 }
 
-void MayaHydraSceneIndex::HandleCompleteViewportScene(
-    const MDataServerOperation::MViewportScene& scene,
-    MFrameContext::DisplayStyle                 ds)
+void MayaHydraSceneIndex::UpdateRenderItems(const MDataServerOperation::MViewportScene& scene)
```
debloip-adsk, line 441: Only used by viewport, but still here as it relates to core scene data translation. Render item translation also relies on the general structure handled by the MayaHydraSceneIndex (e.g. the whole rprims/sprims paths, materials creation, etc.) so keeping it here is less of a hassle for now.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -693,63 +571,8 @@ void MayaHydraSceneIndex::Populate()
...
-    if (lightsParamsWeUseAreDifferent) {
-        // Update our light
-        _mayaDefaultLight.SetDiffuse(light.GetDiffuse());
-        _mayaDefaultLight.SetSpecular(light.GetSpecular());
-        _mayaDefaultLight.SetPosition(light.GetPosition());
-        MarkSprimDirty(MayaDefaultLightPath(), HdLight::DirtyParams);
-    }
-}
-
-void modifyDefaultMaterialOpacity(HdMaterialNetworkMap& materialNetworkMap, bool xrayEnabled)
```
debloip-adsk, line 729: Dead code

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -961,16 +683,10 @@ MayaHydraSceneIndex::LightDagPathMap MayaHydraSceneIndex::_GetGlobalLightPaths()
     return allLightPaths;
 }
 
-void MayaHydraSceneIndex::PreFrame(const MHWRender::MDrawContext& context)
+void MayaHydraSceneIndex::FlushPendingUpdates()
```
debloip-adsk, line 686: MayaHydraSceneIndex listens to Maya scene changes and stores changed objects in maps such as _lightsToAdd, _adaptersToRecreate, instead of propagating them in real-time (I assume for performance reasons). This method is what actually does the update and sends the data to Hydra, so named it "FlushPendingUpdates".

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -1096,74 +812,8 @@ void MayaHydraSceneIndex::PreFrame(const MHWRender::MDrawContext& context)
...
-            },
-            _lightAdapters);
-
-        _lightsManagementSceneIndex->SetDisabledLightsPrims(disabledLights);
-    }
 }
 
-bool MayaHydraSceneIndex::GetPlaybackRunning() const { return _isPlaybackRunning; }
-
-void MayaHydraSceneIndex::PostFrame() { }
```
debloip-adsk, line 1165: Has been empty forever, so just removing it

lib/mayaHydra/hydraExtensions/sceneIndex/mayaViewportSceneIndex.cpp
```diff
@@ -0,0 +1,473 @@
...
+
+    // Setup the combined scene
+    _mergingSceneIndex->AddInputScene(_inputSceneIndex, SdfPath::AbsoluteRootPath());
+    _mergingSceneIndex->AddInputScene(_viewportDataSceneIndex, SdfPath::AbsoluteRootPath());
+    _mergingSceneIndex->AddObserver(HdSceneIndexObserverPtr(&_observer));
+}
+
+MayaViewportSceneIndex::~MayaViewportSceneIndex()
+{
+    Destroy();
```
debloip-adsk, line 188: See HYDRA-2019, I added the ticket as comment in the header file and where this is called in renderOverride as well. I suspect this issue is going to be a bit of a mess to fix.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaViewportSceneIndex.cpp
```diff
@@ -0,0 +1,473 @@
...
+        auto renderItemAdapters = _mayaDataSceneIndex->GetAdapterMap<MayaHydraRenderItemAdapter>();
+        for (const auto& [id, adapter] : renderItemAdapters) {
+            adapter->SetPlaybackState(_isPlaybackRunning);
+        }
+    }
+
+    _UpdateActiveLights(viewportDrawContext);
+}
+
+void MayaViewportSceneIndex::_UpdateActiveLights(const MDrawContext& viewportDrawContext)
```
debloip-adsk, line 306: I ported this code as-is for now, but at first glance it seems a bit overly complex? I'm not sure why we remove the light prims in one case and just disable them through the _lightsManagementSceneIndex in the other.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1769,14 +1769,18 @@ void MtohRenderOverride::_InitHydraResources(
 
     //Put BlockPrimRemovalPropagationSceneIndex first as it can block/unblock the prim removal propagation on the whole scene indices chain
     _blockPrimRemovalPropagationSceneIndex = Fvp::BlockPrimRemovalPropagationSceneIndex::New(_inputSceneIndexOfFilteringSceneIndicesChain);
+    _inputSceneIndexOfFilteringSceneIndicesChain = _blockPrimRemovalPropagationSceneIndex;
+
+    _mayaViewportSceneIndex = MayaViewportSceneIndex::New(_inputSceneIndexOfFilteringSceneIndicesChain, _mayaHydraSceneIndex);
```
debloip-adsk, line 1777: Added it right after _blockPrimRemovalPropagationSceneIndex so it's very early on in the chain, to keep it close to the Maya data it is primarily designed towards. Its placement shouldn't matter too much though

test/lib/mayaUsd/render/mayaToHydra/MayaComponentsPickingTest/smoothwireframe.png
debloip-adsk, line 1: Previously the faces selection material was only applied on PolyActiveFaces and not PolyActiveFacesOnWire, despite the code being structured as if it should (see https://github.com/Autodesk/maya-hydra/blob/4436d4f9eddda43088f104fb65580494b98f9d7c/lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp#L1477-L1484). 

And as it turns out, with the rework, it applies it to both and actually gives a result that is closer to VP2, so I just updated the reference image.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
ppt-adsk: Some adaptation of that would make a great code comment.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaViewportSceneIndex.h
```diff
@@ -0,0 +1,229 @@
...
+#include <mayaHydraLib/api.h>
+#include <mayaHydraLib/pick/mhPickHitFwd.h>
+#include <mayaHydraLib/sceneIndex/mayaHydraSceneIndex.h>
+
+#include <pxr/imaging/hd/filteringSceneIndex.h>
+#include <pxr/imaging/hd/mergingSceneIndex.h>
+#include <pxr/imaging/hd/retainedSceneIndex.h>
+
+#include <maya/MDrawContext.h>
+#include <maya/MSelectionList.h>
```
ppt-adsk: Should use MApiNamespace.h to forward declare.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaViewportSceneIndex.cpp
ppt-adsk: Can you write a sanitized version of this as a code comment?

ppt-adsk (dismissed):
Fantastic work, thanks for doing this!

ppt-adsk (dismissed):

debloip-adsk:
There were a few conflicts with the latest merge into dev

ppt-adsk (approved):

PR #362, opened 2025-12-17
Allow passing image writer object to image write method.

ppt-adsk:
To be used in proof of concept work.

### DIFF
--- a/lib/flowViewport/imageWriter/fvpImageBufferWriter.cpp
+++ b/lib/flowViewport/imageWriter/fvpImageBufferWriter.cpp
@@ -68,6 +68,15 @@ bool ImageBufferWriter::Write(
 
 /* static */
 bool ImageBufferWriter::Write(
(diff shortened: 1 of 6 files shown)


lib/flowViewport/imageWriter/fvpImageBufferWriter.cpp
```diff
@@ -66,6 +66,15 @@ bool ImageBufferWriter::Write(
     return (writer ? writer->Write(fileName) : false);
 }
 
+/* static */
+bool ImageBufferWriter::Write(
```
ppt-adsk, line 70: The ImageBufferWriter::Create() factory method is not flexible enough to support all the modes required by the proof of concept work (e.g. with and without frame pass).  By exposing a method that allows directly passing in the writer object, it can be created outside the image writer framework to adapt to requirements.

lanierd-adsk (dismissed):

lanierd-adsk (approved):

PR #361, opened 2025-12-17
HYDRA-2013 - fix build errors for usd 25.11

lilike-adsk:
Fix build errors for usd 25.11

PR #363, opened 2025-12-18
Fix compiler error when building against usd v25.11

hamedsabri:
Fix additional compiler error when building against usd v25.11 after https://github.com/Autodesk/maya-hydra/pull/361


lilike-adsk:
Good catch for Non-HVT build!

lilike-adsk (approved):

PR #368, opened 2026-01-08
HYDRA-2005 : Fix depth peeling

debloip-adsk:
Fix depth peeling artifacts.

I tried to add a unit test, but I could not get it to repro in our test environment, even if I tried removing some of our test-specific options. I did find out that the artifacts are only present when MSAA is off, but that still doesn't make the bug work in testing. There is likely some other element needed to repro, possibly an idle event, that is needed for the issue to occur.

Nonetheless, I've tested multiple times and with and without the change and it always fixed the issue when present, and the issue was back without it.

### DIFF
--- a/lib/mayaHydra/mayaPlugin/renderOverrideUtils.h
+++ b/lib/mayaHydra/mayaPlugin/renderOverrideUtils.h
@@ -18,6 +18,8 @@
 #ifndef MTOH_VIEW_OVERRIDE_UTILS_H
 #define MTOH_VIEW_OVERRIDE_UTILS_H
 
+#include <maya/MViewport2Renderer.h>
+
 #include <pxr/pxr.h>
 
 PXR_NAMESPACE_OPEN_SCOPE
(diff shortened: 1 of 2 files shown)


lib/mayaHydra/mayaPlugin/renderOverrideUtils.h
```diff
@@ -35,9 +37,15 @@ class MayaHydraPreRender : public MHWRender::MSceneRender
...
+        /// To skip the generation of some unwanted render lists even if the kRenderPreSceneUIItems
         /// filter is specified.
-        return MFrameContext::kExcludeManipulators | MFrameContext::kExcludeHUD;
+        static MUint64 sObjectTypeExclusions = []() -> MUint64 {
+            /// Exclude everything as a baseline. Restore grid.
+            MUint64 flags = MFrameContext::kExcludeAll;
+            flags &= ~MFrameContext::kExcludeGrid;
+            return flags;
+        }();
+        return sObjectTypeExclusions;
```
debloip-adsk, line 61: Same logic as there is in MayaHydraPostRender. This way we ensure only the grid is rendered, which is the only thing we use currently. It seems like the artfiacts come from some other thing trying to be rendered incorrectly and interfering with the depth peeling.

debloip-adsk:
Deprecated in favor of #370

PR #370, opened 2026-01-14
HYDRA-2005 : Fix depth peeling with USD data

debloip-adsk:
Using depth peeling in MayaHydra was causing z-fighting artifacts with USD data due to it being "rendered" twice, once by MRenderOperations and once by MayaHydra's specific USD handling. To fix this, we specifically exclude MayaUsd from MRenderOperations, so they are only handled by MayaHydra's special USD handling.

I could not get the issue to repro in our test environment, even if I tried removing some of our test-specific options. I did find out that the artifacts are only present when MSAA is off, but that still doesn't make the bug work in testing. There is likely some other element needed to repro, possibly an idle event, that is needed for the issue to occur.

Depends on the following PR in maya-usd : https://github.com/Autodesk/maya-usd/pull/4447

### DIFF
--- a/lib/mayaHydra/mayaPlugin/renderOverrideUtils.h
+++ b/lib/mayaHydra/mayaPlugin/renderOverrideUtils.h
@@ -61,10 +61,12 @@ public:
         return MFrameContext::kExcludeManipulators | MFrameContext::kExcludeHUD;
     }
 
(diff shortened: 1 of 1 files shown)


ppt-adsk (approved):
Thanks for the nice documentation.

PR #373, opened 2026-01-20
HYDRA-1706 : Update selection docs

debloip-adsk:
Updated existing images so they are up-to-date with HVT and FramePasses, and added a short section on Maya data highlighting


ppt-adsk (approved):
Thanks for doing this.

PR #376, opened 2026-01-21
HYDRA-2047 : Remove non-HVT code path

debloip-adsk:
- Remove all non-HVT code paths (which includes the entire SelectionTask, as it was only used in the non-HVT path)
- Remove HVT flags from CMake setup
- Remove a dead function (_NeedToRecreateTheSceneIndicesChain) and its dependent code


lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1204,61 +1076,29 @@ MStatus MtohRenderOverride::Render(
...
+        auto& isolateSelectMgr = Fvp::IsolateSelectManager::Get();
+        auto isSi = isolateSelectMgr.GetIsolateSelectSceneIndex();
+        auto isolateSelection = isolateSelectMgr.GetOrCreateIsolateSelection(panelNameStr);
+        if (isSi && (isSi->GetIsolateSelection() != isolateSelection)) {
             TF_DEBUG(MAYAHYDRALIB_RENDEROVERRIDE_SCENE_INDEX_CHAIN_MGMT)
-                .Msg("Re-creating scene index chain to render %s\n", panelNameStr.c_str());
-            _CreateSceneIndicesChainAfterMergingSceneIndex(drawContext);
-            
-            const Fvp::InformationInterface::RenderViewDesc hydraViewportInformation(panelNameStr, true);
-            manager.AddRenderViewData(
```
lanierd-adsk, line 1221: I don't see this function call any longer. Is that on purpose ?

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1204,61 +1076,29 @@ MStatus MtohRenderOverride::Render(
     if (_mayaHydraSceneIndex) {
         _mayaHydraSceneIndex->SetParams(delegateParams);
         _mayaHydraSceneIndex->FlushPendingUpdates();
+    }
 
-        auto& manager = Fvp::RenderViewDataManager::Get();
-        if (_NeedToRecreateTheSceneIndicesChain(currentDisplayStyle)){
```
lanierd-adsk, line 1209: Was this unused ?

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1204,61 +1076,29 @@ MStatus MtohRenderOverride::Render(
...
+            // Isolate select scene index is being switched to a different
+            // viewport, set its isolate selection.
+            isSi->SetViewport(panelNameStr, isolateSelection);
         }
         else {
+            // This case includes disabled (null pointer) isolate selection.
             TF_DEBUG(MAYAHYDRALIB_RENDEROVERRIDE_SCENE_INDEX_CHAIN_MGMT)
-                .Msg("Re-using existing scene index chain to render %s\n", panelNameStr.c_str());
-
-#ifdef MAYA_HAS_VIEW_SELECTED_OBJECT_API
```
lanierd-adsk, line 1239: Is this no longer necessary ?

lanierd-adsk (commented):
Thanks for doing all this cleanup. Only a few question to be sure what is being removed is no longer necessary.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
debloip-adsk: Yes, if you look at this function's implementation, it always returned false and had a comment saying to remove it at some point

lib/mayaHydra/mayaPlugin/renderOverride.cpp
debloip-adsk: Yes, since _NeedToRecreateTheSceneIndicesChain always returned false we could never get into this branch path

lib/mayaHydra/mayaPlugin/renderOverride.cpp
debloip-adsk: It's still there, just moved up a bit due to the removed code, see line 1081 in the updated file

lanierd-adsk (approved):

PR #382, opened 2026-02-17
HYDRA-2021 : Support most of the custom maya attributes types when tr…

lanierd-adsk:
…anslating to hydra

### DIFF
--- a/lib/mayaHydra/hydraExtensions/adapters/adapter.cpp
+++ b/lib/mayaHydra/hydraExtensions/adapters/adapter.cpp
@@ -155,7 +155,7 @@ void MayaHydraAdapter::HandleExtensionAttributesDirty(const MPlug& plug)
     MObject attrObj = plug.attribute(&status);
     if (status) {
         MFnAttribute fnAttr(attrObj);
-        if (fnAttr.isExtension()) {
+        if (fnAttr.isExtension() || fnAttr.isDynamic()) {
             _extAttrMapNeedUpdate = true;
             // Notify the change tracker that the primvars have changed.
             // Note there's no fine grained dirty notification mechanism on primvars yet,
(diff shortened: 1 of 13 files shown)


lib/mayaHydra/hydraExtensions/adapters/adapter.cpp
```diff
@@ -155,7 +155,7 @@ void MayaHydraAdapter::HandleExtensionAttributesDirty(const MPlug& plug)
     MObject attrObj = plug.attribute(&status);
     if (status) {
         MFnAttribute fnAttr(attrObj);
-        if (fnAttr.isExtension()) {
+        if (fnAttr.isExtension() || fnAttr.isDynamic()) {
```
lilike-adsk, line 158: Do we need to deal with dynamic attribute?

lib/mayaHydra/hydraExtensions/adapters/cameraAdapter.cpp
```diff
@@ -79,7 +80,10 @@ void MayaHydraCameraAdapter::Populate()
 void MayaHydraCameraAdapter::MarkDirty(HdDirtyBits dirtyBits)
 {
     if (_isPopulated && dirtyBits != 0) {
-        dirtyBits = dirtyBits & HdCamera::AllDirty;
+        // We support extension-attribute primvars on cameras, so keep DirtyPrimvar even though
+        // cameras are sprims and normally only expose HdCamera dirty bits.
+        const HdDirtyBits primvarBits = dirtyBits & HdChangeTracker::DirtyPrimvar;
+        dirtyBits = (dirtyBits & HdCamera::AllDirty) | primvarBits;
```
lilike-adsk, line 86: Nice catch

lib/mayaHydra/hydraExtensions/adapters/adapter.cpp
lanierd-adsk: Yes—if “custom attributes” means user‑defined attributes (added per‑node with addAttr), those are dynamic, not extension. In that case you should handle isDynamic() as well, otherwise they’ll be ignored. If you only care about plugin‑defined extension attributes on built‑in node types, then extension‑only is fine.

lanierd-adsk:
Sorry, I had problems with git and had to create a new PR which is :
https://github.com/Autodesk/maya-hydra/pull/383 I am closing this PR.

PR #383, opened 2026-02-20
Support more custom attributes from maya in hydra translation

lanierd-adsk:
This is the following from PR https://github.com/Autodesk/maya-hydra/pull/382 
I had git problems and had to create a new branch.

### DIFF
--- a/lib/mayaHydra/hydraExtensions/adapters/adapter.cpp
+++ b/lib/mayaHydra/hydraExtensions/adapters/adapter.cpp
@@ -155,7 +155,7 @@ void MayaHydraAdapter::HandleExtensionAttributesDirty(const MPlug& plug)
     MObject attrObj = plug.attribute(&status);
     if (status) {
         MFnAttribute fnAttr(attrObj);
-        if (fnAttr.isExtension()) {
+        if (fnAttr.isExtension() || fnAttr.isDynamic()) {
             _extAttrMapNeedUpdate = true;
             // Notify the change tracker that the primvars have changed.
             // Note there's no fine grained dirty notification mechanism on primvars yet,
--- a/lib/mayaHydra/hydraExtensions/adapters/adapter.cpp
+++ b/lib/mayaHydra/hydraExtensions/adapters/adapter.cpp
@@ -149,7 +149,7 @@ HdPrimvarDescriptorVector MayaHydraAdapter::GetPrimvarDescriptors(HdInterpolatio
     return HdPrimvarDescriptorVector();
 }
 
-void MayaHydraAdapter::HandleExtensionAttributesDirty(const MPlug& plug)
+void MayaHydraAdapter::HandleExtensionAndDynamicAttributesDirty(const MPlug& plug)
 {
     MStatus status;
     MObject attrObj = plug.attribute(&status);
--- a/lib/mayaHydra/hydraExtensions/adapters/adapter.cpp
+++ b/lib/mayaHydra/hydraExtensions/adapters/adapter.cpp
@@ -136,7 +136,9 @@ HdPrimvarDescriptorVector MayaHydraAdapter::GetPrimvarDescriptors(HdInterpolatio
         if (_extAttrMapNeedUpdate) {
             // Apply a global lock to avoid race condition while doing parallel DG node evaluation.
             std::lock_guard<LockType> lock(dg_access_mutex);
-            MAYAHYDRA_NS::GetExtensionAttributesFromNode(GetNode(), _extAttrNameToValueMap);
+            MAYAHYDRA_NS::GetExtensionAndDynamicAttributesFromNode(
+                GetNode(),
+                _extAttrNameToValueMap);
             _extAttrMapNeedUpdate = false;
         }
         // Use constant interpolation and none role for all primvars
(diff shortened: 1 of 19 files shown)


lib/mayaHydra/hydraExtensions/adapters/adapter.cpp
```diff
@@ -155,7 +155,7 @@ void MayaHydraAdapter::HandleExtensionAttributesDirty(const MPlug& plug)
     MObject attrObj = plug.attribute(&status);
     if (status) {
         MFnAttribute fnAttr(attrObj);
-        if (fnAttr.isExtension()) {
+        if (fnAttr.isExtension() || fnAttr.isDynamic()) {
```
lanierd-adsk, line 162: @lilike-adsk : Dynamic is needed : if “custom attributes” means user‑defined attributes (added per‑node with addAttr), those are dynamic, not extension. In that case you should handle isDynamic() as well, otherwise they’ll be ignored. If you only care about plugin‑defined extension attributes on built‑in node types, then extension‑only is fine.

lilike-adsk: It's was skipped as Arnold only used extension attr (per-class), we can keep it and it might be useful for other plugins

lib/mayaHydra/hydraExtensions/mixedUtils.cpp
```diff
@@ -292,6 +1325,131 @@ void GetExtensionAttributesFromNode(
                 }
                 break;
             } break;
+            case MFn::kCompoundAttribute: {
```
lilike-adsk, line 1332: Nice, there's another type we might need to support - kMessageAttribute (e.g., for aovInput, aiFilters )? We can log it as another ticket if it's too complicated

lilike-adsk (dismissed):
Great work!

lib/mayaHydra/hydraExtensions/adapters/adapter.cpp
ppt-adsk: Why were we handling only extension attribute dirtiness in the past?  Is it because normal attribute dirtiness is handled elsewhere?  Does this change mean that we were not handling dynamic attribute dirtiness in the past?  The method name says HandleExtensionAttributesDirty

lib/mayaHydra/hydraExtensions/mixedUtils.cpp
ppt-adsk: Why do we need to support message attributes?  Message attributes carry no data.  They indicate a relationship between one Maya node (the message source) and one or more destination nodes.  How would we translate these attributes?  Just have a list of SdfPath's that point to the destination(s)?  To me this feels like something to discuss with renderer teams (e.g. Arnold and RenderMan), because I fear they are encoding things that require processing in the Maya to Hydra data translation code.  I hope I'm wrong and that the processing can be done in a scene index filter.

lib/mayaHydra/hydraExtensions/adapters/adapter.cpp
lilike-adsk: The dirtiness of most other normal attributes were handled in a different way (visibility/xform/topology change).
HandleExtensionAttributesDirty could be renamed as HandleExtensionAndDynamicAttributesDirty

ppt-adsk: So do I understand things correctly that we were not handling dynamic attribute dirtiness in the past?  Why has this not been a problem up to now?

lib/mayaHydra/hydraExtensions/mixedUtils.cpp
lanierd-adsk: ok, we'll discuss about that. I won't add them now.

lib/mayaHydra/hydraExtensions/adapters/adapter.cpp
lilike-adsk: We didn't translate extension/dynamic attributes and hanlde dirtiness of them at all in the past. We added the translation for extension attributes along with the handling of dirtness of extension attribute for Prod. rendering.  David just added the translation of dynamic attribute in this PR, so no problem in the past

lib/mayaHydra/hydraExtensions/mixedUtils.cpp
```diff
@@ -245,6 +1042,7 @@ PXR_NS::TfToken GetGeomSubsetsPickMode()
     return GeomSubsetsPickModeTokens->None;
 }
 
+// Populate a dictionary with extension/dynamic attribute values.
 void GetExtensionAttributesFromNode(
```
lilike-adsk: Please also rename this.

ppt-adsk (dismissed):

lib/mayaHydra/hydraExtensions/mixedUtils.cpp
lilike-adsk: @lanierd-adsk ^^ rename as GetExtensionAndDynamicAttributesFromNode?

lilike-adsk (approved):

ppt-adsk (approved):

PR #384, opened 2026-02-20
HYDRA-2113 : Fix tumble performance drop

debloip-adsk:
Add a simple check to avoid redundant processing.


lib/flowViewport/sceneIndex/fvpPurposeFilteringSceneIndex.cpp
debloip-adsk, line 192: This method is called every frame and not necessarily only when the included purposes change, which was causing the performance slowdown. We can know from within the scene index if the purposes changed, so add a defensive check here

ppt-adsk: Ouch.  Thanks for the fix.

ppt-adsk (approved):

PR #386, opened 2026-02-23
HYDRA-1911 : Update threshold for OSX

debloip-adsk:


test/lib/mayaUsd/render/mayaToHydra/testUSDLights.py
```diff
@@ -26,7 +26,7 @@ class TestUSDLights(mtohUtils.MayaHydraBaseTestCase): #Subclassing mtohUtils.May
     _file = __file__
 
     IMAGE_DIFF_FAIL_THRESHOLD = 0.01
-    IMAGE_DIFF_FAIL_PERCENT = 0.2
+    IMAGE_DIFF_FAIL_PERCENT = 0.5
```
debloip-adsk, line 29: Didn't add an OSX-specific threshold as it's still quite small and sensitive. While this test is still disabled for now, I ran a PR with it enabled and using this threshold and it passed.

lanierd-adsk (approved):

PR #389, opened 2026-02-27
HYDRA-2100 : Fix unsupported primitive in testViewportFilters

debloip-adsk:


lib/mayaHydra/hydraExtensions/adapters/renderItemAdapter.cpp
```diff
@@ -321,7 +321,9 @@ void MayaHydraRenderItemAdapter::UpdateFromDelta(const UpdateFromDeltaData& data
     }
 
     // Indices
-    if (topoChanged && vertexBuffercount) {
+    // Line strips do not make use of the index buffer, so we can skip this block.
+    // See "Line strips indices are implicitly defined" comment.
+    if (topoChanged && vertexBuffercount && GetPrimitive() != MHWRender::MGeometry::Primitive::kLineStrip) {
```
debloip-adsk, line 326: "Line strips indices are implicitly defined" comment is [this comment](https://github.com/Autodesk/maya-hydra/blob/633338d4805856164da51f10cb3de619f5a7febe/lib/mayaHydra/hydraExtensions/adapters/renderItemAdapter.cpp#L448-L457).

Line strips were not handled in the switch case in this block because they force values into the vertexCounts and vertexIndices buffers anyways, but they were not skipped either, so they fell into the default case which assumed an unknown primitive. However in this case it was intentional to not do anything, so just explicitly skip it.

We ran into this case due to the image plane selection highlights, and while image planes are not supported yet, this problem could happen with anything else using line strips.

ppt-adsk (approved):

PR #390, opened 2026-03-04
HYDRA-2146 : Fix extra notifications sent when changing a custom attribute value on cameras and lights

lanierd-adsk:
Fix extra notifications sent when changing a custom attribute value on cameras and lights

### DIFF
--- a/lib/mayaHydra/hydraExtensions/adapters/lightAdapter.cpp
+++ b/lib/mayaHydra/hydraExtensions/adapters/lightAdapter.cpp
@@ -111,16 +111,6 @@ void _dirtyTransform(MObject& node, void* clientData)
     }
 }
 
-void _dirtyParams(MObject& node, void* clientData)
-{
-    TF_UNUSED(node);
-    auto* adapter = reinterpret_cast<MayaHydraDagAdapter*>(clientData);
-    if (adapter->IsVisible()) {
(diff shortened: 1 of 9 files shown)


lib/mayaHydra/hydraExtensions/adapters/lightAdapter.cpp
```diff
@@ -772,7 +782,9 @@ void MayaHydraLightAdapter::CreateCallbacks()
     MStatus status;
     auto    dag = GetDagPath();
     auto    obj = dag.node();
-    auto    id = MNodeMessage::addNodeDirtyCallback(obj, _dirtyParams, this, &status);
+    // Use attribute-changed on shape (not node-dirty) to avoid redundant light+primvars dirtying
+    // when only custom attributes change.
+    auto id = MNodeMessage::addAttributeChangedCallback(obj, _lightShapeAttributeChanged, this, &status);
```
lilike-adsk: Should we also do the similar change in aiAreaLightAdapter.cpp?

lanierd-adsk: Good catch, yes and in the material adapter as well.

lilike-adsk (dismissed):

lilike-adsk (approved):

PR #391, opened 2026-03-05
Bifrost update test now conditional to USD versions above 0.24.x

ppt-adsk:


test/lib/mayaUsd/render/mayaToHydra/testBifrostGraphUpdate.py
```diff
@@ -31,6 +35,7 @@ def verifySnapshot(self, imageName):
                                  self.IMAGEDIFF_FAIL_THRESHOLD,
                                  self.IMAGEDIFF_FAIL_PERCENT)
 
+    @unittest.skipIf(Usd.GetVersion()[1] == 24, "Bifrost errors with OpenUSD 0.24.11.")
```
lanierd-adsk, line 38: We already have the usd version in the class, so please use something like 
 @unittest.skipIf(self._usdVersion[1]  <= 24, "Bifrost errors with OpenUSD 0.24.11.")

lanierd-adsk (changes requested):

test/lib/mayaUsd/render/mayaToHydra/testBifrostGraphUpdate.py
ppt-adsk: Unfortunately self is not available in the decorator, as it is run when the class itself is created, at which point there is no instance of the class.

lanierd-adsk (approved):

PR #395, opened 2026-03-09
HYDRA-2140 : Add MayaHydra profiling utilities

debloip-adsk:
Adds some basic profiling utilities to bridge the gap between Maya and USD profiling.

- mayaHydraProfiling command to start/stop both Maya and USD profiling simultaneously
- MH_PROFILING macros to trace MayaHydra code in both profilers

### DIFF
--- a/lib/mayaHydra/mayaPlugin/profilingCommand.cpp
+++ b/lib/mayaHydra/mayaPlugin/profilingCommand.cpp
@@ -1,0 +1,84 @@
+//
+// Copyright 2026 Autodesk
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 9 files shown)


lib/mayaHydra/mayaPlugin/profilingCommand.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include <mayaHydraLib/profilingUtils.h>
+
+#include <maya/MArgDatabase.h>
+#include <maya/MSyntax.h>
+
+/**
+ * mayaHydraProfiling command
+ * 
+ * This command allows the user to start/stop both the Maya profiler and USD tracing system at the same time.
```
lanierd-adsk, line 27: I would add more information about is the USD tracing system, maybe also add alink to the doc from it if possible.

lanierd-adsk (changes requested):
Thank you. Only a few comments to add I think.

ppt-adsk (commented):
Thanks for doing this, LGTM!

lanierd-adsk (approved):
Thanks !

PR #398, opened 2026-03-16
HYDRA-1760 : Support translation of maya extension attributes to hydra for lights, cameras and meshes

lanierd-adsk:


lib/mayaHydra/hydraExtensions/adapters/renderItemAdapter.cpp
```diff
@@ -328,7 +329,6 @@ void MayaHydraRenderItemAdapter::UpdateFromDelta(const UpdateFromDeltaData& data
         MIndexBuffer* indices = geom->indexBuffer(0);
         if (indices) {
             int indexCount = indices->size();
-            vertexIndices.resize(indexCount);
```
lanierd-adsk, line 331: Redundant with the line :
vertexIndices.assign(indicesData, indicesData + indexCount);

test/lib/mayaUsd/render/mayaToHydra/cpp/testUtils.cpp
```diff
@@ -391,12 +484,29 @@ bool dataSourceMatchesReference(
     referenceString.erase(
         std::remove(referenceString.begin(), referenceString.end(), '\r'), referenceString.end());
 
+    auto trimTrailingWhitespace = [](std::string& value) {
+        while (!value.empty()) {
+            const char ch = value.back();
+            if (ch != ' ' && ch != '\t' && ch != '\n' && ch != '\v' && ch != '\f') {
```
ppt-adsk: Should be std::isspace().

test/lib/mayaUsd/render/mayaToHydra/cpp/testUtils.cpp
```diff
@@ -391,12 +484,29 @@ bool dataSourceMatchesReference(
     referenceString.erase(
         std::remove(referenceString.begin(), referenceString.end(), '\r'), referenceString.end());
 
+    auto trimTrailingWhitespace = [](std::string& value) {
+        while (!value.empty()) {
```
ppt-adsk: Perhaps
```
auto it = std::find_if_not(
        value.rbegin(), value.rend(), [](unsigned char c) { return std::isspace(c); }
);

value.erase(it.base(), value.end());

File: test/lib/mayaUsd/render/mayaToHydra/cpp/testUtils.cpp [hunk 4]
```diff
@@ -315,6 +386,24 @@ Fvp::SelectionSceneIndexRefPtr findSelectionSceneIndexInTree(
     return TfDynamic_cast<Fvp::SelectionSceneIndexRefPtr>(selectionSiBase);
 }
 
+HdSceneIndexBaseRefPtr FindMayaHydraSceneIndex(
+    const HdSceneIndexBaseRefPtr& terminalSceneIndex)
+{
+    auto isDataProducerMerging = SceneIndexDisplayNamePred("Data Producer Merging Scene Index");
```
ppt-adsk, line 397: Why can't we just use findSceneIndexInTree() once, with the MayaHydraSceneIndex?

lib/mayaHydra/hydraExtensions/adapters/adapter.h
```diff
@@ -87,12 +90,60 @@ class MayaHydraAdapter
...
+    /// override: return ShouldMarkPrimvarDirtyForParamAttrs(plug, kXxxParamAttributeNames).
+    template <size_t N>
+    static bool ShouldMarkPrimvarDirtyForParamAttrs(const MPlug& plug, const char* const (&names)[N])
+    {
+        return !IsParamAttribute(plug, GetParamAttributeSet(names));
+    }
+
+    /// Return the top-level plug for a child/array element plug.
+    /// (e.g. aiLookAt[0].child(0) -> aiLookAt)
+    static MPlug GetTopPlug(const MPlug& plug);
```
ppt-adsk: This is not adapter-specific, right?  Should be in some kind of utility file as a namespaced function.

lib/mayaHydra/hydraExtensions/adapters/adapter.h
```diff
@@ -87,12 +90,60 @@ class MayaHydraAdapter
...
+
+    /// When true, all non-builtin attributes are translated to primvars (e.g. for lights).
+    /// When false, only extension and dynamic attributes. Default is false.
+    virtual bool IncludeAllAttributesInPrimvars() const { return false; }
```
ppt-adsk: I don't understand this.  To my understanding, there are only 3 types of attribute creation categorizations in Maya: built-in, dynamic, and extension.  I am therefore missing something, or this method is mis-named.

lib/mayaHydra/hydraExtensions/adapters/adapter.h
```diff
@@ -87,12 +90,60 @@ class MayaHydraAdapter
...
 
     /// Returns true if \p plug is an extension or dynamic attribute. Used by adapters
     /// to route extension/dynamic attribute changes to HandleExtensionAndDynamicAttributesDirty
     /// instead of marking schema params dirty.
     static bool IsExtensionOrDynamicAttribute(const MPlug& plug);
+    /// Mark primvars dirty for extension or dynamic attributes without touching schema params.
+    void HandleExtensionAndDynamicAttributesDirty(const MPlug& plug);
```
ppt-adsk: What is the responsibility of this method now that there are all the new methods?

lib/mayaHydra/hydraExtensions/adapters/adapter.h
```diff
@@ -87,12 +90,60 @@ class MayaHydraAdapter
...
+    }
+
+    /// Helper for adapters with param attribute lists. Use in ShouldMarkPrimvarDirtyForAttributeChange
+    /// override: return ShouldMarkPrimvarDirtyForParamAttrs(plug, kXxxParamAttributeNames).
+    template <size_t N>
+    static bool ShouldMarkPrimvarDirtyForParamAttrs(const MPlug& plug, const char* const (&names)[N])
+    {
+        return !IsParamAttribute(plug, GetParamAttributeSet(names));
+    }
+
```
ppt-adsk, line 169: These 3 static methods seem to be deeply related.  Do we need all 3, or could we condense them down to 1?

lib/mayaHydra/hydraExtensions/adapters/adapter.h
```diff
@@ -87,12 +90,60 @@ class MayaHydraAdapter
 
     bool IsPopulated() const { return _isPopulated; }
 
-    void HandleExtensionAndDynamicAttributesDirty(const MPlug& plug);
+    void MarkPrimvarDirtyForAttributeChange(const MPlug& plug);
```
ppt-adsk, line 117: What is the intended behavior of this method if the argument plug doesn't correspond to a primvar?

lib/mayaHydra/hydraExtensions/adapters/adapter.h
```diff
@@ -87,12 +90,60 @@ class MayaHydraAdapter
...
+
+    /// Override to add extra dirty bits when a primvar-affecting attr changes (e.g. light param
+    /// attrs need DirtyParams|DirtyShadowParams). Consolidating into one MarkDirty reduces
+    /// redundant scene index notifications.
+    virtual HdDirtyBits GetExtraDirtyBitsForPrimvarAttributeChange(const MPlug& plug) const
```
ppt-adsk: If the goal of this method is dirtiness consolidation, could we name it as such?

lib/mayaHydra/hydraExtensions/adapters/adapter.h
```diff
@@ -87,12 +90,60 @@ class MayaHydraAdapter
...
+
+    /// Call from attribute-changed callbacks when an attr change should mark primvars dirty.
+    /// Resolves \p plug to the appropriate parent/root plug before marking primvars dirty.
+    /// This can still be invoked for child plugs of compound/array attributes (e.g. color.r),
+    /// so callers may see multiple invocations that coalesce to the same top plug.
+    /// Override ShouldMarkPrimvarDirtyForAttributeChange to return false when another callback
+    /// already marks DirtyPrimvar (e.g. mesh _dirtyBits, light param attr list) to avoid
+    /// duplicate notifications.
+    void MaybeMarkPrimvarDirtyForAttributeChange(const MPlug& plug);
+    virtual bool ShouldMarkPrimvarDirtyForAttributeChange(const MPlug& plug) const { return true; }
```
ppt-adsk, line 143: What is the responsibility of MaybeMark versus ShouldMark?

lib/mayaHydra/hydraExtensions/adapters/adapter.h
ppt-adsk, line 1: I have to admit that I have lost sight of the responsibility of the various methods in this class.  I can see that there are attribute helpers, which should probably be moved, there are dirtiness methods, and there are dirtiness coalescing methods, but I'm not sure who does what.

lib/mayaHydra/hydraExtensions/adapters/adapter.cpp
```diff
@@ -165,19 +189,43 @@ bool MayaHydraAdapter::IsExtensionOrDynamicAttribute(const MPlug& plug)
...
+{
+    // Trace to top-level plug: when a compound/array element's child changes
+    // (e.g. aiLookAt[0].child(0)), Maya may only fire for the child. We must
+    // still mark primvars dirty so the scene browser refreshes when resetting
+    // to default (primvar removal). Duplicate MarkDirty calls are idempotent.
+    MPlug topPlug = GetTopPlug(plug);
+    if (ShouldMarkPrimvarDirtyForAttributeChange(topPlug)) {
+        MarkPrimvarDirtyForAttributeChange(topPlug);
+    }
+}
```
ppt-adsk, line 197: If this method was recursive, could it not call itself if the argument plug is not the top-level plug?

lib/mayaHydra/hydraExtensions/adapters/adapter.cpp
```diff
@@ -165,19 +189,43 @@ bool MayaHydraAdapter::IsExtensionOrDynamicAttribute(const MPlug& plug)
...
-            // Note there's no fine grained dirty notification mechanism on primvars yet,
-            // like dirty flags for adding/removing/changing a specific primvar, only
-            // DirtyPrimvar for all changes. Currently, no performance bottleneck was spotted
-            // around this. This could be improved if a more fine grained mechanism is provided.
-            MarkDirty(HdChangeTracker::DirtyPrimvar);
+            // Notify the change tracker. Include extra bits from subclass (e.g. light param attrs
+            // need DirtyParams|DirtyShadowParams) to consolidate into one MarkDirty and avoid
+            // redundant scene index notifications.
+            MarkDirty(
+                HdChangeTracker::DirtyPrimvar | GetExtraDirtyBitsForPrimvarAttributeChange(plug));
```
ppt-adsk: GetExtraDirtyBitsForPrimvarAttributeChange() is for derived class implementation, right?  Its name should reflect that this is for derived class use.

lib/mayaHydra/hydraExtensions/adapters/cameraAdapter.cpp
```diff
@@ -25,13 +25,79 @@
...
 #include <maya/MNodeMessage.h>
+#include <maya/MPlug.h>
 
 PXR_NAMESPACE_OPEN_SCOPE
 
 namespace {
 
+// Attributes that affect HdCamera params (from GetCameraParamValue). When these change,
+// we mark DirtyParams | DirtyPrimvar in one call to avoid duplicates.
+static const char* const kCameraParamAttributeNames[] = {
```
ppt-adsk, line 40: If we're going to search this, should it not be a set to avoid O(n) lookup?

lib/mayaHydra/hydraExtensions/mixedUtils.h
```diff
@@ -257,18 +257,21 @@ MAYAHYDRALIB_API
...
+ * @param[in] includeAllAttributes when true, process all non-builtin attributes; when false, only
+ *            extension and dynamic attributes.
  */
 MAYAHYDRALIB_API
-void GetExtensionAndDynamicAttributesFromNode(
+void GetAttributesFromNode(
     const MObject& node,
-    PXR_NS::VtDictionary& attrs);
+    PXR_NS::VtDictionary& attrs,
+    bool includeAllAttributes = false);
```
ppt-adsk: If includeAllAttributes is false, what is included?

ppt-adsk (commented):
Did not look at the tests, but we can start with this.

lib/mayaHydra/hydraExtensions/adapters/adapter.cpp
lanierd-adsk: GetTopPlug already walks up to the root in a loop, so recursion would just wrap that logic (at best one extra call). The current non‑recursive pattern is the simplest and safest.

lib/mayaHydra/hydraExtensions/adapters/adapter.cpp
lanierd-adsk: Yes, I am not sure about a new name, do you have any suggestions ? Cursor was proposing GetDerivedExtraDirtyBitsForPrimvarAttributeChange which I am not sure is a good idea...

lib/mayaHydra/hydraExtensions/adapters/adapter.h
lanierd-adsk: I've added some comments to clarify, thanks for suggesting this.

lib/mayaHydra/hydraExtensions/adapters/adapter.h
lanierd-adsk: I've added a comment :
// No-op if the plug is not an extension or dynamic attribute.

lib/mayaHydra/hydraExtensions/adapters/adapter.h
lanierd-adsk: I am adding comments to clarify.

lib/mayaHydra/hydraExtensions/adapters/adapter.h
```diff
@@ -87,12 +90,60 @@ class MayaHydraAdapter
...
+
+    /// Helper for adapters with param attribute lists. Returns true if plug's attribute is in the set.
+    /// Use: ShouldMarkPrimvarDirtyForAttributeChange returns !IsParamAttribute(plug, paramAttrs).
+    static bool IsParamAttribute(const MPlug& plug, const std::unordered_set<std::string>& paramAttrs);
+
+    /// Build a static param attribute set from an array. Each instantiation caches its own set.
+    template <size_t N>
+    static const std::unordered_set<std::string>& GetParamAttributeSet(const char* const (&names)[N])
+    {
+        static const std::unordered_set<std::string> s(names, names + N);
```
lanierd-adsk, line 157: BTW, Function‑local static initialization is thread‑safe in C++11 and later, and the returned set is const and never mutated. So no races after initialization. You asked me to check this.

lib/mayaHydra/hydraExtensions/adapters/adapter.h
lanierd-adsk: I've added a comment: 
 // These three helpers are intentionally layered:
    // - IsParamAttribute: low-level predicate usable in multiple contexts.
    // - GetParamAttributeSet: cached set per static name array (no rebuild per call).
    // - ShouldMarkPrimvarDirtyForParamAttrs: convenience wrapper for the common gating pattern.

But they should be kept separated

lib/mayaHydra/hydraExtensions/adapters/adapter.h
lanierd-adsk: I moved it to mayaUtils.

lib/mayaHydra/hydraExtensions/adapters/adapter.h
lanierd-adsk: You are right, it's a leftover from previous iterations... I am removing it.

lib/mayaHydra/hydraExtensions/adapters/adapter.h
lanierd-adsk: I removed that function now.

lib/mayaHydra/hydraExtensions/adapters/cameraAdapter.cpp
lanierd-adsk: We already do: the param lists are const char*[], but lookups go through GetParamAttributeSet(...), which returns an unordered_set and makes IsParamAttribute O(1).
So the linear list is just the source of truth; we don’t search it directly for lookups.

lib/mayaHydra/hydraExtensions/mixedUtils.h
lanierd-adsk: I removed this, it wasn't used. Thanks for pointing this out.

test/lib/mayaUsd/render/mayaToHydra/cpp/testUtils.cpp
lanierd-adsk: findSceneIndexInTree() could find a MayaHydraSceneIndex directly, but the current implementation intentionally scopes the search to the Data Producer Merging Scene Index first. That node is the known junction where the Maya‑authored scene index lives, and using it as the parent avoids accidentally picking a MayaHydraSceneIndex from another branch if multiple exist.

So the two‑step search is about selecting the correct branch deterministically, not about a limitation of findSceneIndexInTree() itself. If you want, we can simplify it to a single findSceneIndexInTree() call and note the assumption that there is only one MayaHydraSceneIndex in the tree.

lanierd-adsk: I checked renderOverride.cpp: each MtohRenderOverride instance creates one _mayaHydraSceneIndex via MayaHydraSceneIndex::New(...) and inserts it into the data‑producer merging scene index. So within a given terminal scene index tree (i.e., per render‑override/viewport), you should have a single MayaHydraSceneIndex.

That said, the app can still have multiple render overrides / viewports, which would each have their own scene index trees and their own MayaHydraSceneIndex. So “multiple” is possible at the application level, but not within one terminal tree. The current two‑step search is a safety measure to ensure we pick the correct branch in that tree.

ppt-adsk: Would be nice to add that as a code comment.

lib/mayaHydra/hydraExtensions/adapters/meshAdapterTestUtils.h
```diff
@@ -0,0 +1,33 @@
...
+#ifndef MAYAHYDRALIB_MESH_ADAPTER_TEST_UTILS_H
+#define MAYAHYDRALIB_MESH_ADAPTER_TEST_UTILS_H
+
+#include <mayaHydraLib/api.h>
+#include <pxr/pxr.h>
+
+#include <string>
+#include <unordered_set>
+
+PXR_NAMESPACE_OPEN_SCOPE
```
ppt-adsk: Sorry for the late catch.  Why is this in the PXR namespace?

lib/mayaHydra/hydraExtensions/mixedUtils.cpp
```diff
@@ -91,30 +92,22 @@ bool UseEnumLabelsForExtensionAndDynamicAttrs()
...
 
-// Update attribute dictionary with optional default comparison.
+// Store value in attrs. Default check is done via MPlug::isDefaultValue() before calling.
 void UpdateAttrsValue(
     const char* attrName,
     const VtValue& val,
-    const VtValue& defaultVal,
-    const bool hasDefault,
+    const VtValue& /* defaultVal */,
+    const bool /* hasDefault */,
```
ppt-adsk: If we don't need these any more, shouldn't we just remove them?

ppt-adsk (commented):
Just a couple more questions.

lib/mayaHydra/hydraExtensions/adapters/meshAdapterTestUtils.h
lanierd-adsk: It's a mistake, I moved it to the maya hydra namespace. Thanks  for catching this !

lib/mayaHydra/hydraExtensions/mixedUtils.cpp
lanierd-adsk: You're right, some leftovers of iterations

test/lib/mayaUsd/render/mayaToHydra/cpp/testMeshPrimvars.cpp
```diff
@@ -178,7 +178,7 @@ TEST(MeshPrimvars, ParamAttributesMatchGetLogic)
         "uvPivot", "displaySmoothMesh", "smoothLevel", "instObjGroups"
     };
 
-    const auto& paramAttrs = GetMeshParamAttributeNamesForTest();
+    const auto& paramAttrs = MAYAHYDRA_NS_DEF::GetMeshParamAttributeNamesForTest();
```
ppt-adsk: Should be just MayaHydra::GetMeshParamAttributeNamesForTest().

ppt-adsk (commented):
Nit-pick, if you could fix would be nice.

ppt-adsk (approved):

lanierd-adsk:
Failures on Linux are not due to these changes.

PR #406, opened 2026-04-08
HYDRA-2162 : Support outliner selection highlighting for Hydra Generative Procedurals

lanthia-adsk:
Moves the Generative Procedural Resolving Scene Index forward so generated prims are visible to the Wireframe Highlights scene indices. Caches material displacement to improve performance when selecting an animated object.


lib/mayaHydra/hydraExtensions/sceneIndex/mhGenerativeProceduralResolvingSceneIndex.h
```diff
@@ -0,0 +1,102 @@
...
+
+#include <pxr/usd/sdf/path.h>
+#include <pxr/imaging/hd/filteringSceneIndex.h>
+#include <pxr/imaging/hdGp/generativeProceduralResolvingSceneIndex.h>
+
+#include <flowViewport/sceneIndex/fvpSceneIndexUtils.h>
+
+namespace MAYAHYDRA_NS_DEF {
+
+class MhGenerativeProceduralResolvingSceneIndex;
```
lanierd-adsk, line 33: Could you please generate some comments about what this scene index does ? It would help when readin the code, later. Thanks !

lanierd-adsk (changes requested):
Thanks you for doing this ! Only a small comment to add to make things clearer.

lib/mayaHydra/hydraExtensions/sceneIndex/mhGenerativeProceduralResolvingSceneIndex.h
lanthia-adsk: Done! Let me know if anything is still unclear. 🙂

lanierd-adsk: Thank you. Actually the comment doesn't tell me why we are using it. 
What I'd like is when I look at a header file : I understand what the class does and why we are using it. 
Thank you!

lanierd-adsk (changes requested):

lib/mayaHydra/hydraExtensions/sceneIndex/mhGenerativeProceduralResolvingSceneIndex.h
lanthia-adsk: I modified the comment, let me know if it still needs improvement! Thanks for your feedback

lib/flowViewport/sceneIndex/wireframeHighlights/fvpBaseWhSi.cpp
```diff
@@ -375,6 +376,7 @@ class _MaterialBlockingContainerDataSource : public HdContainerDataSource {
     HdContainerDataSourceHandle const _inputDataSource;
     HdSceneIndexBaseRefPtr const _inputSceneIndex;
     TfToken const _materialPathToken;
+    std::unordered_map<SdfPath, bool, SdfPath::Hash>* _displacementCache;
```
lilike-adsk: You can switch to use `const std::unordered_map<SdfPath, bool, SdfPath::Hash>& _displacementCache`, to avoid the potential null pointer deferencing issue

lanthia-adsk: Will do, thanks for pointing that out!

lib/mayaHydra/mayaPlugin/plugin.cpp
```diff
@@ -191,8 +191,14 @@ PLUGIN_EXPORT MStatus initializePlugin(MObject obj)
 
     // For now this is required for the HdSt backend to use lights.
     setEnv("USDIMAGING_ENABLE_SCENE_LIGHTS", "1");
+#if defined(_WIN32) && PXR_VERSION < 2511
```
lanierd-adsk: Why is this only for Windows builds ?
If this is intended then it needs a comment. If it's a mistake, please remove it. Thank you !

lanierd-adsk (changes requested):

lib/mayaHydra/mayaPlugin/plugin.cpp
lilike-adsk: I'd suggest to remove defined(_WIN32) as we want consistent behavior for all platforms

lib/mayaHydra/hydraExtensions/sceneIndex/mhGenerativeProceduralResolvingSceneIndex.h
```diff
@@ -0,0 +1,111 @@
...
+        const PXR_NS::HdSceneIndexBase& sender,
+        const PXR_NS::HdSceneIndexObserver::DirtiedPrimEntries& entries) override
+    {
+        if (!_IsObserved())
+            return;
+        _SendPrimsDirtied(entries);
+    }
+
+private:
+    std::set<PXR_NS::SdfPath> _excludedSceneRoots;
```
lilike-adsk: I didn't see _excludedSceneRoots was used anywhere, should we apply the filtering in GetPrim() or just Remove it?

lanierd-adsk (dismissed):

lib/mayaHydra/hydraExtensions/sceneIndex/mhGenerativeProceduralResolvingSceneIndex.h
lanthia-adsk: I think we can remove it. Since excluded prims are not generative procedurals I think they will be ignored anyway by the Gp resolving scene index.

lilike-adsk (approved):

lanierd-adsk (approved):

PR #410, opened 2026-04-16
HYDRA-1935 : Fix isolate-select visibility for USD light gizmos and USD Cameras gizmos

lanierd-adsk:
Also fix a missing Ufe::GlobalSelection::removeObserver in ~MtohRenderOverride that could cause a use-after-free crash at shutdown.


lib/flowViewport/sceneIndex/fvpIsolateSelectSceneIndex.cpp
```diff
@@ -207,6 +218,18 @@ HdSceneIndexPrim IsolateSelectSceneIndex::GetPrim(const SdfPath& primPath) const
...
+        // visibility=false (e.g. from Maya's internal scene management
+        // filtering render items during isolate select) is overridden.
+        // This makes IsolateSelectSceneIndex the authoritative source of
+        // visibility when isolate select is active in Hydra Storm.
+        // HYDRA-1242: skip GeomSubset prims here as well — setting
+        // visibility on them hangs Hydra Storm.
+        inputPrim.dataSource = HdContainerDataSourceEditor(inputPrim.dataSource)
+            .Set(HdVisibilitySchema::GetDefaultLocator(), visOn)
+            .Finish();
+    }
```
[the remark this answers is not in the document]
lanierd-adsk: This is intentional :
When isolate select is active, the user has explicitly chosen to view only the selected prims. Forcing visOn for included prims is the expected semantic -- it mirrors VP2's behavior where isolate-selecting an object always shows it, even if it was previously hidden. The IsolateSelectSceneIndex is meant to be the sole authority on visibility during isolate select.

The specific scenario the visOn fixes: Maya's internal scene management (upstream of this scene index) applies its own isolate-select filtering by setting visibility=false on render items. Without the visOn override, those upstream opinions leak through and hide prims that the user explicitly isolate-selected in Hydra Storm.

If a user hides an object (cmds.hide) and then isolate-selects it, showing it is correct -- they're explicitly asking to view it in isolation. When isolate select is turned off, the original upstream visibility is restored because GetPrim() passes through inputPrim unmodified (the early return at the top of the function when _isolateSelection is empty).

test/lib/mayaUsd/render/mayaToHydra/testIsolateSelectPerShape.py
ppt-adsk, line 1: Thanks for the thorough test.

lib/flowViewport/sceneIndex/fvpIsolateSelectSceneIndex.cpp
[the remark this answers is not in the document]
ppt-adsk: As discussed privately, in VP2 hidden status is preserved by isolate select.

lib/mayaHydra/mayaPlugin/envSettings.cpp
```diff
@@ -23,6 +23,12 @@ TF_DEFINE_ENV_SETTING(
     MAYA_HYDRA_SINGLE_FRAME_PASS,
     false,
     "Use single frame pass when using the same renderer for all passes.");
+
+TF_DEFINE_ENV_SETTING(
+    MAYA_HYDRA_ISOLATE_SELECT_TF_DEBUG,
+    false,
+    "When true, mayaHydra enables FVP_ISOLATE_SELECT_VIEW_SELECTED and FVP_ISOLATE_SELECT_SCENE_INDEX at "
```
ppt-adsk: Why is this more useful than just specifying `TF_DEBUG="FVP_ISOLATE_SELECT*"`?

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -2323,6 +2335,365 @@ void MtohRenderOverride::_RenderOverrideChangedCallback(
...
+    if (!_mayaHydraSceneIndex) {
+        return;
+    }
+
+    // Quick check: skip all work when no view-selected object lives in a
+    // USD proxy.  USD prims are identified by the MayaUsd UFE runtime id;
+    // pure Maya DAG objects use a different runtime and don't need the
+    // native-rprim expansion performed below.
+    bool anyUsdPrimFromProxy = false;
+    for (const auto& ufeStr : viewSelectedUfePathStrings) {
```
ppt-adsk: At line 2833 we're already looping over all view selected (isolate selected) objects and creating a Ufe::Path from them.  Can we not add to that loop some filtering and path translation functionality, instead of looping over the full set of inputs again?

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -2323,6 +2335,365 @@ void MtohRenderOverride::_RenderOverrideChangedCallback(
...
+    auto addPrimPath = [&](const SdfPath& p) {
+        if (p.IsEmpty() || added.count(p) != 0) {
+            return;
+        }
+        if (selection.Add(Fvp::PrimSelection { p })) {
+            added.insert(p);
+        }
+    };
+
+    auto addMayaUsdProxyShapeNativePrefixFromUfe = [&](const std::string& ufeStr) {
```
ppt-adsk: This searching for a command to separate out the UFE path segments is more UFE path string processing.  We have already created a Ufe::Path as part of this processing, we should be able to just extract needed segment(s) from the Ufe::Path itself.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -2323,6 +2335,365 @@ void MtohRenderOverride::_RenderOverrideChangedCallback(
...
+    };
+
+    // Native visual rprims (camera gizmos, light shapes, etc.) live in the frame-pass render index
+    // that feeds the viewport; MayaHydraSceneIndex::GetRenderIndex() may not list them.
+    HdRenderIndex* renderIndexForScan = renderIndex(0);
+    if (!renderIndexForScan) {
+        renderIndexForScan = &_mayaHydraSceneIndex->GetRenderIndex();
+    }
+
+    auto addMayaCameraDrawables = [&](const MDagPath& camShapeDag) {
```
ppt-adsk: We should be able to ask the UFE scene item (and only for the UFE run time ID that is for USD) if it's a camera or a light.  See
https://git.autodesk.com/media-and-entertainment/ufe/blob/5497b4c3d1e87a9ec274d218e25d807402b68543/include/light2Handler.h#L43
and
https://git.autodesk.com/media-and-entertainment/ufe/blob/5497b4c3d1e87a9ec274d218e25d807402b68543/include/cameraHandler.h#L44

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -2323,6 +2335,365 @@ void MtohRenderOverride::_RenderOverrideChangedCallback(
...
+        // GetRprimIds() at isolate-build time, or may appear in a different HdRenderIndex than
+        // renderIndexForScan, but still resolve under the same DAG-derived prefixes at draw time.
+        if (!xfRprimPrefix.IsEmpty()) {
+            addPrimPath(xfRprimPrefix);
+        }
+        if (!shapeRprimPrefix.IsEmpty()) {
+            addPrimPath(shapeRprimPrefix);
+        }
+
+        for (const SdfPath& id : renderIndexForScan->GetRprimIds()) {
```
ppt-adsk: Are we looping over all existing Rprim's here?

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -2323,6 +2335,365 @@ void MtohRenderOverride::_RenderOverrideChangedCallback(
...
+    // O(numSelected * numCachedNativeRprims) without repeated full scans.
+    std::vector<SdfPath> nativeRprims;
+    {
+        std::unordered_set<SdfPath, SdfPath::Hash> seen;
+        for (const SdfPath& id : renderIndexForScan->GetRprimIds()) {
+            if (id.HasPrefix(MAYA_NATIVE_ROOT) && seen.insert(id).second) {
+                nativeRprims.push_back(id);
+            }
+        }
+        for (const SdfPath& p :
```
ppt-adsk: How big can this loop be?

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -2323,6 +2335,365 @@ void MtohRenderOverride::_RenderOverrideChangedCallback(
...
+             HdSceneIndexPrimView(_mayaHydraSceneIndex, rprimRoot)) {
+            if (seen.insert(p).second) {
+                nativeRprims.push_back(p);
+            }
+        }
+    }
+
+    for (const auto& ufeStr : viewSelectedUfePathStrings) {
```
ppt-adsk: We seem to be looping over all isolate selected objects again.

lib/mayaHydra/mayaPlugin/envSettings.cpp
lanierd-adsk: You're right, I am removing it.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lanierd-adsk: You're right, I am doing that now.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lanierd-adsk: Right, I am removing the storage of the ufe path.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lanierd-adsk: Thanks, I am doing it this way now.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lanierd-adsk: Yes, but I changed this code , now I am only getting the camera and lights prims , please review it again.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lanierd-adsk: This code has changed, please review it again.

lanierd-adsk:
Preflight has passed now.

lib/flowViewport/sceneIndex/fvpIsolateSelectSceneIndex.h
```diff
@@ -146,6 +148,17 @@ class IsolateSelectSceneIndex :
     FVP_API
     SelectionPtr GetIsolateSelection() const;
 
+    /// Prim paths explicitly added by the native-rprim expansion for USD
+    /// camera/light gizmos.  Maya's VP2 isolate-select filtering incorrectly
+    /// hides these render items because VP2 is unaware that the Hydra pipeline
+    /// has included them.  GetPrim() forces visibility ON for any included prim
+    /// whose path is in this set.
+    FVP_API
+    void SetForceVisiblePaths(std::unordered_set<PXR_NS::SdfPath, PXR_NS::SdfPath::Hash>&& paths);
```
ppt-adsk: Minor: should just use `TfHashSet<SdfPath>`

lib/flowViewport/sceneIndex/fvpIsolateSelectSceneIndex.h
```diff
@@ -146,6 +148,17 @@ class IsolateSelectSceneIndex :
     FVP_API
     SelectionPtr GetIsolateSelection() const;
 
+    /// Prim paths explicitly added by the native-rprim expansion for USD
```
ppt-adsk: Minor: this is for UFE camera / light gizmos, of which USD is a very important client (and currently perhaps the only one?)

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -2395,22 +2395,16 @@ bool ReadMayaDagPathFromUsdPullMetadata(const UsdPrim& prim, MDagPath& outDagPat
...
-        return false;
-    }
-    return sel.getDagPath(0, outDagPath) == MS::kSuccess && outDagPath.isValid();
-}
-
 } // namespace
 
-// When isolate select is built from view-selected UFE paths, certain USD prim types (cameras, lights,
-// etc.) have native Maya visual representations (gizmo rprims, HdCamera sprims) drawn under
+struct MtohRenderOverride::SelectedUsdGizmoPrim {
```
ppt-adsk: Feels like there is duplication in this struct.  A UsdPrim can be obtained from the Ufe::Path, and we can check whether a UsdPrim is a camera by applying a schema to it.

Also, this struct implies that UFE lights and cameras that are not USD wouldn't be supported, which feels very awkward.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -2420,61 +2414,49 @@ bool DagPathFromString(const std::string& dagPathStr, MDagPath& outDagPath)
...
 // clientData: multiple overrides each register a view-selected callback, and a single global
 // expander would capture the wrong `this`.
 void MtohRenderOverride::_ExpandIsolateSelectionForUsdPrims(
-    Fvp::Selection&                        selection,
-    const std::vector<std::string>&        viewSelectedUfePathStrings,
-    const MDagPath&                        panelCameraDag)
+    Fvp::Selection&                                              selection,
+    const std::vector<MtohRenderOverride::SelectedUsdGizmoPrim>& selectedGizmoPrims,
+    const MDagPath&                                              panelCameraDag,
+    std::unordered_set<SdfPath, SdfPath::Hash>&                  outForceVisiblePaths)
```
ppt-adsk: Could just return this from the method, with Return Value Optimization this will be just as efficient.

ppt-adsk (dismissed):
Minor points and questions, which you may choose to implement if you want.  Great work!

ppt-adsk (dismissed):

lib/flowViewport/sceneIndex/fvpIsolateSelectSceneIndex.cpp
```diff
@@ -207,6 +218,20 @@ HdSceneIndexPrim IsolateSelectSceneIndex::GetPrim(const SdfPath& primPath) const
...
+        // client): VP2's isolate-select filtering sets their MVS_visible to
+        // false because VP2 does not know that the Hydra pipeline has
+        // included them.  Force visibility ON so these gizmos are drawn when
+        // their owning UFE prim is isolate-selected.  All other included
+        // prims keep their upstream visibility as-is.
+        // HYDRA-1242: skip GeomSubset prims — setting visibility on them
+        // hangs Hydra Storm.
+        inputPrim.dataSource = HdContainerDataSourceEditor(inputPrim.dataSource)
+            .Set(HdVisibilitySchema::GetDefaultLocator(), visOn)
+            .Finish();
```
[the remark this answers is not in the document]
lanierd-adsk: We are not implementing this suggestion as we know that for a usd Camera/Light we will always have a single matching maya render tiem being its gizmo. So dealing with a hierarchy is not necessary.

ppt-adsk (approved):

PR #414, opened 2026-04-17
Handle duplicate fastIds in render item removal list

dj-mcg:
The removal list passed to UpdateRenderItems can contain duplicate fastIds.  After the first _RemoveRenderItem succeeds, subsequent lookups for the same id return nullptr.  The previous assert(ria != nullptr) fires in debug/non-NDEBUG builds when this happens.

Replace the assert with a nullptr check and continue, since a duplicate removal is benign — the item was already removed.

### DIFF
--- a/lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
+++ b/lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
@@ -458,7 +458,12 @@ void MayaHydraSceneIndex::UpdateRenderItems(const MDataServerOperation::MViewpor
         if (_GetRenderItem(fastId, ria)) {
             _RemoveRenderItem(ria);
         }
-        assert(ria != nullptr);
+        // The removal list can contain duplicate fastIds.  After the first
+        // removal succeeds, subsequent duplicates will not be found — skip
+        // them.
+        if (ria == nullptr) {
+            continue;
+        }
     }
 
     // My version, does minimal update


lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -458,7 +458,12 @@ void MayaHydraSceneIndex::UpdateRenderItems(const MDataServerOperation::MViewpor
         if (_GetRenderItem(fastId, ria)) {
             _RemoveRenderItem(ria);
         }
-        assert(ria != nullptr);
+        // The removal list can contain duplicate fastIds.  After the first
+        // removal succeeds, subsequent duplicates will not be found — skip
+        // them.
+        if (ria == nullptr) {
+            continue;
+        }
```
debloip-adsk, line 466: Since this is at the end of the for loop, we don't really need to continue and we could just remove this and the assert entirely. I wonder, would using a TF_WARN instead be a viable alternative? That way we still log something but not make it as critical. Are there a lot of duplicates or is this a rare thing?

dj-mcg: I'm only starting to dig in here, but the assert was causing some maya-hydra unit tests to fail for us since we auto-fail unit tests on asserts. Since the tests themselves were creating duplicates it seems like it might be pretty common

debloip-adsk: Yeah I took a quick look and this doesn't really seem all that critical, and I'm not sure it even needs a warning.

debloip-adsk (approved):

PR #415, opened 2026-04-21
HYDRA-2194 : Support transforms at object level for Hydra Generative Procedurals

lanthia-adsk:
When selecting a procedural, you can move it directly from the viewport. All generated meshes will get the same transform applied to them.
Also added unit tests for picking from the viewport and transforms (rotation, translations, scale) for procedurals.


lanierd-adsk (dismissed):

lib/mayaHydra/hydraExtensions/sceneIndex/mhGenerativeProceduralResolvingSceneIndex.cpp
```diff
@@ -36,4 +45,126 @@ MhGenerativeProceduralResolvingSceneIndex::New(const HdSceneIndexBaseRefPtr& inp
...
 
+// Composes the parent procedural's world matrix into generated children's transform, 
+// so the generated geometry moves correctly when the procedural root is transformed.
+// Returns the modified prim with the overlaid world-space transform.
+HdSceneIndexPrim MhGenerativeProceduralResolvingSceneIndex::GetPrim(const SdfPath& primPath) const
+{
+    HdSceneIndexPrim prim = GetInputSceneIndex()->GetPrim(primPath);
+
+    // Walk up from the prim to find the nearest GP root ancestor.
+    SdfPath generativeProceduralRoot;
```
lilike-adsk, line 60: GetPrim() is a hot call, could we add a check here to return eariler if _generativeProceduralPaths is empty()?

lilike-adsk (approved):

lib/mayaHydra/hydraExtensions/sceneIndex/mhGenerativeProceduralResolvingSceneIndex.cpp
```diff
@@ -36,4 +45,130 @@ MhGenerativeProceduralResolvingSceneIndex::New(const HdSceneIndexBaseRefPtr& inp
...
+                break;
+            }
+            parentPath = parentPath.GetParentPath();
+        }
+    }
+    if (generativeProceduralRoot.IsEmpty())
+        return prim;
+
+    // Compose transforms from the prim up through all ancestors to the GP root.
+    GfMatrix4d result(1.0);
```
lanierd-adsk, line 75: I am wondering if we should rather do this matrix composition in the _PrimsAdded function, so we would do it once only.
As opposed to what we do presently in the GetPrim. We'll recompute the matrix many times when the other scene indices call GetPrim.
@lilike-adsk what do you think ?

lanthia-adsk: The `ancestor.dataSource` value changes every time there is a transform applied to the root. If we do the matrix composition in `PrimsAdded`, it is not going to be called every time the transforms changes so the matrix will not be properly updated

ppt-adsk: Feels strange that it is up to a filtering scene index to do this work of flattening the transforms.  Why is this not done in the render delegate itself?  The render delegate will have pulled all the parents, and so could perform the transform evaluation.
If this is simply the way that generative procedural transforms are expected to be supported in Hydra, please ignore this comment.

lilike-adsk: USD Imaging also leverages HdFlatteningSceneIndex to do flatterning on USD nodes, not sure why it's not working on USD procedural node though

lilike-adsk: @lanierd-adsk you're right for the performance concern, if we do it in _PrimsAdded and cache xform, we need to add a GetPrim() call in _PrimsAdded, how about we add xform flattern cache directly in GetPrim()? I saw HdFlatteningSceneIndex added a similar cache.

lanierd-adsk: I agree. Anais and I were discussing this and it's probably the best solution to have a cache in the GetPrim function.

lilike-adsk: Seems the caching is bit complicated than expected, I'd suggest we revisit this in the future

lanierd-adsk (approved):

PR #422, opened 2026-05-04
Update HVT submodule

debloip-adsk:


debloip-adsk:
Deferred - HYDRA-2275

PR #430, opened 2026-05-12
HYDRA-1767 : Pass subdiv info to Hydra and change vertex normals to face normals in MeshAdapter

benyoon-adsk:
After discussing with the team, we decided on the following behavior for Hydra subdivision handling:
1. If no subdivision is applied (refineLevel == 0), use the normals (including custom normals) provided by Maya.
2. Else, ignore Maya normals and use Hydra generated normals after subdivision.


lib/mayaHydra/hydraExtensions/adapters/meshAdapter.cpp
```diff
@@ -263,7 +276,7 @@ class MayaHydraMeshAdapter : public MayaHydraShapeAdapter
         return VtValue(ret);
     }
 
-    /// Return vertex normals as a primvar value.
+    /// Return face-varying normals as a primvar value.
```
benyoon-adsk, line 279: Custom normals from Maya requires face-varying normals. As a side effect, this also fixes the issue where meshes with hard edges, such as cubes, appeared incorrect due to the use of vertex normals.

lilike-adsk: yes, originally, we use unexpaded vertices + unexpanded vertices normals + vertex interpolation, and produce that inconsistent smoothish effect, since Maya MfnMesh can produce expanded face vertices, we have two options:
1. Use unexpanded vertices + expanded face vertice normals/uvs + face-varying interpolation.
2. Use expanded face vertices + expanded face vertice normals/uvs + vertex interpolation.

For Viewport rendering with renderItemAdapter, we choose #2 for faster fps over less data transfering as face-varying interpolation is more complicated at shading side, storm needs geometry shader to do face-varying interpolation.

For batching rendering with meshAdapter, we can use #1 for less vertices data transfering.

lib/mayaHydra/hydraExtensions/adapters/meshAdapter.cpp
```diff
@@ -361,17 +386,11 @@ class MayaHydraMeshAdapter : public MayaHydraShapeAdapter
...
-            HdMeshTopology(
-                        PxOsdOpenSubdivTokens->none,//For the OGS normals vertex buffer to be used, we need to use PxOsdOpenSubdivTokens->none
-                        UsdGeomTokens->rightHanded,
-                        faceVertexCounts,
-                        faceVertexIndices) :
-            HdMeshTopology(
-            (GetMayaHydraSceneIndex()->GetParams().displaySmoothMeshes || GetDisplayStyle().refineLevel > 0)
-                ? PxOsdOpenSubdivTokens->catmullClark
-                : PxOsdOpenSubdivTokens->none,
+        return HdMeshTopology(
```
lanierd-adsk, line 389: I think you have removed the case of GetMayaHydraSceneIndex()->GetParams().displaySmoothMeshes which should still be handled if the user wants to see faceted meshes.

lanierd-adsk (changes requested):
Thanks for doing this, only one more case to handle I think.

lib/mayaHydra/hydraExtensions/adapters/meshAdapter.cpp
lilike-adsk: Is MayaHydraParams.displaySmoothMeshes a global flag? But I didn't see it was connected to anything, it was always true currently. The meshAdapter::GetDisplayStyle() is respecting the per mesh flag from MayaAttrs::mesh::displaySmoothMesh.

For the subdivion info, I saw there was a PxOsdSubdivTags in HdMeshTopology, should we use meshAdapter::GetSubdivTags() and feed it to HdMeshTopology, it including more information than just catmullClark/none

benyoon-adsk: @lanierd-adsk Shouldn't subdivision not be dependent on viewport shading mode? I noticed MayaHydraParams::displaySmoothMeshes is actually a toggle for Smooth Shade All / Flat Shade All, as we see in: https://github.com/Autodesk/maya-hydra/blob/dev/lib/mayaHydra/mayaPlugin/renderOverride.cpp#L1025-L1028

I think only refineLevel should influence whether we use Maya vs Hydra normals. Also I verified faceted mesh (Flat Shade All) works with current code, as it's handled in the link above. Should we change the naming from `displaySmoothMeshes` to something else like: `smoothShadeEnabled` ?

benyoon-adsk: @lilike-adsk Yes `MayaHydraParams::displaySmoothMeshes` appears to be a global flag. [This](https://github.com/Autodesk/maya-hydra/blob/dev/lib/mayaHydra/mayaPlugin/renderOverride.cpp#L1025-L1028) (Smooth Shade All / Flat Shade All) toggles displaySmoothMeshes which is used [here](https://github.com/Autodesk/maya-hydra/blob/dev/lib/mayaHydra/hydraExtensions/adapters/renderItemAdapter.cpp#L502) but I think it should only be refineLevel, like I wrote in the comment above. What do you think?

It's also used [here](https://github.com/Autodesk/maya-hydra/blob/dev/lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp#L946) but I'm not sure what `MayaHydraSceneIndex::SetParams()` does.

> For the subdivion info, I saw there was a PxOsdSubdivTags in HdMeshTopology, should we use meshAdapter::GetSubdivTags() and feed it to HdMeshTopology, it including more information than just catmullClark/none

I think we should, I'll try to implement this.

benyoon-adsk:
Added `PxOsdSubdivTags` in `HdMeshTopology` along with a unit test. We can set crease/corners in Maya and pass the following to Hydra:

- faceVaryingLinearInterpolation
- interpolateBoundary
- triangleSubdivisionRule
- cornerIndices
- cornerSharpnesses
- creaseIndices
- creaseLengths
- creaseSharpnesses

In mayaHydraSceneBrowser, with mesh subdivided:

Screenshot 2026-05-13 161256

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraDataSource.cpp
```diff
@@ -144,6 +146,36 @@ MayaHydraDataSource::Get(const TfToken& name)
                         .Build())
                 .SetSubdivisionScheme(
                     HdRetainedTypedSampledDataSource<TfToken>::New(topology.GetScheme()))
+                .SetSubdivisionTags(
+                    HdSubdivisionTagsSchema::Builder()
+                        .SetFaceVaryingLinearInterpolation(
+                            HdRetainedTypedSampledDataSource<TfToken>::New(
+                                subdivTags.GetFaceVaryingInterpolationRule()))
+                        // PxOsdSubdivTags::VertexInterpolationRule and
```
lanierd-adsk, line 154: I think this comment is no longer relevant, right ? If so, please remove it.

benyoon-adsk: It's a new comment that I added myself, I thought it was confusing because GetVertexInterpolationRule doesn’t semantically match SetInterpolateBoundary the way the other APIs do (e.g. SetTriangleSubdivisionRule ↔ GetTriangleSubdivision()). If it's obvious, I'll remove it.

lanierd-adsk: It's not obvious. But I may have missed where you use this. Are you using PxOsdSubdivTags::VertexInterpolationRule or HdSubdivisionTagsSchema::interpolateBoundary in the code ? 
As for me it's a comment that doesn't seem related to the code.

lanierd-adsk (commented):
There is still one comment we should clarify. Thank you !

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraDataSource.cpp
benyoon-adsk: Yup, right below the comment:
```
                        // HdSubdivisionTagsSchema::interpolateBoundary are the same value under two
                        // different names
                        .SetInterpolateBoundary(
                            HdRetainedTypedSampledDataSource<TfToken>::New(
                                subdivTags.GetVertexInterpolationRule()))
```
I wrote HdSubdivisionTagsSchema::interpolateBoundary and PxOsdSubdivTags::VertexInterpolationRule because I thought their naming felt more global/generalized than saying:
```
                        // PxOsdSubdivTags::GetVertexInterpolationRule and
                        // HdSubdivisionTagsSchema::SetInterpolateBoundary are the same value under two
                        // different names
```

lanierd-adsk (approved):

lilike-adsk (approved):
Nice work!

benyoon-adsk:
Preflight failed due to "machine went offline" error. Re-ran the failed tests and they now pass. See Build External GitHub Branch (github.com) #489

PR #431, opened 2026-05-13
Use new API for HdRendererPlugin::IsSupported

dj-mcg:
This API will change in the next release version to take an HdRendererCreateArgsSchema argument

### DIFF
--- a/lib/mayaHydra/mayaPlugin/pluginUtils.cpp
+++ b/lib/mayaHydra/mayaPlugin/pluginUtils.cpp
@@ -29,7 +29,7 @@
 #include "tokens.h"
 
 #include <pxr/imaging/glf/contextCaps.h>
-#if HD_API_VERSION >= 98
+#if HD_API_VERSION >= 98 // PXR_VERSION > 2605
 #include <pxr/imaging/hd/rendererCreateArgsSchema.h>
 #endif
 #include <pxr/imaging/hd/rendererPlugin.h>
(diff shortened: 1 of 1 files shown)


debloip-adsk (dismissed):
Thank you for the change!

debloip-adsk:
We are experiencing some build instability issues currently so the preflight failure is expected for now

lib/mayaHydra/mayaPlugin/pluginUtils.cpp
```diff
@@ -29,6 +29,9 @@
 #include "tokens.h"
 
 #include <pxr/imaging/glf/contextCaps.h>
+#if HD_API_VERSION >= 98
```
lanierd-adsk: I think it would help to know which version of OpenUSD/Hydra this is as, such as 26.05 or 25.11. 
As HD_API_VERSION >= 98 isn't really meaningful.

dj-mcg: This is really what Hydra callsites should be using for versioning. I can add a comment to say what version of USD this will be released with if you want, but using the Hydra API version itself is more appropriate here.

dj-mcg: https://github.com/PixarAnimationStudios/OpenUSD/blob/6a11468ddb15f547dd0b31288fee1546d2a919ea/pxr/imaging/hd/version.h#L145

dj-mcg: I added a comment with the latest commit indicating you need whatever version will be after 26.05 (probably 26.08 but that's not guaranteed)

lanierd-adsk (approved):

debloip-adsk:
Preflight failures were unrelated and are being looked into, so I went ahead and merged the PR

PR #432, opened 2026-05-15
HYDRA-1935 : Fix when selecting multiple instances and doing isolate …

lanierd-adsk:
…select the result was incorrect/crashed.


lib/flowViewport/sceneIndex/wireframeHighlights/fvpPiInstancerWhSi.cpp
```diff
@@ -197,7 +197,18 @@ PiInstancerWhSiRefPtr PiInstancerWhSi::New(
...
     SelectionKey selectionKey = SelectionKeyFromPath(selectionPath);
-    auto primSelection = _selections.at(selectionKey)._primSelection;
+
+    // During scene-index teardown, HdDependencyForwardingSceneIndex::_PrimsRemoved
+    // can drive _UpdateDependencies → GetPrim on a path whose entry has already
+    // been erased from _selections, while BaseWhSi::_selectionPaths (consulted
+    // before dispatching here) is still populated.  Without this guard, .at()
+    // throws std::out_of_range from a destructor-side path and crashes Maya.
+    // Same defensive pattern as the .find() checks at the bottom of this file.
+    auto selectionIt = _selections.find(selectionKey);
```
lanierd-adsk, line 207: HEre I fixed a crash when closing maya after an isolate select with multiple instances

lib/flowViewport/sceneIndex/fvpIsolateSelectSceneIndex.cpp
```diff
@@ -588,13 +616,26 @@ IsolateSelectSceneIndex::_CollectInstancers(
...
     for (const auto& primSelectionsEntry : *isolateSelection) {
-        // If the prim itself is a point instancer, add it and continue.
-        if (isPointInstancer(GetInputSceneIndex(), primSelectionsEntry.first)) {
+        const auto& primSelections = primSelectionsEntry.second;
+
+        // If any PrimSelection for this key carries nestedInstanceIndices, the
+        // key IS the point instancer path.  Collect it directly, regardless of
+        // what isPointInstancer() returns for the prim in the scene index — the
+        // prim's representation may have been modified by intermediate scene
+        // index filters and could be misclassified.
```
debloip-adsk: If the prim may have been misclassified, should the correct fix not be to avoid this misclassification in the first place?

lib/flowViewport/sceneIndex/fvpIsolateSelectSceneIndex.cpp
```diff
@@ -631,9 +672,23 @@ IsolateSelectSceneIndex::_CreateInstancerMasks(
...
         auto instanceLocationsDs = instancerTopologySchema.GetInstanceLocations();
-        if (!instanceLocationsDs) {
-            auto primSelections = isolateSelection->GetPrimSelections(instancerPath);
 
+        // Fetch the selections for this instancer path up front so we can
+        // detect point-instance selections (those carrying nestedInstanceIndices)
+        // independently of instanceLocationsDs.  An intermediate scene-index
+        // filter may have populated instanceLocations for a prim that is really
+        // a point instancer, which would cause the native-instancer branch to
+        // produce a wrong mask.  Whenever the selection itself carries
```
debloip-adsk: AFAIK there should not be instanceLocations on a point instancer. If there is, then that suggests to me the bug is elsewhere and this is just a band-aid patch, I think we should understand what the root issue is before resorting to workarounds

lib/flowViewport/sceneIndex/fvpIsolateSelectSceneIndex.cpp
lanierd-adsk: The right source of truth for "was this path picked as a point-instancer instance?" is the PrimSelection data built at pick time, not a runtime GetPrim query through the filtered scene index chain. The picking code sets nestedInstanceIndices precisely when an instance of a PointInstancer is selected — that information is stable regardless of how any intermediate filter represents the prim. Fixing the classification in every intermediate filter is neither feasible (we don't own them) nor appropriate (they may correctly transform prim topology for their own purposes). Using hasPointInstanceSelection is the correct design: it queries the data that was built specifically to carry this information, rather than inferring it from a potentially-transformed scene representation.

lib/flowViewport/sceneIndex/fvpIsolateSelectSceneIndex.cpp
lanierd-adsk: You're right that instanceLocations should not be present on a point instancer, and if it were, the fix would belong elsewhere. The comment was overstating the rationale. The actual reason for checking hasPointInstanceSelection is simpler and correct regardless: nestedInstanceIndices is set by the picking code at pick time with exact knowledge of what was selected, making it the most reliable signal for "this path is a point instancer" — more reliable than querying primType or isPointInstancer() through the scene index at mask-building time, which could reflect a different scene state. The comment has been updated to reflect this.

debloip-adsk (dismissed):
Thanks for the clarifications!

lanierd-adsk:
Preflight is failing only for a coverage build due to a node connection lost. 
If you agree @debloip-adsk please merge the PR if the code review is ok for you.

debloip-adsk (approved):

PR #433, opened 2026-05-15
HYDRA-2292 : Fix timeout errors on build jobs

benyoon-adsk:
Interactive unit tests might be failing due to console handshake with the default Windows terminal with our new build machines. Skip the console handshake.

### DIFF
--- a/test/testUtils/imageUtils.py
+++ b/test/testUtils/imageUtils.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 #
 import os
+import platform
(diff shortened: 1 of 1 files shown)


debloip-adsk (approved):
Let's hope we don't run into too many issues like this :sweat_smile: Thanks!

ppt-adsk (approved):

debloip-adsk:
Force-merging, the failed runs were due to intermittent errors and were successfully re-run individually

PR #435, opened 2026-05-21
Add -warn and -warnpercent idiff args to batch rendering

benyoon-adsk:
Currently in batch rendering, we don't set -warn and -warnpercent and warnpercent defaults to 0. For small pixel differences between ref image vs rendered image which could be considered as test pass results in test failure in consequence. We will consider WARN and FAIL as equivalent.


ppt-adsk (approved):

benyoon-adsk:
Preflight failed due to agent offline error again. Passed after re-running failed tests. See Build External GitHub Branch (github.com) #499

PR #438, opened 2026-05-22
HYDRA-2290 - Animation perf optimization: do not query material if no material change

lilike-adsk:
HYDRA-2290 - animation perf optimization: do not query material if no material change

Only query material shading node with Maya API when MDataServerOperation::MViewportScene::MVS_changedEffect is set.


vlasovi (approved):
Looks good to me.

PR #440, opened 2026-05-25
HYDRA-2289 : Add support for -s (start frame) and -e (end frame) flags to command-line batch rendering

lanierd-adsk:
Users can now restrict animation playback to a specific frame range (e.g., Render -s 5 -e 9 scene.ma) for all Hydra renderers without modifying scene settings. Implementation includes MEL-Python bridge for setting Maya render globals, validation to prevent invalid frame ranges, and comprehensive test coverage for both HdStorm and HdArnold.


cmake/test.cmake
```diff
@@ -289,6 +290,8 @@ function(_mayaHydra_setup_test_plugins)
          "${CMAKE_INSTALL_PREFIX}/lib/maya")
     list(APPEND MAYAHYDRA_VARNAME_MAYA_SCRIPT_PATH
          "${CMAKE_INSTALL_PREFIX}/scripts")
+    list(APPEND MAYAHYDRA_VARNAME_PYTHONPATH
```
ppt-adsk: Why was this addition necessary?  Why do other tests work without this PYTHONPATH change?

ppt-adsk (commented):
Looks good to me, just a question about changes in the test framework!

cmake/test.cmake
lanierd-adsk: The PYTHONPATH change doesn't belong in _mayaHydra_setup_test_plugins() — it belongs in the multi-image macro itself, mirroring line 916 of the single-image macro that was already there. The single-image tests (resolution, crop, etc.) work because they have this line; the multi-image macro was simply missing it, and the frame-range tests are the first multi-image tests that call a mayaHydra.* import.

ppt-adsk: Hmm, now I'm confused, your comment says "The PYTHONPATH change doesn't belong in _mayaHydra_setup_test_plugins()" --- but it is in the _mayaHydra_setup_test_plugins function!

Is it more a question that since it needs to be in the multi-image and single-image cases, you're factoring it out into common code, and it should be removed from the single-image case?

lanierd-adsk: You are right, here is what I did in the latest commit :

Removed the erroneous list(APPEND MAYAHYDRA_VARNAME_PYTHONPATH "${CMAKE_INSTALL_PREFIX}/scripts") from _mayaHydra_setup_test_plugins() — that was the wrong path (install dir, not source dir) and the wrong place (a shared function used by all test types).

Added list(APPEND MAYAHYDRA_VARNAME_PYTHONPATH "${MAYA_HYDRA_DIR}/scripts") directly into the multi-image macro body after the _mayaHydra_setup_test_plugins() call (line 155–156), exactly mirroring what the single-image macro already had at lines 917–918.

ppt-adsk (approved):

lanierd-adsk:
Merging as the preflight failure is a build node disconnection.

PR #442, opened 2026-05-25
HYDRA-1919 : Disable Hydra parallel RPrim sync and remove dg_access_mutex

benyoon-adsk:
Accessing data from the Maya DG in parallel will cause crashes. Until we get a proper integration with the Maya EM, we should disable Hydra parallel sync.

### DIFF
--- a/lib/mayaHydra/mayaPlugin/plugin.cpp
+++ b/lib/mayaHydra/mayaPlugin/plugin.cpp
@@ -36,6 +36,8 @@
 
 #include <pxr/base/plug/plugin.h>
 #include <pxr/base/plug/registry.h>
+#include <pxr/base/tf/debug.h>
+#include <pxr/imaging/hd/debugCodes.h>
 
 #include <mayaUsdAPI/utils.h>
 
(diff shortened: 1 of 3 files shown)


ppt-adsk (approved):

benyoon-adsk:
Preflight passed, see Build External GitHub Branch (github.com) #510

ppt-adsk:
Want to hear from @lilike-adsk about performance impact on USD data producers.

lib/mayaHydra/mayaPlugin/plugin.cpp
```diff
@@ -212,6 +214,10 @@ void onFileOpenCheckOrCreateRenderSettings(void* /*clientData*/)
 
 void initialize()
 {
+    // HYDRA-1919: Maya DG API is not thread-safe. Disable Hydra's parallel
+    // RPrim sync until proper Maya EM integration lands.
+    PXR_NS::TfDebug::Enable(PXR_NS::HD_DISABLE_MULTITHREADED_RPRIM_SYNC);
```
lilike-adsk, line 219: Could we do some tests with viewport animation playback with this PR?  In viewport rendering, most node attributes from Maya DG are already evaluated by Maya EM + VP2, so, after EM evaluation phase, it's ususally safe to call Maya DG APIs to query attributes from multithreads as most attributes are not dirty anymore, threfore no DG evaluation will be triggered, that's why there's no crash even without those locks for viewport rendering, until the custom attributes support was introduced, custom attributes are ususally not evaluated by Maya EM + VP2. If we get a big slowdown with this single pull, we might want to isolate the changes, only applied for production rendering.

PR #450, opened 2026-06-03
Tremblp/hydra 2337/allow fallback values

ppt-adsk:
Default USD render settings resolution attributes (for render settings and render products) have a (USD) fallback value, which is not authored.  This is the value shown in the attribute editor.

Existing code required render settings attributes to be explicitly authored, for no good reason.  A fallback value for a USD attribute is entirely legitimate, and should be respected.  Now it is.


lib/mayaHydra/mayaPlugin/batchRendering/hydraRenderCmdHydraV1RenderSettings.cpp
ppt-adsk, line 1: Removed the requirement for an attribute to have been explicitly authored.  A fallback value is perfectly acceptable.

scripts/mayaHydra/renderSettings/renderProducts.py
```diff
@@ -61,3 +61,15 @@ def applyToProductName(modifierFn):
                 "Could not obtain a product name attribute for render product "
                 "%s." % str(rp.GetPrim().GetPath()))
         pnAttr.Set(modifierFn(pnAttr.Get()))
+
+
+def applyToResolution(modifierFn):
```
ppt-adsk, line 66: Helper to apply a resolution setting to all render product(s) that need it (currently all).

scripts/mayaHydra/renderSettings/resolution.py
ppt-adsk, line 1: Set resolution not only on the render settings prim, but also on the render products, as render product values have precedence over render settings values.

test/lib/cmdLineRender/scenes/basic/sphereFallbackResolution.ma
ppt-adsk, line 1: Added test scene without explicitly authored USD resolution.  Fallback resolution is (2048x1080).

test/lib/cmdLineRender/CMakeLists.txt
```diff
@@ -78,6 +78,16 @@ mayaHydra_add_cmd_line_render_test(renderSettings/stormResolution.ma
...
 )
 
+# Test a render using the fallback USD resolution settings
+mayaHydra_add_cmd_line_render_test(basic/sphereFallbackResolution.ma
+    cmake_path(GET scene STEM sceneFileNameNoExt)
+    COPY_SCENE
+    RENDERED_IMAGE_SUBDIR projects/default/images
+    RENDERED_IMAGE_NAME ${sceneFileNameNoExt}
+    RENDERER HdStormRendererPlugin
+    RENDERER_ARGS "-rd ../images -im ${sceneFileNameNoExt}"
```
ppt-adsk, line 88: No explicit resolution set from the command line, use the resolution in the file (in this case, the fallback USD resolution).

benyoon-adsk (approved):

PR #453, opened 2026-06-05
HYDRA-2280 - maya-hydra material and mesh issues uncovered while investigating the walrus chess-set scene

lanierd-adsk:
## Summary

This PR addresses several maya-hydra material and mesh issues uncovered while investigating the walrus chess-set scene, plus the supporting build configuration. The changes span MaterialX network translation, mesh tangents, multi-material (per-face) mesh binding, and batch-mode DAG traversal.

### MaterialX
- **Warn on unresolved MaterialX node definitions** (`materialAdapter.cpp`): after parsing the MaterialX document, emit a `TF_WARN` listing any node definitions that could not be resolved, along with the search paths used. Previously, `UsdMtlxRead` silently dropped unresolved nodes *and the branches feeding them*, causing the interactive viewport network to diverge from a batch/USD render with no diagnostic. The warning makes this actionable.
- **Guard MaterialX usage behind `WANT_MATERIALX_BUILD`** (`materialAdapter.cpp`, `CMakeLists.txt`): the repo's top-level CMake supports USD packages built without `hdMtlx` (it disables MaterialX in that case). All MaterialX-dependent includes (`hdMtlx`, `usdMtlx`, MaterialX headers) and code (`PopulateMaterialXNetworkMap`) are now gated behind `WANT_MATERIALX_BUILD`, and the `hdMtlx`/`usdMtlx` link libraries are gated behind `CMAKE_WANT_MATERIALX_BUILD`. When MaterialX is disabled, material resolution falls back to the standard Maya material network converter.

### Mesh tangents
- **Fix tangent primvar type** (`meshAdapter.cpp`): tangents from Maya's `MFnMesh::getTangents` (3-float `MFloatVectorArray`) were `reinterpret_cast` to `GfVec2f`, which mis-strided the buffer and corrupted the data. They are now built into a `VtVec3fArray` with an explicit `GfVec3f(x, y, z)` conversion, matching OpenUSD/Hydra, where tangents are `VtVec3fArray`/`GfVec3f` and consumed as `vec3`.
- **Fix tangent primvar role** (`meshAdapter.cpp`): the `tangents` primvar role was `textureCoordinate`; it is now `vector`, matching how OpenUSD tags direction vectors (`UsdImagingUsdToHdRole`; velocities/accelerations use `vector`). `st` continues to use `textureCoordinate`.

### Multi-material (per-face) meshes
- **Emit geom subsets for multi-material meshes** (`mayaHydraSceneIndex.cpp/.h`, `mayaUtils.cpp/.h`, `shapeAdapter.cpp/.h`): added `GetAllShadingAssignments` / `FindShadingEngine` helpers and `_InsertGeomSubsetsForMesh`, which create the bound material(s) and, when a mesh has more than one shading group, emit one `HdGeomSubset` (faceSet + material binding) per assignment so per-face/multi-material meshes shade correctly. `MayaHydraShapeAdapter::GetMaterial` now accepts an optional component to support per-face lookups.

### Batch-mode DAG traversal
- **Skip invisible shapes in batch `InsertDag`** (`mayaHydraSceneIndex.cpp`): in batch mode (`useMeshAdapter`), `MItDag` visits every DAG node including LOD meshes and corrective blend-shape targets that VP2 never makes render items for. Invisible shapes are now skipped so only the intended renderable geometry reaches Hydra, matching viewport behavior.

## Notes

- An earlier approach that bridged `MATERIALX_SEARCH_PATH` by importing custom nodedefs/nodegraphs into the per-material document was implemented and reverted: mutating the document changed what `UsdMtlxRead` interpreted, producing a corrupted network (wrong terminal) in both viewport and batch. The recommended way to achieve viewport↔batch library parity is configuration (add the custom library to `PXR_MTLX_STDLIB_SEARCH_PATHS`), not document mutation.
- The walrus eye material issue that originally triggered the investigation was a separate material-authoring bug (missing `Mask`), fixed by the artist in LookdevX.

## Testing

- [ ] Walrus chess-set scene renders correctly in the HdArnold viewport and batch.
- [ ] Multi-material meshes shade per-face correctly in the viewport.
- [ ] Meshes with tangents (e.g. normal-mapped MaterialX materials) render correctly.
- [ ] Builds configure and compile both with and without MaterialX support (`CMAKE_WANT_MATERIALX_BUILD` ON/OFF).
- [ ] Batch render does not include hidden LOD/blend-shape geometry.

### DIFF
--- a/lib/mayaHydra/hydraExtensions/adapters/materialAdapter.cpp
+++ b/lib/mayaHydra/hydraExtensions/adapters/materialAdapter.cpp
@@ -32,16 +32,23 @@
 #include <pxr/usd/sdf/types.h>
 #include <pxr/usd/usd/prim.h>
 #include <pxr/usd/usd/stage.h>
+#include <pxr/imaging/hdMtlx/hdMtlx.h>
 #include <pxr/usd/usdMtlx/reader.h>
 #include <pxr/usd/usdShade/material.h>
 #include <pxr/usdImaging/usdImaging/materialParamUtils.h>
 #include <pxr/usdImaging/usdImaging/tokens.h>
 
+#include <maya/MGlobal.h>
 #include <maya/MNodeMessage.h>
 #include <maya/MPlug.h>
 #include <maya/MPlugArray.h>
 
+#include <set>
+#include <string>
+#include <vector>
+
 #include <MaterialXCore/Document.h>
+#include <MaterialXFormat/File.h>
 #include <MaterialXFormat/XmlIo.h>
 
 PXR_NAMESPACE_OPEN_SCOPE
@@ -58,6 +65,25 @@ TF_DEFINE_PRIVATE_TOKENS(
 const VtValue       _emptyValue;
 const TfToken       _emptyToken;
 const TfTokenVector _stSamplerCoords = { TfToken("st") };
+
+// Collect the node-definition identifiers referenced by 'doc' (including inside
+// nodegraphs) that cannot currently be resolved from the document itself.
+void _CollectUnresolvedNodeDefs(const MaterialX::DocumentPtr& doc, std::set<std::string>& out)
+{
+    std::vector<MaterialX::NodePtr> nodes = doc->getNodes();
+    for (const auto& nodeGraph : doc->getNodeGraphs()) {
+        const auto graphNodes = nodeGraph->getNodes();
+        nodes.insert(nodes.end(), graphNodes.begin(), graphNodes.end());
+    }
+    for (const auto& node : nodes) {
+        if (node && !node->getNodeDef()) {
+            const std::string def = node->getNodeDefString();
+            if (!def.empty()) {
+                out.insert(def);
+            }
+        }
+    }
+}
 
 } // namespace
 
@@ -241,6 +267,16 @@ private:
         auto       p = node.findPlug(MayaAttrs::shadingEngine::surfaceShader, true);
         MPlugArray conns;
         p.connectedTo(conns, true, false);
+
+        // LookdevX / MaterialX materials connect to the separate
+        // "materialXSurfaceShader" attribute rather than "surfaceShader".
+        if (conns.length() == 0) {
+            auto pMtlx = node.findPlug("materialXSurfaceShader", true);
+            if (!pMtlx.isNull()) {
+                pMtlx.connectedTo(conns, true, false);
+            }
+        }
+
         if (conns.length() > 0) {
             _surfaceShader = conns[0].node();
             MFnDependencyNode surfaceNode(_surfaceShader, &status);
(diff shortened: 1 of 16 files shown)


lib/mayaHydra/hydraExtensions/adapters/materialAdapter.cpp
```diff
@@ -276,70 +318,200 @@ class MayaHydraShadingEngineAdapter : public MayaHydraMaterialAdapter
...
+                    cmd += ".renderDocument";
+                    MGlobal::executeCommand(cmd, false, false);
+                    TF_DEBUG(MAYAHYDRALIB_ADAPTER_MATERIALS)
+                        .Msg("PopulateMaterialXNetworkMap: Triggered evaluation via dgeval, retrying renderDocument\n");
+                    mtlxDocStr = mtlxDocPlug.asString();
+                } catch (...) {
+                    TF_DEBUG(MAYAHYDRALIB_ADAPTER_MATERIALS)
+                        .Msg("PopulateMaterialXNetworkMap: dgeval attempt failed\n");
+                }
+            }
```
ppt-adsk: This seems very awkward and redundant: why would this succeed if getting the value failed?

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -1442,6 +1427,91 @@ void MayaHydraSceneIndex::OnDagNodeRemoved(const MObject& obj)
...
+            MFnSingleIndexedComponent fnComp(sa.component);
+            MIntArray elements;
+            fnComp.getElements(elements);
+            faceIndices.resize(elements.length());
+            for (uint32_t j = 0; j < elements.length(); ++j) {
+                faceIndices[j] = elements[j];
+            }
+        } else {
+            continue;
+        }
```
ppt-adsk, line 1517: When can this happen?

ppt-adsk (commented):
Great work!  Just a couple of questions.

ppt-adsk:
Is there an automated test we can add for this?

lib/mayaHydra/hydraExtensions/adapters/materialAdapter.cpp
lanierd-adsk: You're right, it was a leftover from various attempt in fixing the problem, I am deleting it.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
lanierd-adsk: I've added a comment but it should not happen.

ppt-adsk (approved):

PR #458, opened 2026-06-15
HYDRA-2274 : Make maya hydra compliant with usd 26.05

lanierd-adsk:
The change in this PR is to make MayaHydra compile with usd 26.05.
The main change is in the HVT submodule where I created a branch in the HVT repo https://github.com/Autodesk/hydra-viewport-toolbox/tree/mayaHydra/usd-26.05 . This branch contains some commits to compile with usd 26.05 but updating to the latest from main broke usd 24.11 for maya 2026. So I had to make a branch and cherry-pick some specific commits to make maya hydra compile with usd 26.05 and 24.11. In the future we will switch to the main branch again from HVT when it compiles with usd 24.11 and delete the maya hdyra branch and update the [tag](https://github.com/Autodesk/hydra-viewport-toolbox/releases/tag/v0.26.05.0_hydra_viewport_toolbox) which currently points on the head of the maya hydra branch.

### DIFF
--- a/lib/mayaHydra/mayaPlugin/renderOverride.cpp
+++ b/lib/mayaHydra/mayaPlugin/renderOverride.cpp
@@ -946,6 +946,14 @@ MStatus MtohRenderOverride::Render(
             if (isPass0) {
                 // Do not share the AOVs, for the first pass only
                 HdTaskSharedPtrVector passTasks = currentPass->GetRenderTasks();
+#if PXR_VERSION >= 2605
+                if (_sceneGlobalsSceneIndex) {
+                    const SdfPath& cameraPath = currentPass->params().renderParams.camera;
+                    if (!cameraPath.IsEmpty()) {
+                        _sceneGlobalsSceneIndex->SetPrimaryCameraPrimPath(cameraPath);
+                    }
+                }
+#endif
                 
                 /*Debug code left here if needed later
                 hvt::FramePass& framePassToDebug = *currentPass;
@@ -966,6 +974,14 @@ MStatus MtohRenderOverride::Render(
                             { PXR_NS::HdAovTokens->color, PXR_NS::HdAovTokens->depth }
                     );
                     HdTaskSharedPtrVector passTasks = currentPass->GetRenderTasks(inputAOVs);
+#if PXR_VERSION >= 2605
+                    if (_sceneGlobalsSceneIndex) {
+                        const SdfPath& cameraPath = currentPass->params().renderParams.camera;
+                        if (!cameraPath.IsEmpty()) {
+                            _sceneGlobalsSceneIndex->SetPrimaryCameraPrimPath(cameraPath);
+                        }
+                    }
+#endif
 
                     /*Debug code left here if needed later
                     hvt::FramePass& framePassToDebug = *currentPass;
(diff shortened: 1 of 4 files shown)


lanierd-adsk:
For the preflight, I have only one failure with timeouts in maya 2026 Windows :
The following tests FAILED:
14:46:52      	 26 - testNewSceneWithStage (Timeout)
14:46:52      	 27 - testMayaShadingModes (Timeout)
14:46:52      	 28 - testMayaDisplayLayers (Timeout)
14:46:52      	 29 - testMayaIsolateSelect (Timeout)
14:46:52      Errors while running CTest

Which I think is safe to merge anyway.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -946,6 +946,14 @@ MStatus MtohRenderOverride::Render(
...
                 // Do not share the AOVs, for the first pass only
                 HdTaskSharedPtrVector passTasks = currentPass->GetRenderTasks();
+#if PXR_VERSION >= 2605
+                if (_sceneGlobalsSceneIndex) {
+                    const SdfPath& cameraPath = currentPass->params().renderParams.camera;
+                    if (!cameraPath.IsEmpty()) {
+                        _sceneGlobalsSceneIndex->SetPrimaryCameraPrimPath(cameraPath);
+                    }
+                }
+#endif
```
debloip-adsk, line 956: Why do we need to call `_sceneGlobalsSceneIndex->SetPrimaryCameraPrimPath` again on each frame pass after getting the render tasks, is the call in the `if (_mayaHydraSceneIndex && !camera.isOrtho())` block not sufficient?

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1292,6 +1308,21 @@ MStatus MtohRenderOverride::Render(
...
+                } else if (camera.isOrtho()) {
+                    // Orthographic views use the free camera derived from viewport matrices.
+                    const int numFramePasses = _GetNumFramePasses();
+                    for (int i = 0; i < numFramePasses; ++i) {
+                        const hvt::FramePassPtr& currentPass = _GetFramePass(i);
+                        if (!currentPass) {
+                            continue;
+                        }
+                        currentPass->params().renderParams.camera = SdfPath();
```
debloip-adsk, line 1324: I'm guessing yes, but just to make sure, setting an empty path makes it so Hydra/render delegates falls back to the free camera derived from viewport matrices?

debloip-adsk (approved):
Looks good, thanks! Couple questions I'd like to understand before merging, but nothing really blocking

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lanierd-adsk: No — the call in the if (_mayaHydraSceneIndex && !camera.isOrtho()) block is not enough. The per-pass call after GetRenderTasks() is the one that actually matters. For perspective views, the early call is largely redundant and may even set the wrong path briefly.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lanierd-adsk: Setting renderParams.camera to an empty path doesn’t trigger a Hydra “fallback to free camera” API. Ortho rendering works because we skip SetCameraViewport, apply Maya’s view/projection matrices to all passes, and FramePass::GetRenderTasks() always pushes those matrices into HdxFreeCameraSceneDelegate and sets the task camera to that delegate’s prim (/PassN/camera). The empty assignment is a defensive reset: it clears any Maya scene-index camera path the perspective branch would set, so orthographic views aren’t associated with a scene camera prim. The authoritative camera for rendering and scene globals is established in GetRenderTasks() (and synced to primaryCameraPrim per pass afterward).

debloip-adsk (approved):

PR #459, opened 2026-06-15
Moved path mapping support to MayaUsdProxyShapeSceneIndexBase.

ppt-adsk:
Path mapping used to be just for selection highlighting, to map an application (selection) path into a Hydra path.

It is now used to map an application render settings path into a Hydra render settings path, and therefore can no longer be reserved for interactive use.

Simply moved over all of the path mapping functionality from the derived interactive proxy shape scene index class to the base.  The only functionality left in the derived interactive proxy shape scene index class is for pick handling.

### DIFF
--- a/lib/flowViewport/selection/fvpPathMapperRegistry.cpp
+++ b/lib/flowViewport/selection/fvpPathMapperRegistry.cpp
@@ -152,7 +152,9 @@ PrimSelections PathMapperRegistry::UfePathToPrimSelections(
             else {
                 primSelections = mapper->UfePathToPrimSelections(appPath);
                 if (primSelections.empty()) {
-                    TF_WARN("Mapping for path %s returned no prim path.", Ufe::PathString::string(appPath).c_str());
+                    // Maintenance warning: test usdProxyShapeRenderSettings
+                    // checks that this warning string is not emitted.
+                    TF_WARN("Mapping for application path %s returned no prim path.", Ufe::PathString::string(appPath).c_str());
                 }
             }
         }
--- a/lib/flowViewport/selection/fvpPathMapperRegistry.cpp
+++ b/lib/flowViewport/selection/fvpPathMapperRegistry.cpp
@@ -151,11 +151,8 @@ PrimSelections PathMapperRegistry::UfePathToPrimSelections(
             }
             else {
                 primSelections = mapper->UfePathToPrimSelections(appPath);
-                if (primSelections.empty()) {
-                    // Maintenance warning: test usdProxyShapeRenderSettings
-                    // checks that this warning string is not emitted.
-                    TF_WARN("Mapping for application path %s returned no prim path.", Ufe::PathString::string(appPath).c_str());
-                }
+                TF_VERIFY(!primSelections.empty(),
+                          "Mapping for application path %s returned no prim path.", Ufe::PathString::string(appPath).c_str());
             }
         }
     }
--- a/lib/mayaHydra/hydraExtensions/sceneIndex/mhMayaUsdProxyShapeSceneIndex.cpp
+++ b/lib/mayaHydra/hydraExtensions/sceneIndex/mhMayaUsdProxyShapeSceneIndex.cpp
@@ -19,131 +19,7 @@
 #include <mayaHydraLib/pick/mhPickHandlerRegistry.h>
 #include <mayaHydraLib/pick/mhUsdPickHandler.h>
 
-#include <flowViewport/fvpInstruments.h>
-#include <flowViewport/selection/fvpPathMapperRegistry.h>
-
-// mayaHydra headers
-#include "ufeExtensions/Global.h"
-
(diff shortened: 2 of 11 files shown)


test/lib/cmdLineRender/CMakeLists.txt
ppt-adsk, line 1: Used test-driven development (TDD) and wrote the test first.  Without the fix it fails, with the fix it passes.

test/lib/cmdLineRender/CMakeLists.txt
```diff
@@ -243,6 +243,52 @@ endif()
...
+
+    _mayaHydra_setup_test_common_path_vars()
+    list(APPEND ALL_PATH_VARS MAYA_RENDER_DESC_PATH)
+    _mayaHydra_setup_test_common_defaults("${test_name}")
+    _mayaHydra_setup_test_plugins()
+    list(APPEND MAYAHYDRA_VARNAME_MAYA_RENDER_DESC_PATH
+         "${CMAKE_INSTALL_PREFIX}/renderDesc")
+    list(APPEND MAYAHYDRA_VARNAME_PYTHONPATH "${MAYA_HYDRA_DIR}/scripts")
+    _mayaHydra_setup_test_USD_paths()
+    _mayaHydra_setup_test_finalize_env("${test_name}")
```
ppt-adsk, line 277: AI reproduced the test setup from test.cmake, which I then cleaned up to the minimum.  If more tests use this approach we can factor out the CMake code to test.cmake into a function.

lib/mayaHydra/hydraExtensions/sceneIndex/registration.cpp
ppt-adsk, line 1: Batch base class and interactive derived class now take the same construction arguments.

lib/mayaHydra/hydraExtensions/sceneIndex/mhMayaUsdProxyShapeSceneIndexBase.h
ppt-adsk, line 1: Code moved from base to derived class.

lib/mayaHydra/hydraExtensions/sceneIndex/mhMayaUsdProxyShapeSceneIndexBase.cpp
ppt-adsk, line 1: No changes, simply moved from interactive derived to batch base class.

lib/mayaHydra/hydraExtensions/sceneIndex/mhMayaUsdProxyShapeSceneIndex.h
ppt-adsk, line 1: Moved to base class.

lib/mayaHydra/hydraExtensions/sceneIndex/mhMayaUsdProxyShapeSceneIndex.cpp
ppt-adsk, line 1: Moved to base class.  Only remaining functionality in interactive derived class is pick handling.

lib/flowViewport/selection/fvpPathMapperRegistry.cpp
```diff
@@ -152,7 +152,9 @@ PrimSelections PathMapperRegistry::UfePathToPrimSelections(
             else {
                 primSelections = mapper->UfePathToPrimSelections(appPath);
                 if (primSelections.empty()) {
-                    TF_WARN("Mapping for path %s returned no prim path.", Ufe::PathString::string(appPath).c_str());
+                    // Maintenance warning: test usdProxyShapeRenderSettings
+                    // checks that this warning string is not emitted.
+                    TF_WARN("Mapping for application path %s returned no prim path.", Ufe::PathString::string(appPath).c_str());
```
lanierd-adsk: Is it TF_VERIFY or TF_WARN that you want to write ? 
As there is no check that this warning is not emitted here

lanierd-adsk (commented):
Thanks for doing all this work. Only one question.

test/lib/cmdLineRender/CMakeLists.txt
```diff
@@ -243,6 +243,51 @@ endif()
...
+
+    set_property(TEST "${test_name}" APPEND PROPERTY ENVIRONMENT
+        "MAYA_IGNORE_DIALOGS=1")
+    # Set TF_VERIFY to generate a TF_FATAL_ERROR.
+    # PathMapperRegistry::UfePathToPrimSelections() use TF_VERIFY to check
+    # for path mapper failures, so this will cause the test to fail.
+    set_property(TEST "${test_name}" APPEND PROPERTY ENVIRONMENT
```
ppt-adsk, line 284: With this environment variable a failed TF_VERIFY will crash the test, and all path mapping checks were changed to use TF_VERIFY instead of TF_WARN (one already did use TF_VERIFY).

lanierd-adsk (approved):

PR #460, opened 2026-06-18
HYDRA-2367 : Remove or update outdated comments

lanierd-adsk:


ppt-adsk (approved):
Thanks for the cleanup.

lanierd-adsk:
Merged as the only error in one preflight build was some unrelated exception which look like a glitch.

PR #476, opened 2026-07-28
HYDRA-2275 : Update HVT to recent main

debloip-adsk:
- Ported over and slightly tweaked @julia-b-grenier's changes from [this commit](https://github.com/Autodesk/maya-hydra/commit/3062579eb72803d169cd2469db43886ed4fc264f)
- Made two fixes in HVT : [188](https://github.com/Autodesk/hydra-viewport-toolbox/pull/188) and [189](https://github.com/Autodesk/hydra-viewport-toolbox/pull/189)
- Disable HVT codesign

### DIFF
--- a/lib/mayaHydra/mayaPlugin/renderOverride.cpp
+++ b/lib/mayaHydra/mayaPlugin/renderOverride.cpp
@@ -504,12 +504,17 @@ VtValue MtohRenderOverride::_GetUsedGPUMemory() const
 
 int MtohRenderOverride::GetUsedGPUMemory()
 {
-    int totalGPUMemory = 0;
+    size_t totalGPUMemory = 0;
     std::lock_guard<std::mutex> lock(_allInstancesMutex);
     for (auto* instance : _allInstances) {
-        totalGPUMemory += instance->_GetUsedGPUMemory().UncheckedGet<int>();
+        const VtValue usedGpuMemory = instance->_GetUsedGPUMemory();
+        if (usedGpuMemory.IsHolding<size_t>()) {
+            totalGPUMemory += usedGpuMemory.UncheckedGet<size_t>();
+        } else if (usedGpuMemory.IsHolding<int>()) {
+            totalGPUMemory += usedGpuMemory.UncheckedGet<int>();
+        }
     }
-    return totalGPUMemory / (1024*1024);
+    return static_cast<int>(totalGPUMemory / (1024u * 1024u));
 }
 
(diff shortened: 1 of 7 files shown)


julia-b-grenier (approved):
LGTM!

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1337,55 +1342,20 @@ MStatus MtohRenderOverride::Render(
...
-                    for (int i = 0; i < numFramePasses; ++i) {
-                        const hvt::FramePassPtr& currentPass = _GetFramePass(i);
-                        if (!currentPass) {
-                            continue;
-                        }
-                        currentPass->params().renderParams.camera = SdfPath();
-                    }
-                }
-            }
+    // Leave renderParams.camera empty so HVT renders through its free camera
```
lilike-adsk, line 1345: This sound will break the custom attribute support, if the free camera is used (renderParams.camera was set as empty), the custom camera attributes from Maya will not be applied as the free camera is a different HdCamera item under the hood if I recall correctly.

lilike-adsk: That's why we want the HdCamera with Maya attributes (including custom) to be set to renderParams.camera, we can log another ticket to track the **wrong values** issue if it's hard to solve

debloip-adsk: Hm so I wasn't too involved in the whole custom attributes work, but I have two things in mind : 
- The free camera _is_ different than the renderParams camera, but I also know that the cameraPath we were setting here was never actually used for the actual rendering by HVT, it always used the free camera regardless. So there should be no change in behavior in that regard. Unless setting renderParams.camera was required to do the custom attributes translation itself to Hydra scene index, I think this won't change the existing behavior
- I see in testCustomAttributes.cpp we seem to test the custom attributes translation for cameras, and the preflights passed so I'm assuming it's not breaking it? Unless the unit tests are not testing what you are referring to

lilike-adsk: ok, if HVT always used the free camera, then it's not a problem for this PR. Please log a ticket to track, thanks!

lilike-adsk (approved):

lib/mayaHydra/mayaPlugin/renderOverride.cpp
debloip-adsk: Logged as HYDRA-2485. Thanks!

PR #478, opened 2026-07-29
HYDRA-2438 : Fix random camera position in  testLightingRenderDelegates

lanierd-adsk:


debloip-adsk (approved):

lanierd-adsk:
The preflight is reporting an error but is an internal glitch not related to this PR. @debloip-adsk agrees on merging in this case.

PR #481, opened 2026-07-31
HYDRA-2449 : Fix infinite loop

lanierd-adsk:
Fix infinite loop when rendering an animated scene with production rendering

### DIFF
--- a/lib/mayaHydra/hydraExtensions/adapters/meshAdapter.cpp
+++ b/lib/mayaHydra/hydraExtensions/adapters/meshAdapter.cpp
@@ -21,7 +21,6 @@
 #include <mayaHydraLib/adapters/tokens.h>
 #include <mayaHydraLib/sceneIndex/mayaHydraSceneIndex.h>
 
-#include <pxr/base/arch/hash.h>
 #include <pxr/base/gf/interval.h>
 #include <pxr/base/tf/type.h>
 #include <pxr/imaging/hd/changeTracker.h>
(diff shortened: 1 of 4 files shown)


lib/mayaHydra/hydraExtensions/adapters/meshAdapter.cpp
```diff
@@ -79,71 +78,6 @@ PXR_NAMESPACE_OPEN_SCOPE
 
 namespace {
 
-// Snapshot of mesh connectivity + a hash of point positions used to distinguish UV-only
-// inMesh/pnts dirties (polyEditUV and similar) from true geometry edits. Maya often dirties
```
lanierd-adsk, line 83: This has been removed and we always emit dirtyPoints in this case, we lost the grabular dirtyUVs which we will need to put back  at some point.

benyoon-adsk (approved):
David and I discussed on Zoom. Thanks for the explanation!

lib/mayaHydra/hydraExtensions/adapters/meshAdapter.cpp
```diff
@@ -619,31 +552,24 @@ class MayaHydraMeshAdapter : public MayaHydraShapeAdapter
     }
 
 private:
-    /// inMesh/pnts dirties fire for both geometry edits and UV-only edits. Compare against the
-    /// last captured connectivity/points to emit granular UV locators when only face-varying data
-    /// changed (e.g. polyEditUV when MPolyMessage::addUVSetChangedCallback does not run).
+    /// inMesh/pnts dirties fire for geometry edits, skinning/deformation, and can also be triggered by UV-only edits.
+    /// Do not read mesh data here: NodeDirtiedCallback runs during DG dirty propagation and
```
ppt-adsk, line 556: Understood, can't evaluate during dirty propagation.

ppt-adsk (commented):
Understand why fix was needed and agree with the fix.  Thanks for the test.

PR #482, opened 2026-07-31
HYDRA-2450 : Improve speed of rendering with PRMan render delegate in our viewport unit tests

lanierd-adsk:
When using the PRMan render delegate in our viewport unit tests, this was adding between 10 and 20 mn to the preflight so this PR reduces the time taken to do a PRMan render delegate rendering in the viewport and also fixes when you try to locally run the test as the result was a black rendering while the viewport was correct.


test/lib/mayaUsd/render/mayaToHydra/cpp/testWriteFile.py
```diff
@@ -15,28 +15,12 @@
 
 import maya.cmds as cmds
 import fixturesUtils
+import imageUtils
```
ppt-adsk, line 18:
```suggestion
from imageUtils import WriteFile
```

test/testUtils/imageUtils.py
```diff
@@ -33,8 +32,105 @@
...
+# PRMan/playblast failures where geometry and lighting never reach the AOV.
+SNAPSHOT_MIN_MEAN_LUMINANCE = 0.02
+
+
+# Context manager that routes viewport output through ImageBufferWriter.
+#
+# Playblast reads Maya's model-editor framebuffer. Progressive delegates such as
+# PRMan render into Hydra AOVs instead, so playblast can be blank locally even
+# when the viewport looks correct. ImageBufferWriter reads the same Hydra buffer
+# the override presents. Requires the mayaHydraCppTests plugin (TestWriteFile.*).
```
ppt-adsk, line 45: I don't understand this.  Does this mean that playblast with RenderMan as the renderer (without Hydra) has never worked?

test/testUtils/imageUtils.py
```diff
@@ -33,8 +32,105 @@
...
+        import maya.cmds as cmds
+        cmds.mayaHydraCppTest(f="TestWriteFile.unsetImageSize")
+        cmds.mayaHydraCppTest("", f="TestWriteFile.setFileName")
+        return False
+
+
+# Capture the Hydra viewport color AOV to an image file.
+#
+# Prefer over snapshot()/playblast for progressive Hydra delegates (e.g. PRMan)
+# that render into AOVs rather than the model-editor framebuffer.
```
ppt-adsk, line 84: Does this mean we need to have renderer-specific capture functions?

test/testUtils/imageUtils.py
```diff
@@ -67,6 +163,67 @@ def snapshot(outputPath, width=400, height=None):
...
+    # Older PySide6 builds return a Shiboken buffer without an implicit size;
+    # setsize() must be called before bytes() or the buffer may be empty/short.
+    if hasattr(ptr, "setsize"):
+        ptr.setsize(byteCount)
+    buf = bytes(ptr)[:pixelCount * 4]
+
+    total = 0.0
+    count = 0
+    for r, g, b, a in struct.iter_unpack("4B", buf):
+        if a:
```
ppt-adsk, line 202: So any non-zero alpha is treated the same way?  But if there is a non-zero alpha in the image, the image is probably meaningful?

test/testUtils/imageUtils.py
```diff
@@ -363,14 +522,34 @@ def _artifact_url(abs_path, workspace_dir):
...
+    # Capture-related parameters:
+    #   useHydraWriter -- when True, capture via hydraSnapshot (required for PRMan).
+    #   hydraSettleFn  -- settleFn(captureRefresh) for hydraSnapshot when
+    #                     useHydraWriter is True; must call captureRefresh once.
+    # When useHydraWriter is True, assertImageNonBlank runs on the capture first.
+    #
+    # Usage:
+    #   self.assertSnapshotClose("Storm_pointLight1.png", 0.1, 7.0)
+    #   self.assertSnapshotClose("PRMan_pointLight1.png", 0.2, 10.0,
+    #       useHydraWriter=True, hydraSettleFn=self._settlePrmanBeforeSnapshot)
```
ppt-adsk, line 539: I'm concerned about the renderer-specific parameter.

test/testUtils/renderManUtils.py
```diff
@@ -39,6 +56,17 @@ def setUp():
...
+    Also disables adaptive sampling and caps max samples so viewport tests are
+    less noisy. HD_PRMAN_* env vars must be set before MayaHydraBaseTestCase
+    setUpClass() loads HdPrman: they seed HdPrmanRenderDelegate's render
+    setting descriptors at construction time (see HdPrmanRenderDelegate
+    ctor / HD_PRMAN_MAX_SAMPLES, HD_PRMAN_DISABLE_ADAPTIVE_SAMPLING in
+    hdPrman). Setting the equivalent defaultRenderGlobals.*__convergedSamplesPerPixel
+    attribute afterwards does *not* work as a substitute or complement: Maya
+    only re-applies render-globals attribute changes to the delegate via the
+    "mtohRenderOverride_ApplySetting" AE callback / mayaHydra -updateRenderGlobals
+    command, neither of which fires for a plain cmds.setAttr from a script.
```
ppt-adsk, line 68: Why is HdPrman using defaultRenderGlobals settings?  We should definitely move away from this.  Is this transitional?

test/testUtils/renderManUtils.py
```diff
@@ -51,13 +79,19 @@ def setUp():
 
     os.environ["RMAN_CONFIG_OVERRIDE"] = log_root
     os.environ["RDIR"] = log_root
+    os.environ["HD_PRMAN_DISABLE_ADAPTIVE_SAMPLING"] = "1"
+    os.environ["HD_PRMAN_MAX_SAMPLES"] = str(PRMAN_TEST_MAX_SAMPLES)
```
ppt-adsk, line 83: I would hope there is a Hydra v2 render settings way of passing this information to HdPrman.  Environment variables are hard to use and control, especially if the property value needs to change.

test/testUtils/renderManUtils.py
```diff
@@ -67,6 +101,75 @@ def setUp():
...
+#   timeout_seconds       -- max wait (default PRMAN_TEST_MAX_SETTLE_SECONDS).
+#   poll_interval_seconds -- sleep between polls (default PRMAN_TEST_SETTLE_POLL_INTERVAL).
+#
+# Returns:
+#   True if converged before timeout, False if timed out (caller should capture anyway).
+#
+# Usage:
+#   renderManUtils.waitForInteractiveConvergence()
+#   captureRefresh()
+def waitForInteractiveConvergence(
```
ppt-adsk, line 132: Does HdArnold need this?

test/lib/mayaUsd/render/mayaToHydra/testLightingRenderDelegates.py
```diff
@@ -176,10 +188,14 @@ def _getIntensityForLight(self, lightName, delegate):
...
-        Uses mayaHydraTesting(converged=True). Skips when convergenceTimeout is 0 (e.g. Storm).
-        If the renderer does not report convergence within the timeout (e.g. PRMan in interactive
-        mode), we proceed anyway and take the snapshot."""
+        """Wait for progressive renderers to converge before a playblast snapshot.
+
+        Uses mayaHydraTesting(converged=True). Skips when convergenceTimeout is 0
+        or absent (e.g. Storm, which converges immediately). Not called at all for
+        PRMan, which is captured via hydraSnapshot/hydraSettleFn instead -- see
+        use_hydra_writer in _runLightSnapshots and _settlePrmanBeforeSnapshot.
+        If the renderer does not report convergence within the timeout, proceeds anyway.
```
ppt-adsk, line 197: Renderer-specific logic doesn't sound like a lot of fun for future maintenance.  I wonder what Pixar does for their own tests.

test/testUtils/imageUtils.py
```diff
@@ -363,14 +522,34 @@ def _artifact_url(abs_path, workspace_dir):
...
         if imageVersion is not None:
             snapImagePath = os.path.join(self.getSnapshotDir(), imageVersion, os.path.basename(refImagePath))
         else:
             snapImagePath = os.path.join(self.getSnapshotDir(), os.path.basename(refImagePath))
-        snapshot(snapImagePath)
+        if useHydraWriter:
+            hydraSnapshot(snapImagePath, settleFn=hydraSettleFn)
+            assertImageNonBlank(snapImagePath)
+        else:
+            snapshot(snapImagePath)
```
debloip-adsk, line 553: This is tying the useHydraWriter option to the usage of a settling/convergence function and the non-blank image check, is this something we really want? I feel like all those things should be independent of each other

debloip-adsk (commented):
Thanks for the work, lots of changes!

I am a bit concerned though with the overall approach for handling renderer convergence, I fear that it is overfitting for the specific task of supporting PRMan, instead of a more streamlined approach that would handle Storm, Arnold and PRMan. It seems like we would end up with a lot of code dedicated to handling renderer convergence, e.g. _settlePrmanBeforeSnapshot, convergenceTimeout, _waitForConvergence, hydraSettleFn, renderManUtils.waitForInteractiveConvergence, mayaHydraTesting(converged=True). This is a lot to keep track of; maybe there is some requirement(s) I am missing, but could we not have just one common path for handling convergence, and that always runs (e.g. in assertSnapshotClose), but just defaults the convergence timeout to 0? That way all renderers are treated the same, and non-converging renderers are essentially just renderers with a convergence of 0s.

Another general, more personal comment : I found many of the comments at the top of functions to be a bit heavy to read personally, I feel that in a lot of cases having them interspersed within the code would help readability (when possible obviously). With the current setup I had to do a first read of each block from start to finish, then read the code, and map the right sections of the code to the right sections of the comments, which didn't flow super well. If each sentence of the comments were right beside the code lines where they are relevant, I think it would make the overall reading flow better. We could also turn this into a rule for AI agents when writing code/adding comments, I'm probably going to do this at least locally for myself for now.

We can discuss further on Slack if you prefer

lanierd-adsk:
Seeing the number of comments from this PR I think I was going in the wrong direction. I am closing it and will restart from scratch.

PR #483, opened 2026-07-31
HYDRA-2453 : Fix paths env vars in tests to handle extra USD copy

debloip-adsk:


debloip-adsk:
@vlasovi I also ran a preflight with POST_ZIPS enabled, all builds passed except for the 2026 Windows coverage one, where only two unit tests timed out, which is a known occasional problem. So seems like it works

vlasovi (approved):
Nice work!

PR #484, opened 2026-08-03
HYDRA-2244 : Improve Footprint example about update notifications

lanierd-adsk:
Summary
Refactors the MhFootPrint DataProducer example to keep Hydra prims stable and push attribute changes through targeted Hydra 2.0 dirty locators (transform, extent, vertexColors, purpose) instead of removing and re-adding mesh prims on every update.

Adds Fvp::DirtyNotifier::dirtyPurpose(), shared Maya attribute helpers (GetDouble3AttributeValue, GetGfVec3fAttributeValue, PlugOrChildIsConnected), and documents the recommended DataProducer update pattern in flowViewportToolkit.md.

Handles timeline scrub, Evaluation Manager playback, and setAttr at the current frame without a new key via attributeChanged, preEvaluation, and conditional timeChanged callbacks, with correct MDG context selection per update path.

Test plan
Run testFootPrintNode.py (new scrub/timeline/setAttr snapshot tests) Run testFvpDirtyNotifier.cpp (dirtyPurpose locator test) Manually scrub keyed size/color on MhFootPrint and confirm viewport updates without flicker


lib/mayaHydra/flowViewportAPIExamples/flowViewportAPILocator/mhFlowViewportAPILocator.cpp
```diff
@@ -818,20 +807,14 @@ void MhFlowViewportAPILocator::setCubeGridParametersFromAttributes()
     GetMatrixAttributeValue(mat, mObj, MhFlowViewportAPILocator::mCubeInitialTransform);
     memcpy(_cubeGridParams._initialTransform.GetArray(), mat[0], sizeof(double) * 16);//convert from MMatrix to GfMatrix4d
 
-    double3 color;
-    GetDouble3AttributeValue(color, mObj, MhFlowViewportAPILocator::mCubeColor);
-    _cubeGridParams._color.data()[0]        = color[0];//Implicit conversion from double to float
-    _cubeGridParams._color.data()[1]        = color[1];
-    _cubeGridParams._color.data()[2]        = color[2];
+    _cubeGridParams._color = MAYAHYDRA_NS_DEF::GetGfVec3fAttributeValue(
+        mObj, MhFlowViewportAPILocator::mCubeColor, GfVec3f(0.f, 1.0f, 0.f));
```
ppt-adsk:
```suggestion
    _cubeGridParams._color = MayaHydra::GetGfVec3fAttributeValue(
        mObj, MhFlowViewportAPILocator::mCubeColor, GfVec3f(0.f, 1.0f, 0.f));
```

lib/mayaHydra/flowViewportAPIExamples/flowViewportAPILocator/mhFlowViewportAPILocator.cpp
```diff
@@ -818,20 +807,14 @@ void MhFlowViewportAPILocator::setCubeGridParametersFromAttributes()
...
     GetAttributeValue(_cubeGridParams._opacity, mObj, MhFlowViewportAPILocator::mCubeOpacity);
     GetAttributeValue(_cubeGridParams._useInstancing, mObj, MhFlowViewportAPILocator::mCubesUseInstancing);
 
-    double3 deltaTrans;
-    GetDouble3AttributeValue(deltaTrans, mObj, MhFlowViewportAPILocator::mCubesDeltaTrans);
-    _cubeGridParams._deltaTrans.data()[0]   = deltaTrans[0];//Implicit conversion from double to float
-    _cubeGridParams._deltaTrans.data()[1]   = deltaTrans[1];
-    _cubeGridParams._deltaTrans.data()[2]   = deltaTrans[2];
+    _cubeGridParams._deltaTrans = MAYAHYDRA_NS_DEF::GetGfVec3fAttributeValue(
+        mObj, MhFlowViewportAPILocator::mCubesDeltaTrans, GfVec3f(5.0f, 5.0f, 5.0f));
```
ppt-adsk:
```suggestion
    _cubeGridParams._deltaTrans = MayaHydra::GetGfVec3fAttributeValue(
        mObj, MhFlowViewportAPILocator::mCubesDeltaTrans, GfVec3f(5.0f, 5.0f, 5.0f));
```

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -78,13 +90,68 @@ namespace {
 void nodeAddedToModel(MObject& node, void* clientData);
 void nodeRemovedFromModel(MObject& node, void* clientData);
 
-// Pick handler for the footprint node.
+// Sampled data source that reads live values from caller-owned storage so Hydra
```
ppt-adsk, line 93: What is the lifescope of the caller-owned storage?  The pointer to the value is passed in a as plain pointer.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -176,24 +288,58 @@ class MhFootPrint : public MPxLocatorNode
     /// Heel path to be used in the retained hydra scene index for the heel primitive
     SdfPath                     _heelPath;
 
-    ///Hydra retained scene index to add the 2 foot print primitives
-    HdRetainedSceneIndexRefPtr  _retainedSceneIndex  {nullptr};
+    /// Live mesh state read by the data sources added to _retainedSceneIndex
+    /// (see FootPrintMeshPrimState). Must outlive any Hydra read of those data
+    /// sources, i.e. must be destroyed after _retainedSceneIndex is unregistered
+    /// and destroyed; _retainedSceneIndex is declared last in this class (and so
+    /// destroyed first) specifically to preserve that order.
```
ppt-adsk: Feels fragile.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -176,24 +288,58 @@ class MhFootPrint : public MPxLocatorNode
...
+    MTime  _cachedSyncTime;
+    /// True right after a kAttributeSet sync on the size/color/purpose plug (resp.),
+    /// at the same timeline time, until EM's next preEvaluation pass consumes it.
+    /// Guards against preEvaluation's dirty-plug branch immediately re-reading the
+    /// plug through the EM context: if that plug is still connected (e.g. to an
+    /// animCurve), that re-read would evaluate through the connection and silently
+    /// discard a setAttr override made on the driven plug without a new key.
+    bool   _sizeManualOverridePendingPreEval = false;
+    bool   _colorManualOverridePendingPreEval = false;
+    bool   _purposeManualOverridePendingPreEval = false;
```
ppt-adsk: I am puzzled about Evaluation Manager involvement with dirty plug notifications, and needing to guard against it.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -293,157 +441,58 @@ namespace
...
-                footPrint->updateFootPrintPrims();
+        if (attributeMessageAffectsFootPrintAttributes(msg)) {
+            auto* footPrint = reinterpret_cast<MhFootPrint*>(footPrintData);
+            // kAttributeSet: read the plug from the datablock (fsNormal).
+            // Other messages: read the plug at the current timeline time.
+            if (msg & MNodeMessage::kAttributeSet) {
+                footPrint->syncHydraFromNodeFromAttributeSet(&plug);
+            } else {
+                footPrint->syncHydraFromNodeAtCurrentTime(&plug);
+            }
```
ppt-adsk, line 482: Is this the usual way to look for attribute value changes?

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -293,157 +441,58 @@ namespace
...
+            footPrint->_UpdateTimeChangedCallback();
```
ppt-adsk, line 484: Why do we need to update the time changed callback here?

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -477,34 +527,79 @@ MString	MhFootPrint::nodeClassification("hydraAPIExample/geometry/footPrint");
...
-        //Trigger a call to compute so that everything is initialized
-        MhFootPrint* footPrintInstance = reinterpret_cast<MhFootPrint*>(clientData);
-        footPrintInstance->updateFootPrintPrims();
-        // No need to call footPrintInstance->addedToModelCb(), as reading the
-        // file will add the node to the model.
+        auto* footPrint = reinterpret_cast<MhFootPrint*>(clientData);
+        {
+            MDGContextGuard guard(MDGContext::fsNormal);
+            footPrint->syncHydraFromNode(/*changedPlug=*/nullptr);
+        }
```
ppt-adsk, line 542: Why do we need to set the DG context?

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -514,64 +609,326 @@ void MhFootPrint::postConstructor()
...
+{
+    state.xform.SetIdentity();
+    state.xform.SetScale(GfVec3d(size, size, size));
+    state.extentMin = GfVec3d(corner1.x * size, corner1.y * size, corner1.z * size);
+    state.extentMax = GfVec3d(corner2.x * size, corner2.y * size, corner2.z * size);
 }
 
-// Retrieve value of the size attribute from the node
+// Read Maya attrs and push size/color/purpose changes to Hydra.
+void MhFootPrint::syncHydraFromNode(const MPlug* changedPlug)
```
ppt-adsk, line 790: How can this be called with a null plug pointer?

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -514,64 +609,326 @@ void MhFootPrint::postConstructor()
...
+bool MhFootPrint::_HasAnyConnectedTimelineAttribute() const
+{
+    const MObject obj = thisMObject();
+    const MPlug   sizePlug(obj, mSize);
+    const MPlug   colorPlug(obj, mColor);
+    const MPlug   purposePlug(obj, mDrawAsSecondaryGraphics);
+    return MAYAHYDRA_NS_DEF::PlugOrChildIsConnected(sizePlug)
+        || MAYAHYDRA_NS_DEF::PlugOrChildIsConnected(colorPlug)
+        || MAYAHYDRA_NS_DEF::PlugOrChildIsConnected(purposePlug);
+}
```
ppt-adsk:
```suggestion
    return MayaHydra::PlugOrChildIsConnected(sizePlug)
        || MayaHydra::PlugOrChildIsConnected(colorPlug)
        || MayaHydra::PlugOrChildIsConnected(purposePlug);
}
```

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -514,64 +609,326 @@ void MhFootPrint::postConstructor()
...
+
+    const MTime currentTime = MAnimControl::currentTime();
+    if (_cachedSyncTimeValid && _cachedSyncTime == currentTime) {
+        return;
+    }
+
+    syncHydraFromNodeAtCurrentTime(/*changedPlug=*/nullptr);
+}
+
+// Register or remove the timeChanged callback when display attrs are connected.
```
ppt-adsk: Why is this necessary?

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -603,25 +961,119 @@ bool MhFootPrint::_GetDrawAsSecondaryGraphics() const
...
+    return _GetDrawAsSecondaryGraphics()
+        ? Fvp::secondaryGraphicsRenderTagToken
+        : HdRenderTagTokens->geometry;
+}
+
+// Read the color attribute as a GfVec3f.
 GfVec3f MhFootPrint::_GetColor() const
 {
+    return MAYAHYDRA_NS_DEF::GetGfVec3fAttributeValue(
+        thisMObject(), MhFootPrint::mColor, GfVec3f(0.f, 0.f, 1.f));
```
ppt-adsk:
```suggestion
    return MayaHydra::GetGfVec3fAttributeValue(
        thisMObject(), MhFootPrint::mColor, GfVec3f(0.f, 0.f, 1.f));
```

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -603,25 +961,119 @@ bool MhFootPrint::_GetDrawAsSecondaryGraphics() const
...
+{
+    // preEvaluation can be invoked for MDGContexts that do not correspond to
+    // the timeline's actual current time, e.g. Maya internally evaluating
+    // other frames while recomputing anim curve tangents when a new key is
+    // added. Treating such a call as a real "current time changed" event would
+    // read size/color/purpose at that other frame and push it to Hydra,
+    // clobbering the value actually being displayed (regression: setting a key
+    // at the current frame with the already-correct value could reset the
+    // displayed size back to a value keyed at another frame). So bail out
+    // early for any non-normal context whose time isn't the current time.
```
ppt-adsk: Why do we push these values to Hydra?  Why don't we simply mark them as dirty and Hydra will pull them?

ppt-adsk (commented):
Some questions about interactions with the EM.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: The caller-owned storage is the FootPrintMeshPrimState fields on MhFootPrint (_solePrimState / _heelPrimState). Each _MutableTypedSampledDataSource holds a non-owning pointer to one of those members (created in _AddFootPrintPrim). The data sources themselves live inside _retainedSceneIndex until the node is torn down.

Hydra may read through those pointers only while the scene index is registered with the viewport. removedFromModelCb() and ~MhFootPrint() both call removeDataProducerSceneIndex() before the node/state is destroyed. _retainedSceneIndex is also declared last so it is destroyed before the prim state members as an extra safeguard.

Updates write into those fields in place, so the stored pointer stays valid for the node's lifetime. This is an example-only pattern — not a general-purpose data source.

I also updated the comment to state this.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: Fair point — the declaration-order comment overstates the real guarantee. The actual safety comes from explicitly calling removeDataProducerSceneIndex in both removedFromModelCb() and ~MhFootPrint() before any member teardown, which stops Hydra from reading those data sources. The "declared last" ordering was meant as defense in depth only. I'll make teardown explicit in the destructor (_retainedSceneIndex.Reset() after unregister) so we don't depend on field order, and trim the comments accordingly.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: EM isn’t the source of the edit here, it’s the second notifier for the same edit.

When the user setAttrs a connected display plug (e.g. size with an animCurve), we get:

kAttributeSet on our attribute-changed callback → we sync from MDGContext::fsNormal (datablock override) and push the correct value to Hydra.
The same setAttr marks the plug dirty in EM’s evaluation graph → EM calls preEvaluation() on this node for the current frame → evaluationNode.dirtyPlugExists(mSize) is true.
Without the guard, step 2 would immediately call syncHydraFromNode(&sizePlug), which reads the plug in EM’s MDG context and evaluates through the connection, discarding the manual override we just applied in step 1. That’s the bug fixed by test_keyedSizeTimelineUpdatesViewport (setAttrNoKey snapshot) and test_keyedSizeThenAddKeyAtCurrentFrameDoesNotResetSize.

We still need the preEvaluation dirty-plug branch for the other case: anim/DG-driven updates during EM playback where we don’t get a useful kAttributeSet on this node. The *ManualOverridePendingPreEval flags only suppress the duplicate resync when we’ve just handled the same plug via syncHydraFromNodeFromAttributeSet.

I've expanded the comment.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: For Flow Viewport locator examples, yes — addAttributeChangedCallback and reading the plug in the callback is the usual pattern. MhFlowViewportAPILocator and MhCustomShaders do exactly that: per-plug reads (plug.asInt(), asDouble(), etc.) and push into the data producer, with no message filtering and a no-op preEvaluation.

FootPrint follows the same callback entry point but adds context selection because its display attrs (size/color/purpose) are tested with animation and connections — cases the locator examples don't cover. On kAttributeSet we read from the datablock (fsNormal) so a manual override on a connected plug is honored; for other messages we read at the current timeline time. We also add preEvaluation and conditional timeChanged for EM/scrub paths that the locator doesn't need.

So: same Flow Viewport mechanism as MhFlowViewportAPILocator, with extra MDG-context handling where keyed/connected display attrs require it. Happy to add a comment pointing back to the locator as the simple baseline if that helps.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: timeChanged is only registered while a display attr has an incoming connection (animCurve, driven key, etc.) — see _HasAnyConnectedTimelineAttribute(). We need it for scrub/playback when the value changes with the timeline but we don't get another useful kAttributeSet on this node.

_UpdateTimeChangedCallback() here refreshes that registration when connection state may have changed — e.g. first setKeyframe connects an animCurve, deleting keys or disconnectAttr removes it. Without this, we'd either miss timeline updates after keying, or keep a timeChanged listener active when attrs are static.

Same idea as imagePlaneMaterialAdapter's conditional timeChanged registration; we also call it from addedToModelCb and afterOpen for initial state.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: syncHydraFromNode() reads size/color/purpose via MPlug::getValue(), which is context-sensitive. The kAfterOpen callback doesn’t run in a well-defined evaluation context, so we set MDGContext::fsNormal to read the saved datablock values that were just restored from the file, rather than evaluating through connections/anim curves at whatever context happens to be active. This matches syncHydraFromNodeFromAttributeSet() and avoids pushing stale or frame-wrong values to Hydra after open. Timeline-driven attrs are handled afterward via _UpdateTimeChangedCallback() / onTimeChanged().

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: changedPlug is optional by design. nullptr means “sync all footprint attributes (size, color, purpose).” We use that for full-resync paths: scene open, timeline scrub, frame change in preEvaluation, and initial registration in addedToModelCb. When we know which attribute changed (attribute callback or a single dirty plug in preEvaluation), we pass &plug so we don’t re-read unrelated attrs in the wrong MDG context — that was the original bug this PR fixes. The !changedPlug || … guards ensure we never dereference a null pointer.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: MhFootPrint manually pushes size/color/purpose into a Hydra retained scene index. When those display attrs are connected (anim curves, driven keys, or other time-varying DG inputs), scrubbing the timeline changes the evaluated values, but Maya does not reliably fire attributeChanged on this node, and preEvaluation may not run either (locator with no compute is not always dirtied on scrub-only updates).

The timeChanged callback is the fallback that re-reads connected attrs at MAnimControl::currentTime() and syncs Hydra on each frame change (onTimeChanged → syncHydraFromNodeAtCurrentTime).

We register it only when _HasAnyConnectedTimelineAttribute() is true so we do not pay a global timeChanged callback on every scrub when attrs are static (static edits still come through attributeChanged / setAttr). _UpdateTimeChangedCallback() is called when connections may change (attribute add/remove/connect, node added to model, after file open) so the callback is added/removed at the right time.

Covered by test_keyedSizeTimelineUpdatesViewport and test_keyedSizeAndColorTimelineUpdatesViewport.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
lanierd-adsk: You're right, we are dirtying them only, I changed the comments in the file to avoid that wrong terminology.

lib/mayaHydra/flowViewportAPIExamples/footPrintNode/mhFootPrintNode.cpp
```diff
@@ -903,7 +908,9 @@ void MhFootPrint::onTimeChanged()
     syncHydraFromNodeAtCurrentTime(/*changedPlug=*/nullptr);
 }
 
-// Register or remove the timeChanged callback when display attrs are connected.
+// Register timeChanged only while display attrs are connected to time-varying
+// sources. Scrubbing does not always dirty this locator or fire attributeChanged,
+// so we re-sync Hydra from MAnimControl::currentTime() on each frame change.
```
ppt-adsk, line 913: For an example this is O.K.  For real production code, this does not feel right.  If an attribute is animated, but does not change values over a range of frames, we should not dirty Hydra.

ppt-adsk (approved):

PR #485, opened 2026-08-03
HYDRA-2429 : Pass OCIO configuration file path to Hydra render delegate

benyoon-adsk:
To allow proper interpretation of color management color space names and other color management information, both the Hydra renderer and the application must refer to the same OCIO configuration file. Pass this information to render delegate's settingsMap via SetRenderSetting(). No unit test will be added as per the discussion with Pierre.

### DIFF
--- a/lib/mayaHydra/mayaPlugin/batchRendering/tokens.h
+++ b/lib/mayaHydra/mayaPlugin/batchRendering/tokens.h
@@ -47,7 +47,8 @@ PXR_NAMESPACE_OPEN_SCOPE
 #define BATCH_RENDER_TOKENS \
     (renderSettingsSrc) \
     (settingsMapRenderSettingsSrc) \
-    (hydraSceneRenderSettingsSrc)
+    (hydraSceneRenderSettingsSrc) \
+    (adsk_ocioPath)  // Autodesk-specific OCIO config path (no standard Hydra token).
 // clang-format on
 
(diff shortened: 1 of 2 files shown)


lib/mayaHydra/mayaPlugin/batchRendering/tokens.h
```diff
@@ -47,7 +47,8 @@ PXR_NAMESPACE_OPEN_SCOPE
 #define BATCH_RENDER_TOKENS \
     (renderSettingsSrc) \
     (settingsMapRenderSettingsSrc) \
-    (hydraSceneRenderSettingsSrc)
+    (hydraSceneRenderSettingsSrc) \
+    (adsk_ocioPath)  // Autodesk-specific OCIO config path (no standard Hydra token).
```
ppt-adsk: Since this is purely our convention, and the risk of collision is probably nil, I wonder if could just call it `ocioConfigPath`.  And I think it should be `ocioConfigPath` rather than `ocioPath`.

benyoon-adsk: I agree, will change it to that.

benyoon-adsk:
Preflight passed, Windows 2027 coverage build failed from timeouts. Re-ran build and passed, see Branch Preflight - Coverage - 2027 (Windows) #731.

ppt-adsk (approved):

PR #486, opened 2026-08-07
HYDRA-2460 : Add PRMan production render test coverage

lanierd-adsk:
Adds HdPrmanLoaderRendererPlugin as a third renderer variant (alongside Storm and Arnold) across the cmdLineRender production test suite — texturing, render settings, relative-path, and USD-proxy-shape tests all now run under PRMan on Windows via a new _mayaHydra_append_prman_production_render_env CMake helper that isolates the RenderMan config dir and drives the render pass through Hydra V2 render settings (HD_PRMAN_RENDER_SETTINGS_DRIVE_RENDER_PASS=1), except for one dedicated test that explicitly exercises the Hydra V1 render-settings path.

Reorganizes test/lib/cmdLineRender/scenes/{texturing,renderSettings,relativePathTest} into per-renderer subfolders (storm/, arnold/, prman/) so each renderer's scene + baseline images live together. Fixes output.py's setOutputFormat to guard on defaultArnoldDriver existing, since PRMan-only scenes don't have an Arnold driver node. Refactors the parallel renderer-variant lists in test/lib/cmdLineRender/CMakeLists.txt into small registration macros to reduce risk of the per-renderer columns drifting out of sync, and simplifies a redundant platform condition. No production (non-test) code changes; all changes are test infrastructure and CMake.


lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndexUtils.h
```diff
@@ -78,9 +78,16 @@ _ToMaterialNetworkSchema(
...
 
-                // Colorspace metadata
+                // Colorspace metadata. Maya may publish colorSpace:<param> as a
+                // string (e.g. file.cs OCIO label) rather than TfToken; accept
+                // both so downstream render delegates can read it as a token.
                 if (res.second) {
-                    paramsInfo[res.first].colorSpace = p.second.Get<TfToken>();
+                    if (p.second.IsHolding<TfToken>()) {
+                        paramsInfo[res.first].colorSpace = p.second.Get<TfToken>();
+                    } else if (p.second.IsHolding<std::string>()) {
```
ppt-adsk, line 87: Are we compensating for a bug in translation earlier on?  How did the Maya colorSpace:<param> end up in the HdMaterialNode as a string in the first place, rather than as a token?

I understand this is forgiving and therefore useful, but my feeling is that this isn't addressing the root cause.

build.py
```diff
@@ -235,6 +235,32 @@ def BuildVariant(context):
         return "RelWithDebInfo"
     return "RelWithDebInfo"
 
+def MergeCtestExcludeRegex(ctestArgs, filesToExclude):
```
lanierd-adsk, line 238: There was a bug in merging the exclusion of the tests from our maya hydra internal repo and the file tests-to-run.json from this repo. This is the fix for that

scripts/mayaHydra/renderSettings/output.py
```diff
@@ -52,7 +52,8 @@ def setRenderProductName(productName):
 def setOutputFormat(of):
     # Set the Maya render setting.
     mel.eval('setMayaSoftwareImageFormat("' + of + '")')
-    cmds.setAttr('defaultArnoldDriver.aiTranslator', of, type='string')
+    if cmds.objExists('defaultArnoldDriver'):
```
ppt-adsk, line 55: The danger with performing renderer-specific work in a general context.  I guess this is still manageable for now, but if we keep on adding renderers from Autodesk or third parties we will need to re-evaluate.

test/lib/cmdLineRender/scenes/basic/usdCamera.ma
ppt-adsk, line 1: This won't work for HdArnold, which requires a RenderVar named "color".  As per
https://openusd.org/release/user_guides/schemas/usdRender/RenderVar.html
`The name of the RenderVar prim drives the name of the data variable that the renderer will produce.`

test/lib/cmdLineRender/scenes/renderSettings/arnold/arnoldImageName.usda
ppt-adsk, line 1: Why was this change necessary?

test/lib/cmdLineRender/scenes/renderSettings/arnold/arnoldOutputFormat.usda
ppt-adsk, line 1: Why was this change necessary?

test/lib/cmdLineRender/scenes/renderSettings/arnold/arnoldRenderDir.usda
ppt-adsk, line 1: Why was this change necessary?

test/lib/cmdLineRender/scenes/renderSettings/arnold/arnoldResolution.usda
ppt-adsk, line 1: Why the directory change?

test/lib/cmdLineRender/scenes/renderSettings/arnold/usdProxyShapeRenderSettings.usda
ppt-adsk, line 1: Why the directory change?

test/lib/cmdLineRender/scenes/renderSettings/storm/stormCropRegion.usda
ppt-adsk, line 1: Unsure about the directory change.

test/lib/cmdLineRender/scenes/renderSettings/storm/stormCropRegion1x1.usda
ppt-adsk, line 1: Directory change.

test/lib/cmdLineRender/scenes/renderSettings/storm/stormImageName.usda
ppt-adsk, line 1: Directory change.

test/lib/cmdLineRender/scenes/renderSettings/storm/stormNoCrop.usda
ppt-adsk, line 1: Directory change.

test/lib/cmdLineRender/scenes/renderSettings/storm/stormOutputFormat.usda
ppt-adsk, line 1: Why was this change needed?

test/lib/cmdLineRender/scenes/renderSettings/storm/stormRenderDir.usda
ppt-adsk, line 1: Directory change.

test/lib/cmdLineRender/scenes/renderSettings/storm/stormResolution.usda
ppt-adsk, line 1: Directory change.

test/lib/cmdLineRender/CMakeLists.txt
ppt-adsk, line 1: I will need some help reviewing this.

build.py
ppt-adsk, line 238: Why does this happen?  Why do we have two separate mechanisms to reconcile?

ppt-adsk (commented):
The separation into renderer-specific directories is nice, but I fear it means that you need to add the same test in 3 places.  In actual fact you do need to add the same test in 3 places, but I wonder if the renderer-specific directories would be subdirectories of the individual tests.  We can discuss this, each organization has its pros and cons.

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndexUtils.h
lanierd-adsk: You're right, I am fixing the root cause.

scripts/mayaHydra/renderSettings/output.py
lanierd-adsk: Right, I've added a comment and modified the code a bit so it's more explicit it's renderer specifc

test/lib/cmdLineRender/scenes/basic/usdCamera.ma
lanierd-adsk: Good catch. A single shared scene cannot use one beauty RenderVar for both Arnold V2 and PRMan V2 — on Hydra V2 the delegate reads the USD prim name directly (color vs Ci). I'll revert usdCamera.ma to color for Storm/Arnold and add a PRMan-specific copy with Ci for the PRMan V2 test, matching the pattern already used under scenes/renderSettings/{arnold,prman}/.

test/lib/cmdLineRender/scenes/renderSettings/arnold/arnoldImageName.usda
lanierd-adsk: Because the structure of the folders has changed and everything stays in the arnold folder now

test/lib/cmdLineRender/scenes/renderSettings/arnold/arnoldOutputFormat.usda
lanierd-adsk: Because the structure of the folders has changed and everything stays in the arnold folder now

test/lib/cmdLineRender/scenes/renderSettings/arnold/arnoldRenderDir.usda
lanierd-adsk: Because the structure of the folders has changed and everything stays in the arnold folder now

test/lib/cmdLineRender/scenes/renderSettings/arnold/arnoldResolution.usda
lanierd-adsk: Because the structure of the folders has changed and everything stays in the arnold folder now

test/lib/cmdLineRender/scenes/renderSettings/arnold/usdProxyShapeRenderSettings.usda
lanierd-adsk: Because the structure of the folders has changed and everything stays in the arnold folder now

test/lib/cmdLineRender/scenes/renderSettings/storm/stormCropRegion.usda
lanierd-adsk: Because the structure of the folders has changed and everything stays in the storm folder now

test/lib/cmdLineRender/scenes/renderSettings/storm/stormCropRegion1x1.usda
lanierd-adsk: Because the structure of the folders has changed and everything stays in the storm folder now

test/lib/cmdLineRender/scenes/renderSettings/storm/stormImageName.usda
lanierd-adsk: Because the structure of the folders has changed and everything stays in the storm folder now

test/lib/cmdLineRender/scenes/renderSettings/storm/stormNoCrop.usda
lanierd-adsk: Because the structure of the folders has changed and everything stays in the storm folder now

test/lib/cmdLineRender/scenes/renderSettings/storm/stormOutputFormat.usda
lanierd-adsk: Because the structure of the folders has changed and everything stays in the storm folder now

test/lib/cmdLineRender/scenes/renderSettings/storm/stormRenderDir.usda
lanierd-adsk: Because the structure of the folders has changed and everything stays in the storm folder now

test/lib/cmdLineRender/scenes/renderSettings/storm/stormResolution.usda
lanierd-adsk: Because the structure of the folders has changed and everything stays in the storm folder now

build.py
lanierd-adsk: We have this mechanism in the open source repo and another in ecg maya hydra and ecg maya hydra was overwriting this.

lanierd-adsk:
@copilot resolve the merge conflicts in this pull request

test/lib/cmdLineRender/CMakeLists.txt
lanierd-adsk: ## What these CMake helpers do

### The fridge chart

The build script keeps a **chart of render tests**. Each row is one test setup:

| Folder | Renderer | Suffix | Extra label | Image size |
|--------|----------|--------|-------------|------------|
| storm  | Storm    | Storm  | *(nothing)* | 960×540    |
| arnold | Arnold   | Arnold | *(nothing)* | 160×120    |
| prman  | PRMan    | PRMan  | needs PRMan plugin | 160×120 |

Storm and Arnold don't need anything special. PRMan adds a label like `|depOnPlugins:prman` so the test only runs when that plugin is available.

---

### Problem 1: CMake eats empty list entries

CMake lists drop empty elements by default (`CMP0007 OLD`). So we can't store a real empty string in a list reliably.

**Workaround:** use `"."` as a placeholder for "no dependency label", then convert it back when reading:

```cmake
function(_mayaHydra_normalize_cmd_line_render_dep_label dep_label_var)
    if(${dep_label_var} STREQUAL ".")
        set(${dep_label_var} "" PARENT_SCOPE)
    endif()
endfunction()
```

That's why Storm/Arnold register with `"."`, while PRMan registers with `"|depOnPlugins:prman"`.

---

### Problem 2: Parallel lists must stay in sync

Each test variant is stored across several parallel lists (`_tex_folders`, `_tex_renderers`, `_tex_suffixes`, `_tex_dep_labels`, `_tex_res_args`, etc.).

If you `list(APPEND ...)` each column separately, it's easy to add a row to one list and forget another — then index `i` no longer describes one coherent test.

**Fix:** one macro call registers a full row at once:

```cmake
macro(_mayaHydra_register_texturing_variant folder renderer suffix dep_label res_arg)
    list(APPEND _tex_folders ${folder})
    list(APPEND _tex_renderers ${renderer})
    list(APPEND _tex_suffixes ${suffix})
    list(APPEND _tex_dep_labels ${dep_label})
    list(APPEND _tex_res_args ${res_arg})
endmacro()
```

Example:

```cmake
_mayaHydra_register_texturing_variant(storm HdStormRendererPlugin Storm "." "-x 960 -y 540")
```

`_mayaHydra_register_render_settings_variant` is the same pattern for render-settings tests, with one extra column (`new_image_name`).

---

### What happens later

The `foreach` loop walks those lists row by row:

1. Read folder, renderer, suffix, dep label, resolution args (and image name for render-settings tests).
2. Normalize `"."` → `""` via `_mayaHydra_normalize_cmd_line_render_dep_label`.
3. Register the actual cmd-line render tests.

---

### One-line summary

Keep a tidy table of per-renderer test variants; use `"."` when a dependency label is empty (because CMake drops blank list entries); and register each row in one macro call so the parallel lists can't drift out of sync.

test/lib/cmdLineRender/RenderPRMan_UsdCamera.bat
```diff
@@ -0,0 +1,34 @@
...
+set TF_DEBUG=HDPRMAN_* MAYAHYDRAPLUGIN_BATCHRENDER_RENDER_SETTINGS
+set TF_DEBUG_EMIT_CODE_NAMES_WITH_MESSAGES=1
+set TF_DEBUG_OUTPUT_FILE=stderr
+
+if not exist "%OUT_ROOT%" mkdir "%OUT_ROOT%"
+
+echo Rendering to: %OUT_FILE%
+echo.
+
+"%MAYA_BIN%\mayabatch.exe" -command "Render -renderer mayaHydra -rfn \"HdPrmanLoaderRendererPlugin\" -x 960 -y 540 -cam \"|stage1|stageShape1,/camera1\" -rd \"%OUT_ROOT%\" -im usdCamera -of png \"%SCENE_SRC%\""
```
ppt-adsk: I don't understand this line.  mayabatch is used to execute a MEL script, not a shell command line, yet the argument to the `command` flag is a shell command line that executes the Render executable.  I don't understand what the `mayaHydra` renderer is either, and I don't know what the `-rfn` flag is.

test/lib/cmdLineRender/CMakeLists.txt
ppt-adsk: In the "fridge chart" presentation, there are 3 almost identical columns.  Why not name the folder with exactly the same name as the renderer, if this is convenient?  Also, having an "extra label" field just for PRMan feels out of place.  Could we not have a single PRMan conditional somewhere, instead of affecting all other tests?  This would avoid the "CMake eats empty list entries" problem.

ppt-adsk: In the "Parallel lists must stay in sync" section, why do we keep parallel lists for a single test variant?  We seem to be creating a problem for ourselves that I'm not sure I understand.

test/lib/cmdLineRender/CMakeLists.txt
```diff
@@ -28,29 +59,82 @@ if(NOT IS_MACOSX)
...
+    _mayaHydra_register_texturing_variant(prman HdPrmanLoaderRendererPlugin PRMan "|depOnPlugins:prman" "-x 160 -y 120")
+endif()
+
+set(_tex_scene_stems
+    mayaSphereMayaMaterial
+    usdSphereMaterialX
+    usdSphereUsdMaterial
+)
+
+list(LENGTH _tex_folders _tex_count)
```
ppt-adsk: This all seems quite complex, and still has a PRMan conditional.  What is this loop doing?

test/lib/cmdLineRender/CMakeLists.txt
```diff
@@ -63,31 +147,224 @@ foreach(scene ${STORM_RENDER_SCENES})
...
-            "USDIMAGINGGL_ENGINE_ENABLE_SCENE_INDEX=1"
-            "TF_DEBUG=MAYAHYDRAPLUGIN_BATCHRENDER_RENDER_SETTINGS"
+# Render settings production tests. Scene + baseline live under renderSettings/<renderer>/.
+_mayaHydra_register_render_settings_variant(storm HdStormRendererPlugin Storm "." "-x 720 -y 480" aNewStormImageName)
+_mayaHydra_register_render_settings_variant(arnold HdArnoldRendererPlugin Arnold "." "-x 720 -y 128" aNewArnoldImageName)
+if(IS_WINDOWS)
+    _mayaHydra_register_render_settings_variant(prman HdPrmanLoaderRendererPlugin PRMan "|depOnPlugins:prman" "-x 720 -y 128" aNewPRManImageName)
+endif()
+
+list(LENGTH _rs_folders _rs_count)
```
ppt-adsk: Same complexity question as for the texturing tests, and again a PRMan conditional.

test/lib/cmdLineRender/CMakeLists.txt
ppt-adsk, line 1: I'm sure the intent of the changes was to factor out common CMake code for tests with multiple renderers.  However, I feel the result is very hard to understand and quite wordy.  I would like to discuss if we can improve things.

ppt-adsk (commented):
Some questions about the batch file and the test CMakeLists.txt changes.

test/lib/cmdLineRender/CMakeLists.txt
lanierd-adsk: Thanks. I simplified this code using your comments

test/lib/cmdLineRender/RenderPRMan_UsdCamera.bat
lanierd-adsk: Sorry, it's not a file that is supposed to be added. It was a local file. I am removing it.

test/lib/cmdLineRender/CMakeLists.txt
```diff
@@ -28,29 +71,56 @@ if(NOT IS_MACOSX)
...
+    usdSphereMaterialX
+    usdSphereUsdMaterial
+)
+
+function(_mayaHydra_add_texturing_variant folder res_arg)
+    _mayaHydra_renderer_for_folder(${folder} renderer suffix)
+    _mayaHydra_dep_label_for_folder(${folder} dep_label)
+
+    set(stems ${_tex_scene_stems})
     if(MAYA_APP_VERSION VERSION_GREATER 2026)
```
ppt-adsk: This is no longer necessary because of test/lib/CMakeLists.txt.

test/lib/cmdLineRender/CMakeLists.txt
```diff
@@ -28,29 +71,56 @@ if(NOT IS_MACOSX)
...
-        list(APPEND STORM_RENDER_SCENES texturing/mayaSphereMaterialXStorm.ma)
+        list(APPEND stems mayaSphereMaterialX)
     endif()
-endif()
 
+    foreach(stem ${stems})
+        set(scene "texturing/${folder}/${stem}${suffix}.ma")
+        cmake_path(GET scene STEM sceneFileNameNoExt)
+        if(folder STREQUAL "prman")
+            set(render_dir ${CMAKE_BINARY_DIR}/test/Temporary/${sceneFileNameNoExt}/projects/default/images)
```
ppt-adsk: I wonder why this line is necessary?

test/lib/cmdLineRender/CMakeLists.txt
```diff
@@ -63,33 +133,219 @@ foreach(scene ${STORM_RENDER_SCENES})
...
+            RENDERED_IMAGE_NAME ${sceneFileNameNoExt}_v2_rs
+            RENDERER_ARGS "-x 160 -y 120 -rd \"${render_dir}\" -im ${sceneFileNameNoExt}_v2_rs"
+            TEST_NAME_SUFFIX v2_rs
+            ENV
+                "MAYA_HYDRA_HD_ARNOLD_HYDRA_V2_RENDER_SETTINGS=1"
+                "USDIMAGINGGL_ENGINE_ENABLE_SCENE_INDEX=1"
+                "TF_DEBUG=MAYAHYDRAPLUGIN_BATCHRENDER_RENDER_SETTINGS"
+        )
+    endforeach()
+    if(MAYA_APP_VERSION VERSION_GREATER 2026)
```
ppt-adsk: This is no longer necessary because of test/lib/CMakeLists.txt.

test/lib/cmdLineRender/CMakeLists.txt
```diff
@@ -63,33 +133,219 @@ foreach(scene ${STORM_RENDER_SCENES})
...
+function(_mayaHydra_add_render_settings_variant folder res_arg new_image_name)
+    _mayaHydra_renderer_for_folder(${folder} renderer suffix)
+    _mayaHydra_dep_label_for_folder(${folder} dep_label)
+
+    set(animcube_scene "renderSettings/${folder}/AnimCubeRenderSettings.ma")
+    set(singlecam_scene "renderSettings/${folder}/AnimCubeRenderSettingsSingleCam.ma")
+
+    set(scene "renderSettings/${folder}/${folder}Resolution.ma")
+    if(folder STREQUAL "prman")
+        set(render_dir ${TEST_OUTPUT_ROOT}/${folder}Resolution/projects/default/images)
```
ppt-adsk: Is this line really needed?

test/lib/cmdLineRender/CMakeLists.txt
```diff
@@ -63,33 +133,219 @@ foreach(scene ${STORM_RENDER_SCENES})
...
+        mayaHydra_add_cmd_line_render_test(${scene}${dep_label}
+            COPY_SCENE
+            RENDERER ${renderer}
+            RENDERER_ARGS "${res_arg} -rd ../images -im ${folder}Resolution"
+        )
+    endif()
+
+    set(scene "renderSettings/${folder}/${folder}RenderDir.ma")
+    if(folder STREQUAL "prman")
+        set(render_dir ${TEST_OUTPUT_ROOT}/${folder}RenderDir/projects/default/images/aSubDirectory)
```
ppt-adsk: Is this line really needed?

test/lib/cmdLineRender/CMakeLists.txt
```diff
@@ -63,33 +133,219 @@ foreach(scene ${STORM_RENDER_SCENES})
...
+            COPY_SCENE
+            RENDERER ${renderer}
+            RENDERED_IMAGE_SUBDIR projects/default/images/aSubDirectory
+            RENDERER_ARGS "${res_arg} -rd ../images/aSubDirectory"
+        )
+    endif()
+
+    set(scene "renderSettings/${folder}/${folder}ImageName.ma")
+    if(folder STREQUAL "prman")
+        set(render_dir ${TEST_OUTPUT_ROOT}/${folder}ImageName/projects/default/images)
```
ppt-adsk: Again, unclear to me why HdPrman would render to a different directory, that would feel like a bug to me.

test/lib/cmdLineRender/CMakeLists.txt
```diff
@@ -63,33 +133,219 @@ foreach(scene ${STORM_RENDER_SCENES})
...
+            COPY_SCENE
+            RENDERER ${renderer}
+            RENDERED_IMAGE_NAME ${new_image_name}
+            RENDERER_ARGS "${res_arg} -rd ../images -im ${new_image_name}"
+        )
+    endif()
+
+    set(scene "renderSettings/${folder}/${folder}OutputFormat.ma")
+    if(folder STREQUAL "prman")
+        set(render_dir ${TEST_OUTPUT_ROOT}/${folder}OutputFormat/projects/default/images)
```
ppt-adsk: Another HdPrman render directory conditional that I'm wondering about.

test/lib/cmdLineRender/CMakeLists.txt
```diff
@@ -356,32 +494,32 @@ function(_add_usdProxyShapeRenderSettings_test)
     _mayaHydra_setup_test_USD_paths()
     _mayaHydra_setup_test_finalize_env("${test_name}")
 
+    if("${renderer}" STREQUAL "HdPrmanLoaderRendererPlugin")
+        _mayaHydra_append_prman_production_render_env("${test_name}")
+    endif()
+
     set_property(TEST "${test_name}" APPEND PROPERTY ENVIRONMENT
         "MAYA_IGNORE_DIALOGS=1")
-    # Set TF_VERIFY to generate a TF_FATAL_ERROR.
```
ppt-adsk, line 386: Why was this comment removed?

ppt-adsk (commented):
Looking much better now, thanks for the changes!  Still have questions about the HdPrman conditional code that feels out of place.

test/lib/cmdLineRender/CMakeLists.txt
lanierd-adsk: Good catch !

test/lib/cmdLineRender/CMakeLists.txt
lanierd-adsk: PRMan needs an absolute -rd here. Storm/Arnold can use -rd ../images because that resolves relative to the copied project layout, but PRMan (Hydra V2 render settings) resolves relative paths against the workspace root. 
Without the absolute path, the image lands outside ${CMAKE_BINARY_DIR}/test/Temporary/<test>/projects/default/images, which is where the test harness compares it. The line mirrors the path the helper already uses for idiff; the helper doesn't pass -rd automatically.

lanierd-adsk: I've added a comment in the file.

test/lib/cmdLineRender/CMakeLists.txt
lanierd-adsk: Same reason as the texturing PRMan branch: PRMan resolves relative -rd against the workspace root, not the copied project/scene dir. Storm/Arnold's -rd ../images would write outside the temp dir idiff checks. This absolute path matches ${TEST_OUTPUT_ROOT}/prmanResolution/projects/default/images, which is the same location the test helper uses for comparison. The helper computes that path for idiff but does not inject it into RENDERER_ARGS.

test/lib/cmdLineRender/CMakeLists.txt
lanierd-adsk: Yes. This test renders into images/aSubDirectory; Storm/Arnold can use -rd ../images/aSubDirectory because that resolves relative to the copied project layout. PRMan resolves relative -rd against the workspace root, so we need the absolute path that matches RENDERED_IMAGE_SUBDIR and the file(MAKE_DIRECTORY) call. Without it, the image lands outside the temp dir idiff checks.

test/lib/cmdLineRender/CMakeLists.txt
lanierd-adsk: PRMan isn't intentionally rendering to a different directory. hdPrman passes the USD render product's productName straight through to Riley (displayDesc.name = product.name) with no path rewriting, so the output location is whatever path Hydra/maya-hydra put in productName.

In cmdLineRender, -rd is handled in maya-hydra (setRenderDirectory), which rewrites USD productName to Path(rd)/basename(productName) and also sets Maya workspace file rules. For Storm/Arnold, relative -rd ../images resolves correctly against the copied projects/default/scenes → projects/default/images layout. For PRMan/RenderMan, a relative path in productName is resolved against the Maya workspace root, which batch Render.exe doesn't set to that copied project root (we already work around this in interactive tests via _sceneWorkspaceContext in testLightingRenderDelegates.py).

The PRMan branch here uses an absolute -rd so the rewritten productName is absolute and idiff finds the file under test/Temporary/<test>/projects/default/images/. With the same absolute path, PRMan writes where we expect — same as Storm/Arnold, just a different way of expressing the directory in batch mode.

I agree this feels like an integration bug rather than intentional delegate behavior; a cleaner fix would be to canonicalize -rd (or set workspace root) in maya-hydra batch rendering so all delegates can share -rd ../images. Happy to follow up with that separately if useful.

test/lib/cmdLineRender/CMakeLists.txt
lanierd-adsk: Please see above.

test/lib/cmdLineRender/CMakeLists.txt
lanierd-adsk: Removed unintentionally during the HYDRA-2460 refactor when generalizing _add_usdProxyShapeRenderSettings_test() for PRMan. The TF_FATAL_VERIFY=1 setting is unchanged — I'll restore the comment.

lanierd-adsk:
@copilot resolve the merge conflicts in this pull request

lanierd-adsk:
@copilot resolve the merge conflicts in this pull request

ppt-adsk (changes requested):
As discussed separately, we should not be setting a full path through the -im flag, rather through the -rd flag.

Also it would be nice if the new helper were in test.cmake.

lanierd-adsk:
> As discussed separately, we should not be setting a full path through the -im flag, rather through the -rd flag.

So as far as I can find, -im is already never given a full path, and the rule the comment states is already honored everywhere. 

> Also it would be nice if the new helper were in test.cmake.

I am working on it. Thanks !

ppt-adsk (dismissed):

ppt-adsk (approved):

PR #487, opened 2026-08-11
HYDRA-2418 - Cache wireframe color in UpdateRenderItems

vlasovi:
Use the new dirty flag introduced in Maya to update wireframe color only when it's changed.


lilike-adsk (approved):
Nice work

PR #488, opened 2026-08-12
HYDRA-2484 : Register Hydra-based renderers with -cap/cpv flags

benyoon-adsk:
We need to describe that a renderer registered with the renderer command is Hydra-based. This will allow clients to filter on Hydra-based renderers only, as this is a requirement for USD Render Setup.

This is done with the -capability / -capabilityValue flags (e.g. `renderer -capability isHydra -capabilityValue true HdArnoldRendererPlugin`) and USD render setup can query renderers with `renderer -q -capability isHydra HdArnoldRendererPlugin`.

### DIFF
--- a/lib/mayaHydra/mayaPlugin/pluginUtils.cpp
+++ b/lib/mayaHydra/mayaPlugin/pluginUtils.cpp
@@ -207,6 +207,9 @@ bool registerRenderer(const MtohRendererDescription& desc)
            << " -showBatchRenderProcedure " << dq << "batchRender -showImage true" << dq
            << " -renderSequenceProcedure mayaRenderSequence"
            << " -supportColorManagement true"
+           // Need these to list Hydra-based renderers in USD Render Setup.
+           << " -capability isHydra"
+           << " -capabilityValue true"
            // To avoid breaking the render settings UI we add the Common tab,
            // even though we don't use it.
            << " -addGlobalsTab Common createMayaSoftwareCommonGlobalsTab updateMayaSoftwareCommonGlobalsTab"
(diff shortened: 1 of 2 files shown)


benyoon-adsk:
Need to wait for Maya rundev yaml file update to run preflight.

lib/mayaHydra/mayaPlugin/pluginUtils.cpp
```diff
@@ -207,6 +207,9 @@ bool registerRenderer(const MtohRendererDescription& desc)
            << " -showBatchRenderProcedure " << dq << "batchRender -showImage true" << dq
            << " -renderSequenceProcedure mayaRenderSequence"
            << " -supportColorManagement true"
+           // Need these to list Hydra-based renderers in USD Render Setup.
```
ppt-adsk:
```suggestion
           // Need these to list Hydra-based renderers.
```

lib/mayaHydra/mayaPlugin/pluginUtils.cpp
```diff
@@ -207,6 +207,9 @@ bool registerRenderer(const MtohRendererDescription& desc)
...
+           << " -capability isHydra"
+           << " -capabilityValue true"
```
ppt-adsk: Style comment, up to you if you like this or not, I would have this in a single line, since it's a single concept.  If you'd rather leave it as two lines that's fine as well.

ppt-adsk (changes requested):
Needs a test.  Probably do something like
test/lib/cmdLineRender/registeredMayaRenderers.py
Doesn't need an interactive Maya.

ppt-adsk (approved):
Thanks for the test!

PR #489, opened 2026-08-14
HYDRA-2483 - Move dg_access_mutex down to where it's actually needed

vlasovi:
The current placement of dg_access_mutex is too high in the stack. In case where many rprims updated, this prevents full multithreading of  Hydra's SyncAll. This change moves the mutex down to the methods that actually access Maya DG.

### DIFF
--- a/lib/mayaHydra/hydraExtensions/adapters/imagePlaneMaterialAdapter.cpp
+++ b/lib/mayaHydra/hydraExtensions/adapters/imagePlaneMaterialAdapter.cpp
@@ -166,7 +166,11 @@ void MayaHydraImagePlaneMaterialAdapter::CreateCallbacks()
 
 VtValue MayaHydraImagePlaneMaterialAdapter::GetMaterialResource()
 {
-    std::string imagePath = GetImagePlaneTexturePath(GetNode());
+    std::string imagePath;
+    {
+        MayaHydra::DgAccessLock dgLock;
+        imagePath = GetImagePlaneTexturePath(GetNode());
+    }
     if (imagePath.empty()) {
         return GetPreviewMaterialResource(GetID());
     }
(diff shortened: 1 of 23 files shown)


lilike-adsk (approved):
Nice, we can continue to add lock if there's any missing place

lib/mayaHydra/hydraExtensions/adapters/imagePlaneMaterialAdapter.cpp
```diff
@@ -166,7 +166,11 @@ void MayaHydraImagePlaneMaterialAdapter::CreateCallbacks()
 
 VtValue MayaHydraImagePlaneMaterialAdapter::GetMaterialResource()
 {
-    std::string imagePath = GetImagePlaneTexturePath(GetNode());
+    std::string imagePath;
+    {
+        MayaHydra::DgAccessLock dgLock;
+        imagePath = GetImagePlaneTexturePath(GetNode());
```
ppt-adsk, line 172: Why is the lock not in GetImagePlaneTexturePath() itself?  Is it because of the call to GetNode()?

lib/mayaHydra/hydraExtensions/adapters/shapeAdapter.cpp
```diff
@@ -72,11 +72,14 @@ MObject MayaHydraShapeAdapter::GetMaterial(const MObject& shadingComp)
             "Called MayaHydraShapeAdapter::GetMaterial() - %s\n",
             GetDagPath().partialPathName().asChar());
 
+    MayaHydra::DgAccessLock dgLock;
```
ppt-adsk, line 75: Should the lock be inside MayaHydra::FindShadingEngine()?

ppt-adsk (approved):
LGTM, just a couple of minor questions.

lib/mayaHydra/hydraExtensions/adapters/imagePlaneMaterialAdapter.cpp
vlasovi: The implementation of GetImagePlaneTexturePath is outside of the adapter, in mixedUtils.cpp and I wanted to keep the lock inside the adapter classes as much as possible.

lib/mayaHydra/hydraExtensions/adapters/shapeAdapter.cpp
vlasovi: The implementation of FindShadingEngine is outside of the adapter, in mixedUtils.cpp and I wanted to keep the lock inside the adapter classes as much as possible.

PR #490, opened 2026-08-17
HYDRA-2488 - Selection highlight state for USD cameras only changes when changing viewport

vlasovi:
Custom render items (Ufe cameras) don't send the dirty wireframe color notification but they change wireframe color by resetting their effect so we catch it this way.


vlasovi:
This PR seems to be in a wrong state due to the GitHub outage. I will close this one and will open a new one.

PR #493, opened 2026-08-19
HYDRA-2402 - Support the Outline Selection Mode and Hover Highlighting in Hydra Storm

julia-b-grenier:
## Context

[HYDRA-2402 - Support the Outline Selection Mode and Hover Highlighting in Hydra Storm](https://autodesk.atlassian.net/browse/HYDRA-2402)

The outline selection highlighting has been migrated to hvt and Maya-Hydra Storm renderer now offers the option to use it as the selection highlighting instead of the legacy selection highlighting(wireframe).

[HYDRA-2403 - Update and expand Maya-Hydra test coverage for outline highlighting mode](https://autodesk.atlassian.net/browse/HYDRA-2403)

The tests are adapted to run in both configurations(wireframe and outline) 

https://github.com/user-attachments/assets/81487f2b-ca92-4170-a7d1-38c541088a6c

## Key changes

- Bump the hydra-viewport-toolbox submodule to HVT main, which is where the OutlineManager lands
- Add the menu option to select the selection highlighting mode. Outline is the default, except on USD 24.11 where only legacy is offered
- Use the maya color preferences for the outline of the object
- Add the menu option to enable hover outline highlighting, by default it is disabled. The performance impact still needs more evaluation and consideration to be on by default.
- Suppress the legacy wireframe highlight only where it would compete with the outline: it is hidden in shaded and textured modes, and left untouched when the viewport already draws wireframes, so wireframe-on-shaded still shows both cues
- Stop translating a wireframe item to Hydra when it would only ever be hidden(like in outline mode), which removes work at scene import where Maya leaves everything selected
- Add `doc/outlineSelectionHighlightingArchitecture.md`, a companion to the existing `doc/selectionHighlightingArchitecture.md`. It covers the HVT outline pipeline and where it installs, and how the added-geometry highlight is disabled for each data type — which is not the same mechanism for USD prims and Maya-native prims. The existing doc is cross-linked and its claim that pixel-based highlighting is unimplemented is corrected
- Register every interactive test to run in both selection-highlight modes (outline and legacy), and add a new test covering hover highlighting, previously untested

### Fixed issues
- Marquee selection and group deselect no longer leave stale colours on non-lead objects. Only the lead object was being invalidated before
- Colour preference edits now apply immediately, dormant colour included, instead of appearing only after the next selection change
- A colour preference that cannot be resolved now falls back to a defined colour and warns once, instead of producing an arbitrary one. Batch mode has no colour translator at all, so every lookup there took that path
- Ensure that in other modes like bounding box in hydra storm, the selected object color would update properly

Video of bug with bounding box that was fixed. The selected bounding boxed were not colored correctly:
https://github.com/user-attachments/assets/c5ea0fb1-64c7-4222-91bb-11658fad07f9

## Notes

- Outline is Storm only. Other render delegates keep the legacy wireframe highlighting
- On USD 24.11 the outline mode is not offered: HgiGL corrupts the prim ids the outline shader samples, so the menu lists legacy only
- The submodule bump also carries five unrelated HVT commits
- Hover highlighting is built only when the Maya devkit ships Qt. Without Qt the plugin builds and runs, minus hover
- mayaHydraEnableDefaultOutlines exposes an HVT outline feature: a faint outline on every prim in the scene, which shows internal edges. It is off by default because its whole-scene prim-id pass runs every frame and its cost scales with the scene rather than with the selection, which is too expensive to impose in Maya. It is kept available so it can be turned on by command if it turns out to be wanted, and is deliberately absent from the option box
- Confirmed against a clean build of the base commit, so not caused by this PR and logged as separate ticket: issues in the Flat Shade Selected Items display styles, USD geometry vanishing when handing back to VP2

## Per-file details of the implementation changes

`lib/flowViewport/.../fvpMeshWhSi.cpp` — comment-only: renamed scene-index reference.

`lib/hydra-viewport-toolbox` — submodule bump to HVT main, which is where `Outline::OutlineManager` lives.

`adapters/renderItemAdapter.{h,cpp}` — the VP2-MRenderItem → Hydra-rprim translator. Adds one flag `_wireframeSelectionHighlightEnabled`; `GetVisible()` returns false when it's off, and the setter dirties visibility. So an already-translated wire item can be hidden without touching anything VP2 sent.

`sceneIndex/mayaHydraSceneIndex.{h,cpp}` — the Maya→Hydra scene index that consumes the `MViewportScene` snapshot. Adds `RenderItemUpdateOptions` (is legacy highlight on / does the viewport draw wireframes / reconsider-skipped pass) to `UpdateRenderItems`, plus:
- `isWireframeItemReplacedByOutline()` — name allowlist (`DormantPolyWire`, `DormantIsoparmWire`) + shape displayStatus is kActive/kLead.
- new items matching that are skipped entirely (Storm still syncs invisible rprims, so hiding costs a full VB/topology/GPU upload); already-translated ones are hidden via the adapter instead.
- `RefreshRenderItemLegacyHighlightTreatment()` — re-evaluates every existing adapter, needed because Maya doesn't re-send unchanged items on a display-style change.
- the reconsider pass re-reads unflagged items from the snapshot to recover previously-skipped wires, synthesizing visibility bits since Maya sent none.
- drive-by: cleaned up the duplicate-fastId removal logic.

`mhDirtyLeadObjectSceneIndex.*` → `mhDirtySelectionColorsSceneIndex.*` (+ `sceneIndex/CMakeLists.txt`, `mhLeadObjectPathTracker.{h,cpp}`, `renderOverride.h`) — rename + widened role: it used to invalidate only lead-object colors; now also `dirtySelectionRelatedPrims()` for any prims whose selection state changed (fixes stale colors on marquee select / group deselect). Tracker also exposes `getLeadObjectPrimSelections()` so the outline can get the lead path.

`mhWireframeColorInterfaceImp.{h,cpp}` — colors were read once in the ctor; now `RefreshColors()` re-reads them so a Maya color-preference change repaints.

`mixedUtils.cpp` — `getPreferencesColor()` was returning uninitialized floats when `getColor()` failed (batch mode has no translator). Now seeded + warns once.

`mayaPlugin/tokens.h`, `renderGlobals.{h,cpp}` — three new globals: `SelectionHighlightMode` (Outline/Legacy, Legacy on USD 24.11), `OutlineHoverHighlighting` (off by default), `EnableDefaultOutlines` (HVT's whole-scene faint outlines; off by default, script-only, absent from the option box).

`mayaPlugin/hoverEventFilter.{h,cpp}` + `mayaPlugin/CMakeLists.txt` — new passive Qt event filter on the viewport widget, feeding cursor position for hover highlighting. Guarded by `MAYAHYDRA_HAS_QT` since the devkit doesn't always ship Qt.

`doc/outlineSelectionHighlightingArchitecture.md` — new companion to `doc/selectionHighlightingArchitecture.md`: behaviour, mode selection and its Storm / USD 24.11 gating, a brief tour of the HVT outline pipeline and where it installs in the frame pass, the two HVT contracts the design is built around (render-thread-only, frame pass outlives the manager), how the added-geometry highlight is disabled per data type, hover, colours, and limitations. `doc/selectionHighlightingArchitecture.md` gains cross-links, and its "pixel-based highlighting is currently unimplemented" claim is corrected.

`mayaPlugin/renderOverride.{h,cpp}` — the bulk. Installs OutlineManager on Storm pass 0 before colorCorrectionTask; pushes selection/lead/hover paths from `Render()` (manager is render-thread-only, so observers only set atomic dirty flags); builds outline style from Maya color prefs; per-panel HoverState with a 1×1 HdxPickTask resolve and view-proj-matrix comparison to re-resolve when the view moves; mode switch triggers `ClearHydraResources` + `ogs -reset`; computes `renderItemOptions` from the display style; does the selection-delta color invalidation via `set_symmetric_difference`.

## Summary of the tests update and expansion

Every interactive test now runs in both selection-highlight modes (outline, the
plugin's default, and a _legacy counterpart), so both shipping modes stay covered.
A test only needs its own reference image per mode where its output actually
depends on the mode: MayaHydraBaseTestCase resolves a "mode" subfolder when a
test supplies one and falls back to the shared image otherwise, so a mode-
independent test needs no entry in any list and no second set of images - and if
it turns out to be mode-sensitive after all, one of its two runs fails against the
shared image instead of silently losing coverage.

| Test | Change | Why |
|---|---|---|
| testGeomSubsetsWireframeHighlight, testNativeInstancingWireframeHighlight, testPointInstancingWireframeHighlight | Legacy-only | Outline doesn't support geometry subsets / single instances yet (HYDRA-2458 / HYDRA-2456) |
| cpp/testSceneIndexDirtying | Legacy-only | Outline's lazy wireframe-item translation makes toggling wireframe-on-shaded add a prim instead of just dirtying one (HYDRA-2457) |
| cpp/testMaterialBlocking | Legacy-only, permanent | Asserts a legacy highlight prim's material binding; outline never creates that prim |
| cpp/testPassFiltering | Split | The 2 assertions on legacy highlight prims moved to new cpp/testLegacyHighlightPassRouting (legacy-only, permanent); the remaining 7 pass-routing assertions now run in both modes |
| testMeshes, testNamespaces | Mode-aware assertions | Legacy adds a wireframe highlight rprim that outline does not |
| testOutlineHover (new) | New | Covers hover, the one visual state nothing else tests: hover alone, hover layered over lead/non-lead selection, hover on an already-selected object, a negative control, and reverse occlusion |
| 15 other tests (BasicRenderTest, DataProducerSelHighlightTest, FootPrintNodeTest, FramePassesTest, IsolateSelectWithUsdLightingTest, MayaDisplayLayersTest, MayaIsolateSelectTest, MayaUsdAPIUsageTest, MeshWireframeHighlightTest, NurbsPrimitivesTest, PolygonPrimitivesTest, StageInstanceablePrimsSelHighlightTest, StageVariantsTest, TransformsTest, cpp/HydraGenerativeProceduralTest) | Reference images added | Minted outline/ images for whichever existing assertions actually differ by mode; everything else stays on the shared image |

## Bug found when toggling Texture mode on MaterialX materials

Fix black surface when toggling Textured mode on MaterialX materials

Toggling the viewport's Textured mode on a USD mesh with a MaterialX
material rendered the surface solid black under Storm. This broke
testTexturedMode on Maya 2026 / USD 24.11 (Linux, Windows and code
coverage preflights).

Storm re-resolves a mesh's primvar requirements when the rprim
re-initialises its repr. PruneTexturesSceneIndex::MarkTexturesDirty()
only dirtied the material and primvars locators, which is enough for a
UsdPreviewSurface network but not for a MaterialX one: the surface
shader is code-generated and reaches its UV set through a geompropvalue
node, so the restored texture connections sampled an "st" primvar the
mesh no longer provided.

The missing invalidation used to arrive by accident.
ReprSelectorSceneIndex::SetReprType() fired on any display-style change
(including the Textured bit) and dirties every prim with
{displayStyle, primvars, purpose, visibility, xform}. Making that push
conditional on the repr actually changing (a correct change: the scene
index is shared by every panel, and the unconditional push was a
whole-scene dirty per style change) removed the side effect the texture
toggle had been relying on.

Request HdLegacyDisplayStyleSchema's locator in MarkTexturesDirty() so
the toggle is self-sufficient rather than dependent on an unrelated
scene index firing on the same frame.

Co-Authored-By: Claude Opus 5 (1M context) [noreply@anthropic.com](mailto:noreply@anthropic.com)

## Bug found when capturing performance profiling metrics

[HYDRA-2540](https://autodesk.atlassian.net/browse/HYDRA-2540) - mhFlowViewportAPILocator: UFE handler mishandles and warns on foreign paths, slowing scene population

Ticket contains confirmed patch to address both issues.

## Bug found when deleting prims after capturing performance profiling metrics

[EMSUSD-4127](https://autodesk.atlassian.net/browse/EMSUSD-4127) - Redefinition of Hydra's smoothHull repr crashes outline feature in Maya Hydra

Ticket contains confirmed patch to address issue in Maya USD.

## Bug found in the outline implementation in HVT out of scope for this story

Found while testing `testFramePasses`: a selected object rendered with a non-default Hydra render
tag/purpose never gets an outline cue, regardless of selection. Neither `OutlinePrimIdsTask` nor
`OutlineManager` overrides `GetRenderTags()`, so both fall back to `HdxTask`'s default, which only
covers the standard "geometry" tag. Reproduced with a custom `MhFootPrint` node left in
`secondaryGraphics` purpose (from an earlier, unrelated assertion in the same test) — legacy
highlights it fine, since it creates its own separate highlight geometry rather than sampling the
original prim's render tag. Confirmed this isn't a selection-resolution issue: the prim's path
resolves correctly, and the outline appears as soon as the object is put back in its default purpose.

This is a gap in the `hydra-viewport-toolbox` submodule (not this repository), so fixing it means
passing through additional render tags in HVT's outline tasks.

This will be fixed in HYDRA-2456, where other HVT changes are required, see [KBSH-1385](https://autodesk.atlassian.net/browse/KBSH-1385).

## Performance Testing

Performance testing was conducted and documented in [Performance Analysis - Outline in Maya-Hydra](https://autodesk.atlassian.net/wiki/x/LypkPQ).

- The results indicate that the outline feature is often significantly more performant than legacy wireframe highlighting.
- These tests can no longer be run directly because the temporary instrumentation and variables used for performance testing were removed during cleanup; however Alain has restored the required code locally.
- Hover highlighting performance is not acceptable due to the per-mouse-move invocation of GPU hit testing; there are better alternatives to consider for this.


julia-b-grenier:
There have been issues found with the tests thanks to the pre-flight. They are being actively worked on.

julia-b-grenier:
The tests for linux and windows are green, MacOS tests need further investigation. https://c12-cot.jenkins.autodesk.com/job/EMS/job/ECG/job/ecg-maya-hydra/job/preflight/job/2026-usd2505-osx/705/

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -522,9 +591,39 @@ void MayaHydraSceneIndex::UpdateRenderItems(const MDataServerOperation::MViewpor
...
+
+        // Maya sent no visibility bits for an unchanged item, so UpdateFromDelta would leave the new
+        // adapter at its default of invisible. Take the visibility from the item itself instead.
+        if (unchanged) {
+            flags |= MDataServerOperation::MViewportScene::MVS_changedVisibility;
+            if (ri.isEnabled()) {
+                flags |= MDataServerOperation::MViewportScene::MVS_visible;
+            }
+        }
+
```
laferra-adsk, line 626: Reported by Claude; I could reproduce the issue locally:

## 1. Missing transform / material on a first-time-translated highlight wire

**File:** `lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp` (`UpdateRenderItems`, ~L563-693)
**Severity:** High — indeterminate `GfMatrix4d` read into a Hydra data source.

### Mechanism

`UpdateRenderItems` now skips translating a selection-highlight wire that the outline replaces:

```cpp
// mayaHydraSceneIndex.cpp:596-616
const bool replacedByOutline
    = isWireframeItemReplacedByOutline(riName.asChar(), ri.sourceDagPath(), options);

int                           fastId = ri.InternalObjectId();
MayaHydraRenderItemAdapterPtr ria = nullptr;
const bool isNewRenderitem = !_GetRenderItem(fastId, ria);

if (unchanged && !isNewRenderitem) {
    continue;
}

if (isNewRenderitem && replacedByOutline) {
    continue;          // <-- adapter creation is DEFERRED to some later frame
}
```

Deferring creation is the problem. The adapter is now built on whatever *delta* frame first
makes `replacedByOutline` false, and that delta's flags decide what gets initialized:

```cpp
// mayaHydraSceneIndex.cpp:658  material
if (flags & MVS::MVS_changedEffect) { ... ria->SetMaterial(material); }

// mayaHydraSceneIndex.cpp:686  transform
if (flags & MVS::MVS_changedMatrix) { ria->UpdateTransform(ri); }
```

Neither bit is guaranteed on that frame:

* On the **reconsider** path the flags are synthesized locally and contain only visibility bits:

  ```cpp
  // mayaHydraSceneIndex.cpp:617-624
  if (unchanged) {
      flags |= MVS::MVS_changedVisibility;
      if (ri.isEnabled()) { flags |= MVS::MVS_visible; }
  }
  ```

* On the **deselect** path (the one the comment at `renderOverride.cpp:1453` relies on) VP2
  typically sends `MVS_changedVisibility` / `MVS_changedEffect` but has no reason to re-send
  `MVS_changedMatrix` for an object that did not move.

`_transform` is never initialized anywhere else:

```cpp
// renderItemAdapter.h:222
GfMatrix4d _transform[2];          // no initializer

// renderItemAdapter.cpp:100-117 — ctor init list does not mention _transform
// renderItemAdapter.cpp:130-143 — UpdateTransform() is the ONLY writer
```

`GfMatrix4d`'s default constructor leaves the value unspecified, and `UpdateFromDelta`
(`renderItemAdapter.cpp:194+`) touches `_positions`, `_bounds`, primvars and visibility — never
`_transform`. The value is read straight into the prim's xform data source:

```cpp
// mayaHydraDataSource.cpp:209-215
else if (name == HdXformSchemaTokens->xform) {
    auto xform = _adapter->GetTransform();      // returns _transform[0]
    return HdXformSchema::Builder()
        .SetMatrix(HdRetainedTypedSampledDataSource<GfMatrix4d>::New(xform))
        .Build();
}
```

Once wrong, it stays wrong: nothing re-runs `UpdateTransform` until Maya actually sets
`MVS_changedMatrix`, i.e. until the object is moved.

### Why it is hard to reproduce

`isWireframeItemReplacedByOutline` only returns true for a **selected** shape:

```cpp
// mayaHydraSceneIndex.cpp:190-195
const MHWRender::DisplayStatus displayStatus
    = MHWRender::MGeometryUtilities::displayStatus(shapeDagPath);
return (MHWRender::kActive == displayStatus) || (MHWRender::kLead == displayStatus);
```

So the skip — and therefore the bug — requires the object to already be selected **the first
time its `DormantPolyWire` reaches the scene index**. If the object was ever visible to
`UpdateRenderItems` while dormant, the adapter was already created normally (with
`MVS_new | MVS_changedMatrix`) and is simply reused forever after; `isNewRenderitem` is false and
nothing goes wrong. **Selecting an already-translated object and then switching display style
will not reproduce it** — this is most likely what the first repro attempt did.

Second reason the symptom is easy to miss: `std::make_shared` on a fresh heap page usually
yields zero-filled memory, so `_transform[0]` is typically the **zero matrix**, not random
values. A zero matrix collapses the wire to a point — it looks like "the wire is missing", not
"the wire is in the wrong place".

### Reproduction steps (deterministic)

Environment: Storm (`mayaHydra`), USD ≥ 24.11, `mayaHydraSelectionHighlightMode = Outline Selection`.

**Variant A — reconsider path (`reconsiderSkippedHighlightWires`)**

1. Set the viewport to **Smooth Shade All only** — no wireframe-on-shaded, no "Shade Active
   Only". This makes `viewportDrawsWireframes(currentDisplayStyle) == false`
   (`renderOverride.cpp:229-235` only counts `kWireFrame | kShadeActiveOnly`).
2. Switch the renderer to `mayaHydra` and confirm Outline mode is active.
3. Create a poly cube **and leave it selected**, off the origin:

   ```
   polyCube -w 2 -h 2 -d 2;
   move -a 10 5 -3;
   ```

   Creation leaves the cube selected, so its `DormantPolyWire` is skipped on its very first
   frame — this is the precondition that is easy to miss.
   *Equivalent and more explicit:* select any existing mesh, then run `ogs -reset;` so all
   adapters are dropped and every item is re-sent while the shape is selected.
4. Do **not** deselect. Switch that same panel to **Wireframe** (hotkey `4`, or
   `modelEditor -e -displayAppearance wireframe modelPanel4;`).
   → `!viewportDrawsWireframes(_oldDisplayStyle) && viewportDrawsWireframes(current)` is true,
   `reconsiderSkippedHighlightWires` is set, and the adapter is created from a flags-0 item.
5. Observed: the cube's wireframe is drawn at the origin / collapsed to a point instead of at
   `(10, 5, -3)`, and the prim has no material binding. The outline itself is still correct,
   which makes it read as "the wireframe is simply absent".
6. It does not self-heal — re-selecting, deselecting or switching styles again will not fix it.
   `move -r 0.001 0 0;` on the cube (which finally sets `MVS_changedMatrix`) does.

**Variant B — deselect path, no display-style change**

1. Steps 1-3 above.
2. Click empty space to deselect. VP2 re-sends `DormantPolyWire` with `MVS_changedVisibility`
   (+ possibly `MVS_changedEffect`) but no `MVS_changedMatrix`; the adapter is created here with
   the same defect.
3. Switch to Wireframe — the wire becomes visible at the wrong transform.

**Confirming it directly (recommended, avoids relying on the visual)**

* Open the Hydra scene browser (`plugin/mayaHydraSceneBrowser`) and inspect the `xform/matrix`
  data source of the `.../DormantPolyWire` prim after step 4 — compare against the shape's
  actual world matrix.
* Or add a temporary check right after `_AddRenderItem(ria)`:

  ```cpp
  TF_VERIFY(flags & MDataServerOperation::MViewportScene::MVS_changedMatrix,
            "New render item adapter '%s' created without a transform.", riName.asChar());
  ```

  To make the garbage obvious rather than zero-filled, churn the heap first (create and delete a
  few hundred objects) before step 3.

### Proposed fix

Do not let a deferred first translation inherit a thin delta. Force full initialization whenever
the adapter is actually new.

```cpp
// mayaHydraSceneIndex.cpp — after the `unchanged` visibility block (~L624),
// before the `if (isNewRenderitem) { ... }` construction block (~L626)

         if (unchanged) {
             flags |= MDataServerOperation::MViewportScene::MVS_changedVisibility;
             if (ri.isEnabled()) {
                 flags |= MDataServerOperation::MViewportScene::MVS_visible;
             }
         }

+        // A wire skipped by an earlier pass is translated for the first time on whatever delta
+        // makes it relevant again -- a deselection, or the reconsider pass above. That delta
+        // describes what changed since the last frame, not what a brand new adapter needs, so it
+        // can legitimately omit MVS_changedMatrix and MVS_changedEffect. Without them
+        // MayaHydraRenderItemAdapter::_transform is never written (it has no initializer, and
+        // UpdateTransform is its only writer) and no material is bound. Treat every new adapter
+        // as a full initialization instead.
+        if (isNewRenderitem) {
+            flags |= MDataServerOperation::MViewportScene::MVS_changedMatrix
+                  |  MDataServerOperation::MViewportScene::MVS_changedEffect;
+        }
+
         if (isNewRenderitem) {
             ...
```

Ordering matters: the new block must run before the material branch at L658 so
`MVS_changedEffect` is already set. `isNewRenderitem` is computed at L602, before the
`replacedByOutline` early-out, so it is in scope here.

Belt and braces — also give the member a defined value so a future path cannot reintroduce an
indeterminate read:

```cpp
// renderItemAdapter.h:222
-    GfMatrix4d                  _transform[2];
+    GfMatrix4d                  _transform[2] { GfMatrix4d(1.0), GfMatrix4d(1.0) };
```

### Suggested regression test

Extend `test/lib/mayaUsd/render/mayaToHydra/cpp/testLegacyHighlightPassRouting.cpp` (added by
this PR) with a case that enables Outline mode, sets the panel shaded, creates a translated cube
that stays selected, switches the panel to wireframe, then asserts the `.../DormantPolyWire`
prim's `HdXformSchema` matrix equals the shape's world matrix and that its material binding is
non-empty.

lib/mayaHydra/hydraExtensions/sceneIndex/mhDirtySelectionColorsSceneIndex.cpp
laferra-adsk, line 82: From Claude analysis:

## 2. Unguarded `GetPrototypes()` now reachable for arbitrary instancers

**File:** `lib/mayaHydra/hydraExtensions/sceneIndex/mhDirtySelectionColorsSceneIndex.cpp:85`
**Severity:** Medium — null dereference.

```cpp
// mhDirtySelectionColorsSceneIndex.cpp:81-88
HdSceneIndexPrim currPrim = GetInputSceneIndex()->GetPrim(currPathToDirty);
if (currPrim.primType == HdPrimTypeTokens->instancer) {
    HdInstancerTopologySchema instancerTopology
        = HdInstancerTopologySchema::GetFromParent(currPrim.dataSource);
    for (const auto& prototypePath : instancerTopology.GetPrototypes()->GetTypedValue(0)) {
        pathsToDirty.push(prototypePath);
    }
}
```

`GetPrototypes()` returns a null handle when the prim has no `prototypes` data source, and
`HdInstancerTopologySchema::GetFromParent` on a prim without instancer topology yields an empty
schema whose accessors all return null.

This was previously only reached from `dirtyLeadObjectRelatedSelections`, i.e. lead-object
selection paths. This PR adds a whole-scene walk:

```cpp
// renderOverride.cpp:1382-1385
if (_wireframeColorsDirty.exchange(false)) {
    ...
    _dirtySelectionColorsSceneIndex->dirtySelectionRelatedPrims(
        { SdfPath::AbsoluteRootPath() });
}
```

so any `instancer`-typed prim anywhere in the scene — including ones contributed by third-party
scene index plugins — now goes through this line on every wireframe colour preference change.

### Proposed fix

```cpp
     HdSceneIndexPrim currPrim = GetInputSceneIndex()->GetPrim(currPathToDirty);
     if (currPrim.primType == HdPrimTypeTokens->instancer) {
         HdInstancerTopologySchema instancerTopology
             = HdInstancerTopologySchema::GetFromParent(currPrim.dataSource);
-        for (const auto& prototypePath : instancerTopology.GetPrototypes()->GetTypedValue(0)) {
-            pathsToDirty.push(prototypePath);
-        }
+        // A prim typed "instancer" is not guaranteed to carry instancer topology, and topology is
+        // not guaranteed to carry prototypes -- the whole-scene walk started from the root reaches
+        // instancers contributed by any scene index, not just ours.
+        if (auto prototypes = instancerTopology.GetPrototypes()) {
+            for (const auto& prototypePath : prototypes->GetTypedValue(0)) {
+                pathsToDirty.push(prototypePath);
+            }
+        }
     }
```

**Related, worth considering:** the loop already pushes every child of every path
(`GetChildPrimPaths`), and prototypes normally live under a path the child walk also reaches.
Under the new root walk the prototype push is therefore redundant, and it can make the same
subtree be visited — and emit duplicate dirty entries — once per referencing instancer. Nested
instancing multiplies this. A cheap visited set bounds it:

```cpp
     std::stack<SdfPath> pathsToDirty({primPath});
+    std::unordered_set<SdfPath, SdfPath::Hash> visited;
     while (!pathsToDirty.empty()) {
         auto currPathToDirty = pathsToDirty.top();
         pathsToDirty.pop();
+        if (!visited.insert(currPathToDirty).second) {
+            continue;
+        }
```

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1268,6 +1431,36 @@ MStatus MtohRenderOverride::Render(
...
+    // without changing whether wireframes are drawn.
+    const unsigned int treatmentRelevantStyleBits
+        = static_cast<unsigned int>(MHWRender::MFrameContext::kWireFrame)
+        | static_cast<unsigned int>(MHWRender::MFrameContext::kGouraudShaded)
+        | static_cast<unsigned int>(MHWRender::MFrameContext::kTextured)
+        // Flat shading is a shaded mode like Gouraud, and "shade active only" decides whether
+        // unselected objects are drawn as wireframe, so both change the treatment.
+        | static_cast<unsigned int>(MHWRender::MFrameContext::kFlatShaded)
+        | static_cast<unsigned int>(MHWRender::MFrameContext::kShadeActiveOnly);
+    if (_mayaHydraSceneIndex
```
laferra-adsk, line 1453: From Claude analysis:

## 3. `_oldDisplayStyle` is per-override, not per-panel

**File:** `lib/mayaHydra/mayaPlugin/renderOverride.cpp:1446-1462`, `renderOverride.h:532`
**Severity:** Medium.

One `MtohRenderOverride` instance serves every model panel using that renderer, and the render
item adapters it mutates are shared across panels, but the display-style memo is a single member:

```cpp
// renderOverride.h:532
unsigned int _oldDisplayStyle {0};

// renderOverride.cpp:1446-1462
if (_mayaHydraSceneIndex
    && ((currentDisplayStyle ^ _oldDisplayStyle) & treatmentRelevantStyleBits) != 0) {
    _mayaHydraSceneIndex->RefreshRenderItemLegacyHighlightTreatment(renderItemOptions);
    renderItemOptions.reconsiderSkippedHighlightWires
        = !viewportDrawsWireframes(_oldDisplayStyle)
        && renderItemOptions.viewportDrawsWireframes;
}

// renderOverride.cpp:1673
_oldDisplayStyle = currentDisplayStyle;
```

With panel1 in wireframe and panel2 shaded, `currentDisplayStyle ^ _oldDisplayStyle` is non-zero
on *every* frame. Consequences:

* `RefreshRenderItemLegacyHighlightTreatment` walks all adapters twice per refresh cycle and
  flips `SetWireframeSelectionHighlightEnabled` back and forth, each flip emitting a visibility
  dirty — steady-state churn proportional to the number of render items.
* `reconsiderSkippedHighlightWires` becomes true on alternating frames, which repeatedly
  re-triggers issue #1.
* Whichever panel drew last decides whether the shared wire prims are visible, so the wireframe
  panel can lose a selection wire the shaded panel just hid.

### Proposed fix

Key the memo by panel, following the existing `_hoverStates` precedent (`renderOverride.h:418`,
`_currentPanelName` at `renderOverride.h:422`):

```cpp
// renderOverride.h
-    unsigned int _oldDisplayStyle {0};
+    // Per panel: one MtohRenderOverride serves every model panel using this renderer, and the
+    // render item adapters it re-treats are shared, so a single memo would make two panels with
+    // different display styles flip the treatment against each other every frame.
+    std::map<std::string, unsigned int> _oldDisplayStyles;
```

```cpp
// renderOverride.cpp — in Render(), before the first use (~L1409)
    const std::string  panelKey = _currentPanelName.asChar();
    const auto         oldStyleIt = _oldDisplayStyles.find(panelKey);
    const unsigned int oldDisplayStyle
        = (oldStyleIt != _oldDisplayStyles.end()) ? oldStyleIt->second : 0u;
```

Then substitute `oldDisplayStyle` for `_oldDisplayStyle` at L1409, L1447 and L1460, and replace
the commit at L1673:

```cpp
-    _oldDisplayStyle = currentDisplayStyle;
+    _oldDisplayStyles[panelKey] = currentDisplayStyle;
```

Clear the map alongside the existing reset at `renderOverride.cpp:2272`, and drop the entry when
a panel is destroyed (same place `_hoverStates` entries are removed).

Note that `RefreshRenderItemLegacyHighlightTreatment` still mutates shared adapter state from a
per-panel decision. Per-panel memoing removes the *thrash*, but the underlying model — one
global "is this wire a highlight" bit for N panels — remains ambiguous when panels genuinely
differ. If that configuration must be correct rather than merely stable, the treatment belongs
in a per-view filtering scene index rather than on the adapter.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -853,11 +952,15 @@ MStatus MtohRenderOverride::Render(
     TF_DEBUG(MAYAHYDRALIB_RENDEROVERRIDE_RENDER).Msg("MtohRenderOverride::Render()\n");
     // We can use the mayaHydraSetVisibleFramePasses command to set the visible passes
 
+    // Filled in below, once the display style for this frame is known, and read by the renderFrame
+    // lambda (which only runs after that point).
+    MayaHydraSceneIndex::RenderItemUpdateOptions renderItemOptions;
+
     auto renderFrame = [&](bool markTime = false) {
         MH_PROFILE_SCOPE("MtohRenderOverride::Render renderFrame lambda");
         if (scene.changed()) {
```
laferra-adsk, line 968: From Claude analysis:

## 4. `reconsiderSkippedHighlightWires` can be consumed without running (low)

**File:** `lib/mayaHydra/mayaPlugin/renderOverride.cpp:961, 1459, 1673`

The flag is only honoured inside `renderFrame`, which gates on `scene.changed()`:

```cpp
// renderOverride.cpp:961-965
if (scene.changed()) {
    if (_mayaHydraSceneIndex) {
        _mayaHydraSceneIndex->UpdateRenderItems(scene, renderItemOptions);
    }
}
```

while `_oldDisplayStyle = currentDisplayStyle;` at L1673 runs unconditionally, so the one-shot
transition cannot fire again.

**Downgraded from the initial review.** In the Maya 2026 devkit (`MViewport2Renderer.h:423`):

```cpp
bool changed() const { return mRemovalCount || mCount; }
```

`mCount` is the size of the item snapshot VP2 hands over, not a count of *modified* items —
which is exactly why `UpdateRenderItems` needs its own `unchanged = (0 == flags)` test. So
`changed()` is false only for an empty scene, where there is nothing to recover. The ordering
within `Render()` is also correct: the flag is computed at L1459, `renderFrame(true)` runs at
L1663/L1666, and the memo is committed afterwards at L1673.

Left as a robustness note rather than a required fix. If you want the guarantee anyway, have
`renderFrame` report whether it ran and commit the memo only then:

```cpp
    bool renderItemsUpdated = false;
    auto renderFrame = [&](bool markTime = false) {
        if (scene.changed()) {
            if (_mayaHydraSceneIndex) {
                _mayaHydraSceneIndex->UpdateRenderItems(scene, renderItemOptions);
                renderItemsUpdated = true;
            }
        }
        ...
    };
    ...
    // Commit only once the pass that consumes the transition actually ran, so a frame that
    // carried no scene snapshot does not swallow the one-shot recovery.
    if (renderItemsUpdated || !renderItemOptions.reconsiderSkippedHighlightWires) {
        _oldDisplayStyle = currentDisplayStyle;
    }
```

(Combine with the per-panel map from #3 if that fix is taken.)

lib/mayaHydra/mayaPlugin/renderGlobals.cpp
```diff
@@ -955,6 +973,46 @@ MObject MtohRenderGlobals::CreateAttributes(const GlobalParams& params)
             return mayaObject;
         }
     }
+    if (filter(MtohTokens->mayaHydraSelectionHighlightMode)) {
+#if PXR_VERSION <= 2411
```
laferra-adsk: From Claude analysis:

## 5. `outlineSelectionHighlight` default contradicts the USD ≤ 24.11 gating (low)

**File:** `lib/mayaHydra/mayaPlugin/renderGlobals.h:104`

`renderGlobals.cpp` deliberately makes Outline mode unavailable on old USD:

```cpp
// renderGlobals.cpp:976-989
#if PXR_VERSION <= 2411
    static const TfTokenVector kSelectionHighlightModes = { TfToken("Legacy Selection") };
#else
    static const TfTokenVector kSelectionHighlightModes
        = { TfToken("Outline Selection"), TfToken("Legacy Selection") };
#endif
    ... kSelectionHighlightModes[0], // default = first entry (Legacy on 24.11, Outline otherwise)

// renderGlobals.cpp:1219-1228
#if PXR_VERSION <= 2411
    TfToken mode("Legacy Selection");
#endif
    globals.outlineSelectionHighlight = (mode == TfToken("Outline Selection"));
```

but the struct member default is unconditional:

```cpp
// renderGlobals.h:102-105
    // The pixel outline replaces the legacy wireframe highlighting rather than adding to it.
    bool            outlineSelectionHighlight = true;
```

Any read of `_globals.outlineSelectionHighlight` that happens before or without the enum being
parsed — the attribute filtered out by `MtohSettingFilter`, or not yet created — enables on
24.11 the mode that is documented there as producing corrupt prim ids.

### Proposed fix

```cpp
-    // The pixel outline replaces the legacy wireframe highlighting rather than adding to it.
-    bool            outlineSelectionHighlight = true;
+    // The pixel outline replaces the legacy wireframe highlighting rather than adding to it.
+    // Mirrors the gating in renderGlobals.cpp: Outline mode is not offered on USD <= 24.11
+    // (corrupt prim ids), so the default must not enable it there either.
+#if PXR_VERSION <= 2411
+    bool            outlineSelectionHighlight = false;
+#else
+    bool            outlineSelectionHighlight = true;
+#endif
```

Confirm `renderGlobals.h` already pulls in something providing `PXR_VERSION` (`pxr/pxr.h`); add
the include if not.

laferra-adsk (changes requested):
Looks good! Added code change suggestions from Claude analysis

debloip-adsk (commented):
Great work! In the interest of time I didn't do a deep dive into the implementation of every method, but the structural and architectural changes look good. Essentially every question or concern that popped up as I was reading ended up being addressed or documented, thank you :)

I did test out the changes locally and hit some issues, but the unit tests passed so it's probably due to my MayaUsd/Maya builds being too old. I think given the significance of the changes we will run it through QA before merging anyways, so I'm not worried

julia-b-grenier:
The outline feature was disabled on MacOS just like it is in Maya with USD 24.11. The Outline feature did not behave correctly on MacOS so the only selection highlighting mode available is the Legacy Wireframe Highlighting mode. See the video attached of the current behavior on MacOS:

https://github.com/user-attachments/assets/55707040-afa1-4ec9-946e-89c8d449abc5

A bug was logged here: https://autodesk.atlassian.net/browse/HYDRA-2504

laferra-adsk:
@beacode rereview

laferra-adsk (dismissed):

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -552,10 +664,6 @@ void MayaHydraSceneIndex::UpdateRenderItems(const MDataServerOperation::MViewpor
             // Update the render item adapter if this render item is an aiSkydomeLight shape
             ria->SetIsRenderITemAnaiSkydomeLightTriangleShape(
                 isRenderItem_aiSkyDomeLightTriangleShape(ri));
-        } else if (ria == nullptr) {
-            // This is a case of null render item adapter.
-            // Used for render items that don't need to be transfered to Hydra.
-            continue;
```
debloip-adsk: Just to be sure, we're never hitting this case?

test/lib/mayaUsd/render/mayaToHydra/cpp/testDeferredHighlightWireInit.cpp
```diff
@@ -0,0 +1,85 @@
+// Copyright 2025 Autodesk
```
debloip-adsk: Copyright should be 2026, same goes for testDeferredHighlightWireInit, testLegacyHighlightPassRouting, testOutlineHover, testSelectionColorPreferences (both cpp and py files)

doc/outlineSelectionHighlightingArchitecture.md
```diff
@@ -0,0 +1,424 @@
...
+
+`UpdateRenderItems()` can only act on what Maya sends, and Maya does not re-send unchanged render
+items when the display style changes. So a switch from shaded into wireframe-on-shaded — where a
+suppressed wire becomes needed again — needs two mechanisms, both driven from `Render()` and both
+gated on the display-style bits that actually affect the treatment (`kWireFrame`, `kGouraudShaded`,
+`kTextured`, `kFlatShaded`, `kShadeActiveOnly`; toggling X-ray or backface culling changes the
+display style without changing whether wireframes are drawn):
+
+- **Already-translated items**: `RefreshRenderItemLegacyHighlightTreatment()` walks the render item
+  adapters and re-evaluates the predicate for each, using the name and DAG path the adapter retains.
```
debloip-adsk: Found by Claude : RefreshRenderItemLegacyHighlightTreatment no longer re-evaluates each time and just uses the cached GetIsReplaceableHighlightWire

lib/mayaHydra/mayaPlugin/renderOverride.h
```diff
@@ -224,6 +238,58 @@ class MtohRenderOverride : public MHWRender::MRenderOverride,
...
+
+    /// Whether any panel this override drives is currently drawing wireframes.
+    bool _AnyPanelDrawsWireframes(const std::string& currentPanel, unsigned int currentStyle) const;
+
+    /// Whether the per-mouse-move pick runs: when hover highlighting is on, or in outline mode when
+    /// mayaHydraForceEnableInteractiveHitTest forces it for profiling.
+    bool _HitTestEnabled() const;
+
+    /// Whether the resolved hover path is drawn using outlines.
+    bool _OutlineHoverHightlightingEnabled() const;
```
debloip-adsk: Found by Claude : kind of a nit but this has an extra "t" between "high" and "light". Not sure how much typos can impact LLM efficacy, but I figure we could fix it alongside the other small fixes

doc/outlineSelectionHighlightingArchitecture.md
```diff
@@ -0,0 +1,424 @@
...
+The mode is a render global, `mayaHydraSelectionHighlightMode`, an enum whose values are
+*Outline Selection* and *Legacy Selection*, created in
+[`renderGlobals.cpp`](../lib/mayaHydra/mayaPlugin/renderGlobals.cpp) and exposed in the renderer's
+option box.
+
+`MtohRenderOverride::_UseOutlineSelectionHighlighting()` is the single predicate the rest of the
+code asks, and it is **not** the same thing as the render global:
+
+```cpp
+return _isUsingHdSt && _globals.outlineSelectionHighlight;
```
debloip-adsk: Found by Claude : this has changed

```suggestion
return _isUsingHdSt && _globals.outlineSelectionHighlight
        && !_globals.forceDisableSelectionHighlight;
```

debloip-adsk (changes requested):
Thanks for all this hard work! I think we're in a pretty good state, just a few minor touchups for consistency and one question. Had Claude do a pass and it didn't really flag anything major either 👍

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
debloip-adsk: Nevermind I just realized it's at the end of a for loop and not doing anything

PR #495, opened 2026-08-20
hydraRender -cf and -f implemented and tested.

ppt-adsk:
The hydraRender command now supports current frame (-cf) and frame (-f) renders.  This is useful when implementing render current frame and render sequence blocking functionality from within Maya, without the use the of the command line Render executable.


cmake/test.cmake
```diff
@@ -1186,3 +1186,183 @@ function(mayaHydra_add_cmd_line_render_test SCENE_FILE_LABELED)
     apply_labels_to_test("${ALL_LABELS}" ${test_name})
 
 endfunction()
+
+# Discover the mayabatch executable (Windows only; macOS/Linux use maya -batch).
+if(IS_WINDOWS)
```
ppt-adsk, line 1191: On Windows the proper way to invoke Maya batch is through the mayabatch executable.  On macOS and Linux it is `maya -batch`.

cmake/test.cmake
```diff
@@ -1186,3 +1186,183 @@ function(mayaHydra_add_cmd_line_render_test SCENE_FILE_LABELED)
...
+        HINTS "${MAYA_LOCATION}" "$ENV{MAYA_LOCATION}" "${MAYA_BASE_DIR}"
+        PATH_SUFFIXES bin/
+        DOC "Maya batch executable")
+    if(NOT MAYA_BATCH_EXECUTABLE)
+        message(FATAL_ERROR "mayabatch not found. Set MAYA_LOCATION.")
+    endif()
+endif()
+
+#
+# mayaHydra_add_mayabatch_render_test( <scene_file_labeled>
```
ppt-adsk, line 1202: Identical to `mayaHydra_add_cmd_line_render_test()` but invokes Maya batch directly to generate an image (through a supplied MEL script) rather than invoking the Render executable.  Still uses idiff to perform the image comparison.

lib/mayaHydra/mayaPlugin/batchRendering/batchRenderer.h
ppt-adsk, line 1: The batch renderer can now optionally carry the frames to be rendered.  Used to implement render current frame and render frame.

lib/mayaHydra/mayaPlugin/batchRendering/hydraRenderCmd.cpp
ppt-adsk, line 1: Implement the -cf and -f flags, and set the render times onto the batch renderer.

lib/mayaHydra/mayaPlugin/batchRendering/hydraRenderCmdHydraV1RenderSettings.cpp
ppt-adsk, line 1: Batch renderer now stores the render times.

test/lib/cmdLineRender/scenes/basic/renderCurrentFrame.mel
ppt-adsk, line 1: Simple Maya batch MEL script to test render current frame (hydraRender -cf).

test/lib/cmdLineRender/scenes/basic/renderFrame.mel
ppt-adsk, line 1: Simple Maya batch MEL script to render an arbitrary frame (hydraRender -f).

test/lib/cmdLineRender/CMakeLists.txt
```diff
@@ -91,6 +91,17 @@ foreach(scene ${ARNOLD_RENDER_SCENES})
     )
 endforeach()
 
+set(ARNOLD_MAYABATCH_RENDER_SCENES
```
ppt-adsk, line 94: New list of tests to run with Maya batch as the source of rendered images, rather than the Render executable.

benyoon-adsk: My understanding is that we don't use command line rendering for the "render current frame" feature in interactive Maya, but rather mayaBatch -> hydraRender, correct?
And this is why you created mayaHydra_add_mayabatch_render_test() ?

test/lib/cmdLineRender/CMakeLists.txt
```diff
@@ -91,6 +91,17 @@ foreach(scene ${ARNOLD_RENDER_SCENES})
...
+    basic/renderCurrentFrame.ma
+    basic/renderFrame.ma
+)
+
+foreach(scene ${ARNOLD_MAYABATCH_RENDER_SCENES})
+    mayaHydra_add_mayabatch_render_test(${scene}
+        COPY_SCENE
```
benyoon-adsk, line 101: Should we only test with V2 render settings from now on?
```
        ENV
            "MAYA_HYDRA_HD_ARNOLD_HYDRA_V2_RENDER_SETTINGS=1"
            "USDIMAGINGGL_ENGINE_ENABLE_SCENE_INDEX=1"
```

benyoon-adsk (approved):
LGTM! Just few questions.

test/lib/cmdLineRender/CMakeLists.txt
ppt-adsk: That's out of scope for this pull request, but definitely something to think about for HdArnold.  HdStorm is not Hydra v2 render setting capable.

test/lib/cmdLineRender/CMakeLists.txt
ppt-adsk: It's rather that when you're in Maya, either interactive or batch, render current frame does not start up the Render executable.  Maya queries the `-renderProcedure` from the `renderer` command (which as you know is the Maya offline renderer registry), and directly calls that.  Since we know that for Hydra-based renderers the render procedure is hydraRender, we just call that directly.

Since the existing test harness only had support for calling the Render executable (except for multi-frame rendering), I had to add CMake test support for calling MEL code directly from mayabatch.

PR #496, opened 2026-08-20
HYDRA-2442 : Add rendering color space resolving scene index

benyoon-adsk:
Given the chosen RenderSettings prim for offline rendering, if the renderingColorSpace property is missing, MayaHydra will fill in the missing property in the Hydra scene with the corresponding value from the Maya color management preferences.

The rendering color space is resolved as follows:
1. If the rendering color space is authored on the active render settings prim, use that value.
    a. If authored but different from the value set in Maya prefs, print a warning message.
    Use the authored value.
    b. If authored but unrecognized (no corresponding value in config.ocio), print a warning
    message. Use the Maya prefs value.
2. If the field is empty or all other case, fallback to Maya prefs.


lib/flowViewport/sceneIndex/fvpRenderingColorSpaceResolvingSceneIndex.h
```diff
@@ -0,0 +1,74 @@
...
+#include "flowViewport/sceneIndex/fvpSceneIndexUtils.h"
+
+#include <pxr/imaging/hd/filteringSceneIndex.h>
+
+#include <string>
+
+namespace FVP_NS_DEF {
+class RenderingColorSpaceResolvingSceneIndex;
+typedef PXR_NS::TfRefPtr<RenderingColorSpaceResolvingSceneIndex>
+    RenderingColorSpaceResolvingSceneIndexRefPtr;
```
benyoon-adsk, line 29: I noticed that in other files, ConstRefPtr such as: `typedef PXR_NS::TfRefPtr<const FrameNbResolvingSceneIndex> FrameNbResolvingSceneIndexConstRefPtr;` are present but not used anywhere. Is there a reason why it exists?

lib/mayaHydra/mayaPlugin/batchRendering/batchRenderer.cpp
```diff
@@ -606,7 +606,12 @@ void BatchRenderer::_CreateSceneIndicesChainAfterMergingSceneIndex()
     //We use as its input scene index : _inputSceneIndexOfFilteringSceneIndicesChain
     _lastFilteringSceneIndexBeforeCustomFiltering = _inputSceneIndexOfFilteringSceneIndicesChain;
 
-    _lastFilteringSceneIndexBeforeCustomFiltering = _sceneGlobalsSceneIndex = HdsiSceneGlobalsSceneIndex::New(_lastFilteringSceneIndexBeforeCustomFiltering);
+    _sceneGlobalsSceneIndex
+        = HdsiSceneGlobalsSceneIndex::New(_lastFilteringSceneIndexBeforeCustomFiltering);
+
+    _lastFilteringSceneIndexBeforeCustomFiltering = _renderingColorSpaceSceneIndex
+        = MhRenderingColorSpaceResolvingSceneIndex::New(_sceneGlobalsSceneIndex);
```
benyoon-adsk, line 616: Inserted right after _sceneGlobalsSceneIndex.

lib/mayaHydra/mayaPlugin/batchRendering/hydraRenderCmd.cpp
```diff
@@ -94,6 +94,9 @@ constexpr auto _layerLong = "-layer";
 constexpr auto _gpuEnabledFlag = "-gpu";
 constexpr auto _gpuEnabledFlagLong = "-gpuEnabled";
 
+constexpr auto _testKeepAliveFlag = "-tka";
+constexpr auto _testKeepAliveFlagLong = "-testKeepAlive";
```
benyoon-adsk: See the comment in hydraRenderCmd.h‎.

lib/mayaHydra/mayaPlugin/batchRendering/hydraRenderCmd.h
```diff
@@ -39,6 +39,13 @@ class HydraRenderCmd : public MPxCommand
 
     MStatus doIt(const MArgList& args) override;
 
+    // Only returns true when -testKeepAlive was passed, so Maya retains this
+    // command (and its BatchRenderer) instead of deleting it right after
+    // doIt() returns. Lets tests inspect the terminal scene index while the
+    // batch renderer is still alive. Not used by real batch renders.
+    bool    isUndoable() const override { return _testKeepAlive; }
+    MStatus undoIt() override { return MS::kSuccess; }
```
benyoon-adsk: From my understanding, doIt() in hydraRenderer terminates the process as soon as it completes, so we can't inspect the scene indices in our unit test. By declaring it as Undoable, it's somehow kept alive ("undo" actions needs to live long enough?) for us to inspect afterwards.

test/lib/mayaUsd/render/mayaToHydra/cpp/testRenderingColorSpaceResolving.py
```diff
@@ -0,0 +1,89 @@
...
+import mtohUtils
+
+from testUtils import PluginLoaded
+
+# The four scenes live alongside the other cmdLineRender renderSettings scenes.
+_SCENES_DIR = os.path.normpath(
+    os.path.join(
+        os.path.dirname(os.path.abspath(__file__)),
+        "..", "..", "..", "..",
+        "cmdLineRender", "scenes", "renderSettings"))
```
benyoon-adsk: Point to the scenes in the repo.
```
test/lib/
├── cmdLineRender/
│   └── scenes/
│       └── renderSettings/
│           ├── renderingColorSpaceDiffering.ma
│           └── renderingColorSpaceDiffering.usda
│
└── mayaUsd/
    └── render/
        └── mayaToHydra/
            └── cpp/
                └── testRenderingColorSpaceResolving.py
```

test/lib/mayaUsd/render/mayaToHydra/cpp/testRenderingColorSpaceResolving.py
```diff
@@ -0,0 +1,89 @@
...
+        # Keep the BatchRenderer alive after doIt() returns.
+        cmds.hydraRender(renderer="HdArnoldRendererPlugin", testKeepAlive=True)
+
+        try:
+            with PluginLoaded('mayaHydraCppTests'):
+                cmds.mayaHydraCppTest(
+                    f="TestRenderingColorSpaceResolving.%s" % testCase)
+        finally:
+            # Release the retained BatchRenderer / Hydra resources.
+            cmds.flushUndo()
```
benyoon-adsk: This is where we destroy hydraRender kept alive from _testKeepAlive in hydraRenderCmd.h.

test/lib/mayaUsd/render/mayaToHydra/CMakeLists.txt
```diff
@@ -278,5 +278,22 @@ foreach(script ${PYTHON_UNIT_TEST_MODULES})
...
+        mayaUsd_add_test(${target}
+            PYTHON_SCRIPT ${test_filename}
+            ENV
+                "MAYA_DEFAULT_SURFACE_SHADER=standardSurface"
+                "MAYAUSD_DISABLE_VP2_RENDER_DELEGATE=1"
+        )
+        apply_labels_to_test("${all_labels}" ${target})
+    endforeach()
+endif()
+
```
benyoon-adsk: Adapted from the script for INTERACTIVE_TEST_SCRIPT_FILES above.

lib/mayaHydra/mayaPlugin/batchRendering/batchRenderer.h
```diff
@@ -159,6 +160,7 @@ class BatchRenderer
     Fvp::BlockPrimRemovalPropagationSceneIndexRefPtr  _blockPrimRemovalPropagationSceneIndex;
     Fvp::PruningSceneIndexRefPtr                      _pruningSceneIndex;
     PXR_NS::HdsiSceneGlobalsSceneIndexRefPtr  _sceneGlobalsSceneIndex;
+    MAYAHYDRA_NS_DEF::MhRenderingColorSpaceResolvingSceneIndexRefPtr _renderingColorSpaceSceneIndex;
```
ppt-adsk:
```suggestion
    MayaHydra::MhRenderingColorSpaceResolvingSceneIndexRefPtr _renderingColorSpaceSceneIndex;
```
MAYAHYDRA_NS_DEF is used when declaring or defining symbols such that this will be done with the proper version suffix in the namespace name.  MayaHydra is a namespace alias so that clients of this code don't have to be aware of the precise version, if they only want to use the latest.

test/lib/mayaUsd/render/mayaToHydra/cpp/CMakeLists.txt
ppt-adsk, line 1: Feels very awkward that this test code would not be in the cmdLineRender hierarchy.  Perhaps under cmdLineRender/cpp?

test/lib/mayaUsd/render/mayaToHydra/CMakeLists.txt
ppt-adsk, line 1: Should be in the cmdLineRender hierarchy, most likely in cmdLineRender/CMakeLists.txt.

test/lib/mayaUsd/render/mayaToHydra/cpp/testRenderingColorSpaceResolving.py
```diff
@@ -0,0 +1,89 @@
...
+import fixturesUtils
+import mtohUtils
+
+from testUtils import PluginLoaded
+
+# The four scenes live alongside the other cmdLineRender renderSettings scenes.
+_SCENES_DIR = os.path.normpath(
+    os.path.join(
+        os.path.dirname(os.path.abspath(__file__)),
+        "..", "..", "..", "..",
```
ppt-adsk: This will go away when you move the tests to the cmdLineRender hierarchy.

test/lib/cmdLineRender/cpp/testRenderingColorSpaceResolving.cpp
ppt-adsk, line 1: Nice test.

lib/mayaHydra/mayaPlugin/batchRendering/hydraRenderCmd.h
ppt-adsk, line 1: I feel this is awkward.  This flag will be public, inviting mis-use.  What you want to keep alive is the scene index tree, starting at the terminal scene index, not necessarily the command that invokes the batch renderer.  What I would do is have some kind of mode, probably triggered by some kind of environment variable, something like MAYA_HYDRA_BATCH_RENDER_TEST_SCENE_INDICES (you and AI can figure out the name).  Under this mode, just before `BatchRenderer::Initialize()` returns, you can save the pointer to the terminal scene index in a static variable in BatchRenderer.  Because scene indices are reference counted, this will keep everything alive.  You can then run your test.

Finally, a static method in BatchRenderer, say `BatchRenderer::ResetTestSceneIndices()`, will set the static pointer variable to null, and the whole scene index tree will disappear.  You can call this at the end of your C++ test, or even add a way to call it from Python.

lib/mayaHydra/mayaPlugin/batchRendering/hydraRenderCmd.cpp
ppt-adsk, line 1: I suggest reverting the change to this file.

lib/mayaHydra/hydraExtensions/sceneIndex/mhRenderingColorSpaceResolvingSceneIndex.h
```diff
@@ -0,0 +1,66 @@
...
+    MAYAHYDRALIB_API
+    PXR_NS::HdSceneIndexPrim GetPrim(const PXR_NS::SdfPath& primPath) const override;
+
+    PXR_NS::SdfPathVector GetChildPrimPaths(const PXR_NS::SdfPath& primPath) const override
+    {
+        return GetInputSceneIndex()->GetChildPrimPaths(primPath);
+    }
+
+    MAYAHYDRALIB_API
+    std::string GetRenderingColorSpaceFromDCC() const override;
```
ppt-adsk: Minor:
```suggestion
    std::string GetApplicationRenderingColorSpace() const override;
```

lib/mayaHydra/hydraExtensions/sceneIndex/mhRenderingColorSpaceResolvingSceneIndex.h
```diff
@@ -0,0 +1,66 @@
...
+    MAYAHYDRALIB_API
+    bool IsKnownColorSpace(const std::string& colorSpace) const override;
+
+    ~MhRenderingColorSpaceResolvingSceneIndex() override = default;
+
+protected:
```
ppt-adsk: Very minor: if we don't expect any further derivation, may as well be private.

lib/mayaHydra/hydraExtensions/sceneIndex/mhRenderingColorSpaceResolvingSceneIndex.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+    const TfToken rcs = renderingColorSpaceDs->GetTypedValue(0.0f);
+    if (!rcs.IsEmpty()) {
+        // If rendering color space is authored and valid with the OCIO config file, use it.
+        if (IsKnownColorSpace(rcs.GetString())) {
+            // Authored rendering color space doesn't match the one from Maya prefs.
+            // Let the user know that we're using the one from the active USD render settings node.
+            if (rcs.GetString() != renderingColorSpaceFromMayaPrefs) {
+                TF_WARN(
+                    "Authored rendering color space '%s' in the active USD render settings prim "
```
ppt-adsk: Technically you can obtain the active render settings prim from any data provider, not just USD, because it's a **Hydra** render settings prim --- though I have never seen this, nor do I expect it would ever happen (but you never know).  I would drop the USD here:
```suggestion
                    "Authored rendering color space '%s' in the active render settings prim "
```

lib/mayaHydra/hydraExtensions/sceneIndex/mhRenderingColorSpaceResolvingSceneIndex.cpp
```diff
@@ -0,0 +1,155 @@
...
+                    "doesn't match the value from Maya preferences '%s'. "
+                    "Using the value from the active USD render settings prim.",
```
ppt-adsk:
```suggestion
                    "Using the value from the active render settings prim.",
```

lib/mayaHydra/hydraExtensions/sceneIndex/mhRenderingColorSpaceResolvingSceneIndex.cpp
```diff
@@ -0,0 +1,155 @@
...
+                    rcs.GetText(),
+                    renderingColorSpaceFromMayaPrefs.c_str());
+                return prim;
+            }
+            return prim;
+        }
+        // If rendering color space is authored but not valid with the OCIO config file, let the
+        // user know that we're using the one from Maya prefs.
+        TF_WARN(
+            "Authored rendering color space '%s' in the active USD render settings prim is not "
```
ppt-adsk:
```suggestion
            "Authored rendering color space '%s' in the active render settings prim is not "
```

lib/flowViewport/sceneIndex/fvpRenderingColorSpaceResolvingSceneIndex.h
ppt-adsk: For clients of this code.  They don't have to define it themselves.  It's an informal part of the API.

lib/flowViewport/sceneIndex/fvpRenderingColorSpaceResolvingSceneIndex.h
```diff
@@ -0,0 +1,74 @@
...
+///
+class RenderingColorSpaceResolvingSceneIndex
+    : public PXR_NS::HdSingleInputFilteringSceneIndexBase
+    , public Fvp::InputSceneIndexUtils<RenderingColorSpaceResolvingSceneIndex>
+{
+public:
+    using PXR_NS::HdSingleInputFilteringSceneIndexBase::_GetInputSceneIndex;
+
+    FVP_API
+    virtual std::string GetRenderingColorSpaceFromDCC() const = 0;
```
ppt-adsk: See naming suggestion in mhRenderingColorSpaceResolvingSceneIndex.h

lib/mayaHydra/hydraExtensions/sceneIndex/mhRenderingColorSpaceResolvingSceneIndex.cpp
```diff
@@ -0,0 +1,155 @@
...
+        return prim;
+    }
+
+    const std::string renderingColorSpaceFromMayaPrefs = GetRenderingColorSpaceFromDCC();
+    if (renderingColorSpaceFromMayaPrefs.empty()) {
+        return prim;
+    }
+
+    const TfToken rcs = renderingColorSpaceDs->GetTypedValue(0.0f);
+    if (!rcs.IsEmpty()) {
```
ppt-adsk: Everything above this line (line 90) is not Maya-specific, and can be moved to the base class.  Lines 91-115 (inclusive) can be a virtual called by the base class, and implemented in this class.

Lines 116-129 (inclusive) are not Maya-specific, and can be moved to the base class.

ppt-adsk (changes requested):
Looking good, a couple of changes to make, but on the right track.

lib/mayaHydra/mayaPlugin/batchRendering/hydraRenderCmd.h
benyoon-adsk: I tried setting a static variable pointing to HdRenderIndex's terminal scene index at initialization and reset it at the end of the test, but it didn't work because it seems like some objects needed to keep the whole scene index tree + prims + data sources are cleared in _ClearHydraResources() (i.e. _mayaHydraSceneIndex, _sceneIndexRegistry, _renderIndex).

From my understanding, the terminal scene index was still alive, but it no longer had queryable scene globals / render-settings data:

```
D:\dev\maya-hydra\ecg-maya-hydra\maya-hydra\test\lib\cmdLineRender\cpp\testRenderingColorSpaceResolving.cpp(76): error: Value of: globalsSchema.IsDefined()
  Actual: false
Expected: true
D:\dev\maya-hydra\ecg-maya-hydra\maya-hydra\test\lib\cmdLineRender\cpp\testRenderingColorSpaceResolving.cpp(146): error: Value of: rcs
  Actual: false
Expected: true
```

What I did is to simply extend the lifetime of the BatchRenderer object when HydraRenderCmd destructs so that everything stays intact and is queryable on render completion. BatchRenderer object is released when python test completes. This is gated by an env var. This mirrors object lifetime in viewport testing (object init -> inspect prims -> destroy viewport object upon test completion).

ppt-adsk, line 1: Nit-pick: at this point the only change to the file is a white space change, usually we don't create a version just for that.

ppt-adsk (approved):
Great work!

PR #499, opened 2026-08-25
HYDRA-2482 : Playback performance regression

lanierd-adsk:
Simplify RenderItemShouldEmitTopologyLocators by removing the redundant !topoChanged && !geomChanged guard and collapsing the topo-only branch into a clearer !topoChanged / !geomChanged early-return sequence. Replace the unreachable return topoChanged with an explicit return true once topoChanged is known, and clarify comments for the geometry-only (HYDRA-2417) and MAYA-134200 disambiguation paths.


lanierd-adsk:
In my build it gave :
With Squid scene
latest commit ~14 FPS
Commit 2548f00 before hydra V2 locators change : ~20FPS 
With fix :  ~ 24 FPS

vlasovi (approved):
Overall looks good. This should eliminate the major slow downs. The index buffer comparison may still run under some narrow conditions and I'm not sure if this is really an optimization. But this shouldn't be a big problem and also after the introduction of GPU buffer sharing this code path shouldn't be touched at all.

lilike-adsk (approved):
Looks good, back to the old code which leveraged `topoChanged ` flag to cover the index change case, this should be fine for deforming workflow, if that flag doesn't cover all case, we can handle case by case later.

PR #504, opened 2026-08-27
HYDRA-2422 - UpdateRenderItems: replace multiple heavy static checks with a single boolean test

vlasovi:
There are multiple cases where a render item adapter doesn't have to be created for a render item. Instead of performing those checks every frame, perform them only once, when decide to create a new adapter and create a null adapter if the render item doesn't need one.


lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
```diff
@@ -500,39 +492,49 @@ void MayaHydraSceneIndex::UpdateRenderItems(const MDataServerOperation::MViewpor
...
             const SdfPath slowId = _GetRenderItemPrimPath(ri);
 
             // Maya/MtoA adds texturedSkyDome mesh object for VP2.
             // We do not want that to be translated to Hydra
             if (slowId.IsEmpty() || IsTexturedSkyDomeRenderItem(slowId)) {
+                createNullRenderItemAdapter = true;
+            }
+            
+            if (createNullRenderItemAdapter) {
+                _renderItemsAdaptersFast.insert({ fastId, nullptr });
```
lilike-adsk, line 533: do we still need to add them to the list if there's no adapter at all?

vlasovi: Yes because this is how we mark that we checked this render item and we don't need it. This is a quick verification that is done on update instead of those heavy name checks every frame.

vlasovi: You can look at line 555. That's where we skip those render items.

lilike-adsk: ok, how much benefit we can get from this PR?

lilike-adsk (approved):

lib/mayaHydra/hydraExtensions/sceneIndex/mayaHydraSceneIndex.cpp
vlasovi: About 2% of the total frame time on the Hulk scene (big number of animated prims). At this point we optimized all the bigger pieces. All that is left is smaller pieces here and there.

PR #505, opened 2026-08-28
Replaced Maya defaultRenderGlobals start and end with UsdRenderSetup equivalents.

ppt-adsk:
Render frame range must now come from a custom attribute on the render settings prim.


lib/mayaHydra/mayaPlugin/renderSettingsUtils.cpp
ppt-adsk, line 1: Core file in the review: replace Maya defaultRenderGlobals frame range with equivalent from USD render settings prim, which is defined using a custom attribute.

The custom attribute supports a vector of frame ranges, but we are only considering the first element in the vector for this release.

scripts/mayaHydra/renderSettings/frames.py
ppt-adsk, line 1: Python script that implements the frame range override for the Render executable must now set the USD property on the render settings prim.

test/lib/cmdLineRender/scenes/renderSettings/images/arnold_frames5to9/secondCamFrame5.png
ppt-adsk, line 1: Added expected output frames for second camera in animated cube test (see renderSettingsMultiImageTest.py comment).

test/lib/cmdLineRender/scenes/renderSettings/arnold/images/singleCam_frames5to9/mainCamFrame5.png
ppt-adsk, line 1: Separated out single camera animated cube test expected images (see renderSettingsMultiImageTest.py comments).

test/lib/cmdLineRender/renderSettingsMultiImageTest.py
ppt-adsk, line 1: When I made the change to the frame range source (Maya to USD), to my surprise the 5-9 animated cube tests still passed.  This was because they were now rendering all frames (1 to 10), and comparing the expected 5-9 files with the output.  Since the output was a superset of the expected 5-9 files, the tests passed, which was incorrect.

Added this verification that the output files must exactly match the expected files, with no extra files.  When I added this code, the test now failed, which is the correct behavior.  I then fixed the frame range override Python code, and the test now passed again.  This meant I had to add expected files for the secondary camera for one test, and remove expected files for the secondary camera in the single camera case.

lanierd-adsk (dismissed):

lib/mayaHydra/mayaPlugin/renderSettingsUtils.cpp
```diff
@@ -284,30 +284,34 @@ RenderTimes::RenderTimes(
...
-    MRenderUtil::getCommonRenderSettings(mayaRenderSettings);
-
-    if (!mayaRenderSettings.isAnimated()) {
-        const auto currentTime = MAnimControl::currentTime();
-        return RenderTimes(false, currentTime, currentTime, 1.0f);
+    UsdRenderSettings usdRenderSettings;
+    const auto rsPath = ExtractUsdRenderSettingsFromScene(usdRenderSettings);
+
+    if (!rsPath.empty()) {
+        const UsdPrim rsPrim = usdRenderSettings.GetPrim();
```
lanierd-adsk, line 295: Is there a possibility this prim is not valid ?

lanierd-adsk (commented):
Thanks for this work. 
Only a simple question.

lib/mayaHydra/mayaPlugin/renderSettingsUtils.cpp
ppt-adsk: Excellent point, will investigate.

ppt-adsk: After investigation, ExtractUsdRenderSettingsFromScene() calls FindUsdRenderSettingsOnStage(), which returns false if the prim is not a valid UsdRenderSettings prim, and thus ExtractUsdRenderSettingsFromScene() will return an empty Ufe::Path, and thus at line 295 rsPrim must be valid.

lanierd-adsk (dismissed):

lanierd-adsk (dismissed):

benyoon-adsk (approved):

PR #506, opened 2026-08-28
HYDRA-2447 : Set the active render description prim on the scene globals scene index

benyoon-adsk:
To get the render pass collections to apply to rendering, the active render pass must be set on the scene globals scene index.

The render pass prim can be found on the UsdDefaultRenderDescription (formerly UsdDefaultRenderSettings) node, in the activeRenderDescriptionPath (formerly activeSettingsPath).  This attribute is a dual-purpose attribute that can store a path to a UsdRenderSettings prim, or a UsdRenderPass prim.

It is resolved as follows:
```
// 1. If the attribute points to a UsdRenderSettings prim, use it directly.
// 2. If the attribute points to a UsdRenderPass prim:
//    2.1 If its renderSource points to a valid UsdRenderSettings prim, use
//         both the pass and its render settings.
//    2.2 If renderSource is missing or invalid, use the default USD render
//         settings and warn.
//    2.3 If the render pass does not exist, use the default USD render
//         settings and warn.
```


lib/mayaHydra/mayaPlugin/batchRendering/batchRenderer.cpp
```diff
@@ -242,6 +242,11 @@ bool BatchRenderer::Initialize()
             _lastFilteringSceneIndexBeforeCustomFiltering);
     }
 
+    // The render view attaches the registered USD-stage data producers to the
+    // merging scene index. Resolve only after that attachment so the active
+    // render settings or render pass can be retrieved from the scene index.
+    _SetActiveRenderDescriptionFromScene();
```
benyoon-adsk: In _SetActiveRenderDescriptionFromScene(), ResolveActiveRenderDescription() calls sceneIndex.GetPrim(activePath) to check primType (is it renderSettings or a renderPass?). That needs the USD prims already merged into _sceneGlobalsSceneIndex's input, which only happens after RenderViewDataManager::AddRenderViewData(). Hence the move to Initialize().

lib/mayaHydra/mayaPlugin/batchRendering/batchRenderer.cpp
```diff
@@ -693,23 +696,36 @@ HdRenderIndex* BatchRenderer::renderIndex() const
     return _renderIndex;
 }
 
-void BatchRenderer::_SetActiveRenderSettingsPrimFromScene()
+// Set ActiveRenderSettingsPrimPath and ActiveRenderPassPrimPath on the scene globals scene index.
+void BatchRenderer::_SetActiveRenderDescriptionFromScene()
```
benyoon-adsk: I know you mentioned you wanted to keep the interface, but I thought it made more sense to rename it as _SetActiveRenderDescriptionFromScene() and add a resolver function ResolveActiveRenderDescription(), which handles what GetActiveRenderSettingsHydraPath() does and what GetActiveRenderPassHydraPath() would have done if it existed. What do you think?

lib/mayaHydra/mayaPlugin/batchRendering/batchRenderer.cpp
```diff
@@ -693,23 +696,36 @@ HdRenderIndex* BatchRenderer::renderIndex() const
...
 {
     if (!TF_VERIFY(_sceneGlobalsSceneIndex, "Scene globals scene index not yet initialized")) {
         return;
     }
 
-    const auto hydraRsPath = GetActiveRenderSettingsHydraPath();
-    if (!TF_VERIFY(!hydraRsPath.IsEmpty(), 
+    const auto activeRenderDescription
+        = ResolveActiveRenderDescription(*_sceneGlobalsSceneIndex);
```
benyoon-adsk: See the code comment in renderSettingsUtils.cpp for resolving logic.

lib/mayaHydra/mayaPlugin/batchRendering/batchRenderer.cpp
```diff
@@ -693,23 +696,36 @@ HdRenderIndex* BatchRenderer::renderIndex() const
...
+
+    TF_DEBUG_MSG(MAYAHYDRAPLUGIN_BATCHRENDER_RENDER_SETTINGS,
+                 "Active render pass set to " +
+                 (activeRenderDescription.renderPassPath.IsEmpty()
+                      ? std::string("none")
+                      : activeRenderDescription.renderPassPath.GetAsString()) + "\n");
 
-    _sceneGlobalsSceneIndex->SetActiveRenderSettingsPrimPath(hydraRsPath);
+    _sceneGlobalsSceneIndex->SetActiveRenderPassPrimPath(
+        activeRenderDescription.renderPassPath);
```
benyoon-adsk: If activeRenderDescription.renderPassPath is empty, it will not be set. Hence why there is no `if (activeRenderDescription.renderPassPath)` condition.

lib/mayaHydra/mayaPlugin/renderSettingsUtils.cpp
```diff
@@ -194,31 +196,102 @@ Ufe::Path GetActiveRenderSettingsAppPath()
...
+//         both the pass and its render settings.
+//    2.2 If renderSource is missing or invalid, use the default USD render
+//         settings and warn.
+//    2.3 If the render pass does not exist, use the default USD render
+//         settings and warn.
+ActiveRenderDescription ResolveActiveRenderDescription(const HdSceneIndexBase& sceneIndex)
+{
+    ActiveRenderDescription result;
+
+    const SdfPath activePath = GetActiveRenderDescriptionHydraPath();
```
benyoon-adsk: Get the ActiveRenderDescription path to check whether it is a renderSettings or RenderPass prim.

test/lib/cmdLineRender/cpp/testActiveRenderDescriptionSceneGlobals.cpp
benyoon-adsk, line 1: Adapted from testRenderingColorSpaceResolving.

test/lib/cmdLineRender/scenes/renderSettings/activeRenderDescriptionInvalidPassSource.usda
```diff
@@ -0,0 +1,105 @@
...
+        rel products = </Render/BeautyProduct>
+        int2 resolution = (1920, 1080)
+    }
+
+    def Scope "Passes"
+    {
+        def RenderPass "foreground"
+        {
+            token passType = "beauty"
+            rel renderSource = </Render/NonExistingSettings>
```
benyoon-adsk, line 22: The invalide renderSource.

test/lib/cmdLineRender/CMakeLists.txt
```diff
@@ -478,14 +479,11 @@ foreach(script ${MAYABATCH_TEST_SCRIPT_FILES})
...
-    if(test_filename STREQUAL "testRenderingColorSpaceResolving.py")
-        list(APPEND test_env "MAYA_HYDRA_BATCH_RENDER_TEST_MODE=1")
-    endif()
-
     mayaUsd_add_test(${target}
         PYTHON_SCRIPT ${test_filename}
-        ENV ${test_env}
+        ENV
+            "DISPLAY=:0"
+            "MAYA_HYDRA_BATCH_RENDER_TEST_MODE=1"
```
benyoon-adsk: registeredMayaRenderers.py doesn't call hydraRender so it's safe to put the env var "MAYA_HYDRA_BATCH_RENDER_TEST_MODE=1" for all tests in MAYABATCH_TEST_SCRIPT_FILES.

test/lib/cmdLineRender/testActiveRenderDescriptionSceneGlobals.py
```diff
@@ -0,0 +1,78 @@
...
+                os.makedirs(destDir)
+            baseName = os.path.splitext(sceneFileName)[0]
+            for ext in (".ma", ".usda"):
+                shutil.copy(_scenePath(baseName + ext), os.path.join(destDir, baseName + ext))
+            scenePath = os.path.join(destDir, sceneFileName)
+        else:
+            scenePath = _scenePath(sceneFileName)
+
+        cmds.file(scenePath, open=True, force=True)
+        cmds.hydraRender(renderer="HdArnoldRendererPlugin")
```
ppt-adsk, line 54: I wonder if we should use the -cf flag, to ensure that only a single frame will be rendered, since that's all that's needed.

test/lib/cmdLineRender/CMakeLists.txt
ppt-adsk: That sounds like a potentially problematic forward looking statement.  What happens when we start adding more tests?

test/lib/cmdLineRender/cpp/testActiveRenderDescriptionSceneGlobals.cpp
```diff
@@ -0,0 +1,117 @@
...
+    EXPECT_EQ(active.renderSettings, _GetSceneIndexPath("/Render/MainRender"));
+    EXPECT_EQ(active.renderPass, _GetSceneIndexPath("/Render/Passes/foreground"));
+}
+
+TEST(TestActiveRenderDescriptionSceneGlobals, InvalidRenderPassSource)
+{
+    const auto active = _GetActiveRenderDescriptionPaths();
+
+    EXPECT_EQ(active.renderSettings, _GetDefaultRenderSettingsPath());
+    EXPECT_EQ(active.renderPass, _GetSceneIndexPath("/Render/Passes/foreground"));
```
ppt-adsk, line 108: I don't understand this test.  If the render pass is invalid, we get some kind of default render settings?  Same with the render pass?

lib/mayaHydra/mayaPlugin/renderSettingsUtils.h
```diff
@@ -83,11 +85,33 @@ bool FindUsdRenderSettingsOnStage(
...
-// Get the Hydra path to the active render settings prim from the Maya scene.
-PXR_NS::SdfPath GetActiveRenderSettingsHydraPath();
+// Get the Hydra path to the active render description prim from the Maya scene.
+PXR_NS::SdfPath GetActiveRenderDescriptionHydraPath();
+
+// Get the UFE application path to the default render settings prim.
+Ufe::Path GetDefaultRenderSettingsAppPath();
+
+// Get the Hydra path to the default render settings prim.
+PXR_NS::SdfPath GetDefaultRenderSettingsHydraPath();
```
ppt-adsk: Why is this different from GetActiveRenderSettingsHydraPath()?

lib/mayaHydra/mayaPlugin/renderSettingsUtils.h
```diff
@@ -83,11 +85,33 @@ bool FindUsdRenderSettingsOnStage(
...
+
+/*! \brief Result of resolving the active render description.
+ *
+ *  activeRenderDescriptionPath can point to either render settings or a
+ *  render pass. A pass uses the render settings from its renderSource.
+ *  If those settings cannot be resolved, rendering uses the defaults.
```
ppt-adsk: What are the defaults?

lib/mayaHydra/mayaPlugin/renderSettingsUtils.cpp
```diff
@@ -84,7 +86,7 @@ namespace MAYAHYDRA_NS_DEF {
 
 Ufe::Path ExtractUsdRenderSettingsFromScene(UsdRenderSettings& usdRenderSettings)
 {
-    const auto rsAppPath = GetActiveRenderSettingsAppPath();
+    const auto rsAppPath = GetActiveRenderDescriptionAppPath();
```
ppt-adsk: What happens if rsAppPath is a path to a render pass?  Can FindUsdRenderSettings deal with this?

lib/mayaHydra/mayaPlugin/renderSettingsUtils.cpp
```diff
@@ -194,31 +196,102 @@ Ufe::Path GetActiveRenderSettingsAppPath()
...
 
+// 1. If the attribute points to a UsdRenderSettings prim, use it directly.
+// 2. If the attribute points to a UsdRenderPass prim:
+//    2.1 If its renderSource points to a valid UsdRenderSettings prim, use
+//         both the pass and its render settings.
+//    2.2 If renderSource is missing or invalid, use the default USD render
+//         settings and warn.
+//    2.3 If the render pass does not exist, use the default USD render
+//         settings and warn.
+ActiveRenderDescription ResolveActiveRenderDescription(const HdSceneIndexBase& sceneIndex)
```
ppt-adsk: I don't understand what's going on here.  Why do we need to interrogate a scene index to figure this out?

ppt-adsk (changes requested):
Discussed the approach with you, still thinking about whether using the Hydra scene is good or bad, certainly works.  Will get back to you on that.

In the meantime, we do need to adjust getRenderSettingsPrim() in scripts/mayaHydra/renderSettings/utils.py to account for whether the Maya node attribute is pointing to a render pass or render settings prim.

test/lib/cmdLineRender/cpp/testActiveRenderDescriptionSceneGlobals.cpp
benyoon-adsk: My understanding was that if the active render description prim points to a render pass, but its render source is invalid, we still use the authored render pass but with the default render settings. Should I check with Product whether we should fall back to a "default render pass"?

ppt-adsk: That makes sense, but let's check just to be sure.

test/lib/cmdLineRender/CMakeLists.txt
benyoon-adsk: Should I add a comment on line 485, something like:
```
# Important: All tests in MAYABATCH_TEST_SCRIPT_FILES MUST release
# the batch renderer by calling cmds.mayaHydraTesting(releaseBatchRenderer=True).
# The batchRenderer object destruction is deferred so tests can inspect the hydra scene index tree.
# See the example in testActiveRenderDescriptionSceneGlobals.py
foreach(script ${MAYABATCH_TEST_SCRIPT_FILES})
...
```
Or should we revert to before?

test/lib/cmdLineRender/testActiveRenderDescriptionSceneGlobals.py
benyoon-adsk: We could, but the test scenes don't have animation so we'll only see a single frame.

lib/mayaHydra/mayaPlugin/renderSettingsUtils.cpp
```diff
@@ -177,32 +178,107 @@ bool FindUsdRenderSettingsOnStage(
...
 
-Ufe::Path GetActiveRenderSettingsAppPath()
+Ufe::Path GetDefaultRenderSettingsAppPath()
+{
+    constexpr const char* rsPrimPath = "/Render/SceneRenderSettings";
+    return Ufe::PathString::path(
+        std::string(kUsdDefaultRenderDescriptionNodeName) + "," + rsPrimPath);
+}
+
+static bool _ReadActiveRenderDescriptionPrim(Ufe::Path& outPath, UsdPrim& outPrim)
```
benyoon-adsk: Helper function to read to tell whether the active render desc prim is a render pass or render settings. Used in GetActiveRenderSettingsAppPath() and GetActiveRenderPassAppPath().

lib/mayaHydra/mayaPlugin/renderSettingsUtils.cpp
```diff
@@ -177,32 +178,107 @@ bool FindUsdRenderSettingsOnStage(
...
+        return false;
+    }
+
+    // Ufe::PathString::path() will throw Ufe::InvalidPath when the string 
+    // does not resolve to an existing UFE item (e.g. a render pass
+    // path that does not exist, such as "/Render/Passes/doesNotExist").
+    // Catch it and return false so that the default USD render settings are used.
+    try {
+        outPath = Ufe::PathString::path(pathStr.asChar());
+    } catch (const std::exception& e) {
```
benyoon-adsk: Ufe will throw an exception if the path is non-existent (|something,/Render/Passes/doesNotExist). We need to handle case 2.3:
```
    // 2. If the attribute points to a UsdRenderPass prim:
    //    2.3 If the render pass does not exist, use the default USD render
    //         settings and warn.
```

lib/mayaHydra/mayaPlugin/renderSettingsUtils.cpp
```diff
@@ -177,32 +178,107 @@ bool FindUsdRenderSettingsOnStage(
...
+            // Case 2.2
+            if (!settingsPrim.IsValid()) {
+                TF_WARN("The activeRenderDescriptionPath points to a UsdRenderPass prim that does not have "
+                        "valid renderSource. The default USD render settings will be used.");
+                return GetDefaultRenderSettingsAppPath();
+            }
+
+            // Case 2.1.
+            // ex: |renderSettings|renderSettingsShape,/Render/Passes/foreground
+            //  -> |renderSettings|renderSettingsShape,/Render/MainRender
```
benyoon-adsk, line 272: For case 2.1, pop out the render pass segment in Ufe and insert the renderSource. It will be used later in     _SetActiveRenderSettingsPrimFromScene().

lib/mayaHydra/mayaPlugin/renderSettingsUtils.cpp
```diff
@@ -219,6 +295,36 @@ SdfPath GetActiveRenderSettingsHydraPath()
...
+{
+    Ufe::Path path;
+    UsdPrim   prim;
+    if (!_ReadActiveRenderDescriptionPrim(path, prim) || !prim.IsValid()) {
+        return {};
+    }
+
+    // If the active render description prim points to a render pass, we simply return its path.
+    // Its renderSource (activeRenderSettingsPrimPath) will be resolved in
+    // GetActiveRenderSettingsAppPath().
```
benyoon-adsk, line 310: See comment.

scripts/mayaHydra/renderSettings/utils.py
```diff
@@ -17,9 +17,35 @@
...
-    # Get the UFE path to the render settings prim to use from the
-    # default USD render settings node.
     rsPath = cmds.getAttr('UsdDefaultRenderDescription.activeRenderDescriptionPath')
+    defaultRenderSettingsPath = "UsdDefaultRenderDescription,/Render/SceneRenderSettings"
+
+    # 2.3 If the active render description prim points to a render pass / settings
+    # that does not exist, use the default USD render settings.
+    try:
+        renderDescriptionPrim = mayaUsdUfe.ufePathToPrim(rsPath)
+    except RuntimeError:
```
benyoon-adsk, line 30: See the comment. Catch the exception like in renderSettingsUtils.cpp.

test/lib/cmdLineRender/scenes/renderSettings/arnold/arnoldSimpleRenderPass.png
benyoon-adsk, line 1: Image diff test to test basic render pass feature in HdArnold

lib/mayaHydra/mayaPlugin/renderSettingsUtils.cpp
ppt-adsk: Nit-pick: in modern C++ you should use the unnamed namespace to avoid exposing a symbol, rather than the static link qualifier.

ppt-adsk: A bit awkward to return a boolean when all call sites check the UsdPrim for validity.  May as well just return the UsdPrim, which can be invalid on failure.

ppt-adsk (commented):
Looks much better, just a couple of nit-picks.  I believe you'll have to merge in the latest dev so that your tests end up in the right place.

test/lib/cmdLineRender/CMakeLists.txt
ppt-adsk: I think it should be the responsibility of tests that need this functionality to clean up, rather than expose other tests to this change.

lib/mayaHydra/mayaPlugin/renderSettingsUtils.cpp
```diff
@@ -48,6 +48,7 @@
 #include <cstdlib>
 #include <string>
 
+using namespace MayaHydra;
```
benyoon-adsk, line 51: kUsdDefaultRenderDescriptionNodeName and GetDependNodeFromNodeName are in MayaHydra namespace. Instead of MayaHydra::kUsdDefaultRenderDescriptionNodeName, I used "using namespace" like what batchRenderer.cpp does.

test/lib/cmdLineRender/scenes/renderSettings/arnold/arnoldSimpleRenderPass.ma
benyoon-adsk, line 1: With the PRMan test coverage PR, we have a new scenes/renderSettings/arnold directory. Only add this test to the arnold directory because it's specific to arnold.

test/lib/cmdLineRender/scenes/renderSettings/activeRenderDescriptionInvalidPassSource.ma
benyoon-adsk, line 1: Did not move activeRenderDescription* scene files to the scenes/renderSettings/arnold/ directory because the unit test uses those to inspect the scene index tree. This is independent of the render delegate, so they can stay in the parent directory.

ppt-adsk (approved):

PR #507, opened 2026-09-02
HYDRA-2518 : Re-enable renderSettings/AnimCubeRenderSettings on OSX

lanierd-adsk:
Re-enable renderSettings/AnimCubeRenderSettings on OSX

### DIFF
--- a/test/lib/cmdLineRender/CMakeLists.txt
+++ b/test/lib/cmdLineRender/CMakeLists.txt
@@ -15,6 +15,27 @@
 
 set(TEST_OUTPUT_ROOT ${CMAKE_BINARY_DIR}/test/Temporary)
 
+# HYDRA-2518 TEMPORARY: isolate the macOS preflight down to a single test
+# (AnimCubeRenderSettings_Arnold) so a blind preflight run returns focused
+# evidence instead of being buried in the rest of the suite's output. All
+# OSX Maya 2028 unit tests are already disabled elsewhere, so this scaffold
+# only needs to fire for Maya 2027 -- there is no point registering it for
+# 2028, since nothing on that combination runs anyway. Remove this whole
+# block (and the return() below) once HYDRA-2518 is diagnosed.
+if(IS_MACOSX AND MAYA_APP_VERSION VERSION_EQUAL 2027)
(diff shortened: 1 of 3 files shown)


test/lib/cmdLineRender/renderSettingsMultiImageTest.py
```diff
@@ -208,6 +208,24 @@ def _find_output_dir(base_dir, rendered_subdir):
     return output_dir
 
 
+def _dump_base_dir_listing(base_dir):
```
lanierd-adsk, line 211: Added this debugging info function when therender fails.

test/lib/cmdLineRender/CMakeLists.txt
```diff
@@ -284,10 +284,8 @@ function(_mayaHydra_add_render_settings_variant folder res_arg new_image_name)
         endif()
     endif()
 
-    # Failure with HdStorm on macOS entered as HYDRA-2139. (prman never runs on
-    # macOS, so this reduces to NOT IS_MACOSX for storm/arnold and is always
-    # true for prman.)
-    if(NOT IS_MACOSX)
+    # Failure with HdStorm on macOS entered as HYDRA-2139.
+    if(NOT IS_MACOSX OR NOT folder STREQUAL "storm")
```
lanierd-adsk, line 288: No real fix as only re-enabling it made it pass on OSX 2027.

test/lib/cmdLineRender/renderSettingsMultiImageTest.py
```diff
@@ -239,7 +257,9 @@ def main(argv):
     scene_copy = _copy_scene_and_usd(scene_path, work_dir)
     _prepare_output_dir(base_dir, rendered_subdir)
 
-    if not _run_render(render_exe, renderer, scene_copy, work_dir, extra_renderer_args):
+    render_succeeded = _run_render(render_exe, renderer, scene_copy, work_dir, extra_renderer_args)
+    _dump_base_dir_listing(base_dir)
```
ppt-adsk: Why are we doing the dump unconditionally?  Shouldn't it be only when render_succeeded is false?

lanierd-adsk: Good catch, that was a mistake.

ppt-adsk (approved):

PR #508, opened 2026-09-02
HYDRA-2516 : Force MayaUSD plugin to be loaded first so registerVersi…

lanierd-adsk:
…onedPlugins() runs before MtohGetRendererDescriptions() freezes HdRendererPluginRegistry via GetPluginDescs()


lanierd-adsk:
@benyoon-adsk said he will check the fix in Maya 2027

ppt-adsk (approved):

benyoon-adsk:
I can confirm that the load ordering bug is fixed in Maya 2027.3, with Bifrost unloaded.

PR #509, opened 2026-09-03
HYDRA-2520 - Bind the viewport camera prim for non-Storm render delegates

vlasovi:
## What this changes

A render delegate may take its camera data from a camera prim in the scene index rather
than from the render pass state. View and projection matrices are available either way,
so this is not about where the matrices come from; it is about the attributes that exist
only on a camera prim.

`MtohRenderOverride::Render()` cleared `renderParams.camera` for every delegate, so no
camera prim was bound for a delegate that wanted one. This binds the viewport camera's
scene index prim path for non-Storm delegates.

Animated cameras are also refreshed on time change. A delegate reading the camera prim is
driven by that prim, and Maya's per-node world-matrix callbacks do not fire reliably under
the Evaluation Manager during playback, so the prim would otherwise keep the transform it
was last dirtied with while the rest of the scene animates.

## Why this lands first

The camera prim is how camera attributes that the free camera does not carry reach a
delegate:

- motion blur needs shutter open and shutter close,
- depth of field needs focal distance and f-stop.

HVT's free camera passes matrices and nothing else, so those attributes have no route to a
delegate without a camera prim. Motion blur, depth of field, and render thread pipelining
all build on this, which makes it the first change in the series that needs to be in dev.

## Changes

- `MtohRenderOverride::Render()` binds `renderParams.camera` to the viewport camera's
  prim path for non-Storm delegates.
- `MayaHydraSceneIndex::GetCameraPrimPath()` maps an `MDagPath` to its camera prim path,
  returning an empty path unless an adapter is actually published for it, so a caller
  cannot bind a prim that does not exist.
- `MayaHydraSceneIndex::RefreshCamerasOnTimeChange()` re-evaluates camera transforms and
  dirties only the ones that moved.
- `MayaHydraDagAdapter::UpdateTransformIfChanged()` refreshes the cached world transform
  and reports whether it changed.

## Storm is unchanged

Storm continues to render through HVT's free camera with an empty camera path. HVT ignored
a supplied camera path until hydra-viewport-toolbox #173; once it began honouring it, the
path we supplied desynced the render from picking. The comment explaining that is kept, and
now says the empty path is specifically the Storm path.

## Scope

Maya perspective cameras. USD cameras and orthographic cameras keep the previous behaviour
and are marked TODO.

## Testing

Storm is covered by the existing viewport tests and is unaffected, since its camera path is
still empty.

No automated test accompanies this change. It only has an observable effect with a delegate
that reads camera data from a camera prim, and HYDRA-2122 records why no such delegate is
usable from the viewport test suite today: requesting the Arnold override silently falls
back to HdStorm, so a test written against it would pass without exercising the binding.


lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1342,20 +1342,44 @@ MStatus MtohRenderOverride::Render(
...
     // When that PR came into effect, our custom camera path was now used, but 
     // it turns out it provided the wrong values, and desynced the render from
-    // the picking. So we now mark the camera path empty to intentionally use
+    // the picking. So for Storm we mark the camera path empty to intentionally use
     // the previous behaviour.
+    //
+    // Other delegates may take camera data from a camera prim rather than the render pass
+    // state. The matrices are available either way, but the shutter interval and depth of
+    // field attributes travel only on the prim, so bind the viewport camera for those.
+    const bool needsCameraPrim = !_isUsingHdSt;
```
lilike-adsk: @debloip-adsk we set the custom camera instead of free camera back for other render delegates, any problem at HVT side?

lib/mayaHydra/mayaPlugin/renderOverride.cpp
```diff
@@ -1342,20 +1342,44 @@ MStatus MtohRenderOverride::Render(
...
+        }
+    }
+
     for (int i = 0; i < numFramePasses; ++i) {
         const hvt::FramePassPtr& currentPass = _GetFramePass(i);
         if (!currentPass) {
             continue;
         }
-        currentPass->params().renderParams.camera = SdfPath();
+        currentPass->params().renderParams.camera = cameraPath;
```
vlasovi, line 1368: @lanierd-adsk Setting the camera path for delegates other than hdStorm here affects PRMan unitest testLightingRenderDelegates, which produces an image with a slightly dufferent camera. So what should we do about that? Update the base image or somehow change the logic to set the camera here only for non-production rendering?

ppt-adsk: Production rendering is controlled by batchRenderer.cpp.  renderOverride.cpp is strictly for viewport rendering.

vlasovi: Yeah, theoretically PRMan should not be affected, but somehow it fails now in testLightingRenderDelegates

lanierd-adsk: In the PRMan testLightingRenderDelegates, the camera is hardcoded in the file, we assign it before the render, so you should change this. This is because, sometimes on different OS/ maya versions, it was jumping to another camera.

lanierd-adsk: Look for :
 def _setupSnapshotViewport(self):
     """Pin the snapshot camera and display flags used when baselines were captured.

     The test scene stores camera1 on the Persp panel in its embedded UI config,
     but that config is not always restored in batch/standalone Maya (depends on
     Maya version and $gUseScenePanelConfig). Without an explicit lookThru, the
     active editor can keep a default persp view and snapshot comparisons fail
     with a large framing mismatch rather than lighting noise.
     """
     panel = cmds.playblast(activeEditor=True)
     cmds.lookThru(panel, "camera1")
     cmds.modelEditor(
         panel,
         edit=True,
         displayLights="all",
         displayTextures=True,
         shadows=True,
         grid=False,
     )
     cmds.refresh(force=True)

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lilike-adsk: @vlasovi Could you please validate if the wireframe highlighting still working well under HdArnold, if yes, we should be fine.

lilike-adsk (dismissed):

lib/mayaHydra/mayaPlugin/renderOverride.cpp
vlasovi: Yes, the wireframe highlighting and other secondary graphics works well with HdArnold, so we can and should use the same camera prim for the main pass and the secondary pass with HdStorm.

However, the old and the new approaches may show discrepancies in camera FOV and some other parameters so I'm adding an env var to configure which one to use, that defaults to the old way. I think in the future, after comprehensive testing, we can enable the new way by default but then we will need to update the autotest images.

vlasovi:
@lilike-adsk I figured out the issue with the camera FOV not matching and pushed the fix. I also enabled the camera prim by default, so now we have it working for all the render delegates including HdStorm. All the autotests passed and also I didn't find any more issues in my local testing so we should be fine.

lilike-adsk:
> @lilike-adsk I figured out the issue with the camera FOV not matching and pushed the fix. I also enabled the camera prim by default, so now we have it working for all the render delegates including HdStorm. All the autotests passed and also I didn't find any more issues in my local testing so we should be fine.

Nice work!

lilike-adsk (approved):

PR #511, opened 2026-09-03
HYDRA-2520 - Publish multi-sample transform and primvar data for motion blur

vlasovi:
## What this changes

Motion blur needs more than one sample of a prim across the shutter interval. The scene
index published a single retained matrix for `xform`, and the primvar value source
returned `false` from `GetContributingSampleTimesForInterval()`, so a delegate had exactly
one sample to work with and produced no motion. This publishes multiple keys for both.

Transforms are served by a matrix data source installed on the `xform` schema when motion
samples are enabled:

- DAG-backed prims (meshes, lights, cameras) re-evaluate under `MDGContextGuard`, so their
  keys come from `MayaHydraDagAdapter::SampleTransform()`.
- Render items are not DAG adapters, and `MRenderItem::getMatrix()` is a Viewport 2.0
  snapshot of the current frame that does not re-evaluate under a context guard.
  `UpdateTransform()` therefore samples the source DAG node's world transform at shutter
  open and close and applies those deltas to the item's matrix
  (`worldKey = worldCentre * centreDag⁻¹ * keyDag`), publishing the two keys as a span when
  they differ.

Primvars are sampled through `MayaHydraShapeAdapter::SamplePrimvar()`, which covers
deforming geometry such as animated mesh points. Three keys (open, centre, close) let a
consumer interpolate a curved motion path rather than a straight line between endpoints.

Shutter offset 0 always returns the live transform rather than a sample, so an interactive
edit such as tumbling an animated camera is not overridden by the animated value at that
time.

## What this replaces

The existing render item motion path had no effect. It converted a matrix that had already
been captured before the context guard was installed:

    MDGContextGuard guard(MAnimControl::currentTime() + 1.0);
    _transform[1] = GetGfMatrixFromMaya(matrix);   // matrix was fetched above the guard

so the second key always equalled the first, and nothing read `_transform[1]`. The offset
was also a fixed one frame rather than the shutter interval.

## Relationship to the camera prim binding

Camera motion blur also needs the camera prim binding from the previous PR in this series.
The shutter interval itself is already published on the camera prim (`shutterOpen` and
`shutterClose` in the camera adapter and data source); what was missing is the
multi-sampled camera transform added here, plus a delegate that reads the camera prim at
all. Geometry motion blur does not depend on that PR.

Must be merged after: https://git.autodesk.com/maya3d/maya-hydra/pull/179

## Changes

- `mayaHydraDataSource.cpp`: `MayaHydraTransformMatrixDataSource` publishes matrix keys
  across the shutter, dispatching on adapter type; `xform` installs it only when motion
  samples are enabled and keeps the retained single-sample source otherwise.
- `mayaHydraPrimvarDataSource`: samples primvars across the shutter on first use and
  caches the result.
- `renderItemAdapter`: captures shutter open and close keys; `_transform` grows to three
  entries (centre, close, open) and is identity-initialised, since `UpdateTransform()`
  only writes the keys when `getMatrix()` succeeds and comparing uninitialised keys would
  publish a bogus motion span.

## Gating

Driven by the existing motion sample parameters (`motionSampleStart`/`motionSampleEnd` via
`motionSamplesEnabled()`); this adds no new switch. With motion samples off, `xform` keeps
its retained single-sample source and the primvar source returns the live value without
doing any sampling work, so scenes that do not ask for motion blur are unaffected.

## Testing

Verified interactively with the Flash render delegate: transform motion blur on animated
geometry and on render items, deforming mesh motion blur, and camera motion blur with the
camera prim bound.

No automated test accompanies this change, for the same reason as the camera prim binding
PR: the published samples only have an observable effect in a delegate that consumes them
to render motion blur, and no such delegate is usable from the viewport test suite today
(HYDRA-2122). Storm does not consume multiple transform samples, and with motion samples
off the published data is byte-identical to before, so the existing viewport tests are
unaffected either way.


lilike-adsk:
Looks good, btw, there's still a way to unit test this through validating the Hydra Scene directly, please refer to test\lib\mayaUsd\render\mayaToHydra\cpp\testMeshPrimvars.cpp if we want to add now.

lilike-adsk (approved):

vlasovi:
It may be better to add an image comparison autotest once we have the motion blur fully working. Let's get it in just like that for now.

PR #512, opened 2026-09-04
HYDRA-2506 : Fix frameNbResolution_Arnold  on MacOS

lanierd-adsk:
Summary
Fixes HYDRA-2506 on macOS Maya 2027, where duplicate OpenUSD libraries caused Tf registry collisions and prevented Arnold renders from producing images.

Changes
Use MayaUSD’s bundled OpenUSD runtime on macOS and retarget imported USD libraries to it.
Add a CMake regression test for USD location selection and library retargeting.
Remove usd_proc.dylib and its plugInfo.json registrations from extracted MtoA test cuts because it embeds another OpenUSD copy.
Disable the duplicate custom-node translation sample now provided by arnold-usd.
Re-enable frameNbResolution_Arnold on macOS.
Give that test an absolute render directory because Hydra V2 does not resolve relative render-product paths.
Validation
Maya 2027 macOS preflight build #822

mayaHydra_resolveUsdLocation: passed.
frameNbResolution_Arnold: passed and produced the expected image.
Duplicate Tf registration errors were no longer present.

Known limitation
Removing usd_proc.dylib disables MtoA workflows that depend on it, including loading USD files through aiStandin and operations using AiSceneLoad or AiSceneWrite, such as Arnold-USD export.


ppt-adsk (approved):
Looks good, but do we see this as a workaround until the Arnold team provides a proper fix?  And therefore will we remove this in the future?

ppt-adsk:
Great work, nice find!

lanierd-adsk:
> Looks good, but do we see this as a workaround until the Arnold team provides a proper fix? And therefore will we remove this in the future?

I am not sure we can remove this workaround as the usdproc.dylib is required by the Arnold team but not in our tests cases. So it's probably safer to keep this change but I called it a workaround.

PR #516, opened 2026-09-08
HYDRA-2520 - Add support for Depth Of Field

vlasovi:
## What this changes

Hydra treats `fStop == 0` as "depth of field off". The camera adapter decided that from
Maya's own `depthOfField` attribute alone, so a camera with Arnold depth of field enabled
published `fStop = 0` and rendered with no depth of field: enabling DOF through the Arnold
camera UI sets `aiEnableDOF`, not Maya's `depthOfField`.

Depth of field is now considered enabled when either attribute is set, and `focusDistance`
prefers Arnold's `aiFocusDistance` when that is set to a positive value, falling back to
Maya's `focusDistance` otherwise.

## Changes

- `_IsDofEnabledOnCamera()` accepts either Maya's `depthOfField` or Arnold's `aiEnableDOF`.
- `_ReadArnoldFocusDistanceIfSet()` returns 0 for "not set", which is what leaves Maya's
  `focusDistance` in play.
- `aiEnableDOF` and `aiFocusDistance` join `kCameraParamAttributeNames`, so editing either
  one dirties the camera's params and primvars.
- The world-matrix callback now dirties camera params and primvars alongside the transform.
  Tumbling moves the camera without touching the `fStop` or `focusDistance` plugs, which
  would otherwise leave the depth of field state stale.
- `depthOfField` is advertised in `MayaHydraCameraDataSource::GetNames()`. It is
  Maya-specific rather than part of `HdCameraSchema`, but advertising it makes it
  discoverable on the prim.

## Prerequisite

Depends on the camera prim binding [PR (#179)]. Depth of field attributes only reach a
delegate that reads the camera prim.

## Ordering against the exposure PR

This and the physical camera exposure PR both extend `kCameraParamAttributeNames` in
`cameraAdapter.cpp`. Taking this one first keeps that conflict to a one-line resolution.

## Testing

Adds `CameraPrimvars.DepthOfFieldSchemaFieldsRoundTrip`, registered through
`testCameraPrimvars.py`: with `depthOfField` enabled, `focusDistance` and `fStop`
round-trip to the camera prim's schema fields, and turning `depthOfField` off makes `fStop`
report 0. It asserts on published scene index data rather than rendered pixels, so it runs
under Storm in the existing suite.


test/lib/mayaUsd/render/mayaToHydra/cpp/testCameraPrimvars.cpp
```diff
@@ -188,3 +189,107 @@ TEST(CameraPrimvars, ParamAttributesMatchGetLogic)
                "kCameraParamAttributeNames in cameraAdapter.cpp.";
     }
 }
+
+namespace {
+
+HdSceneIndexBaseRefPtr FindSceneIndexWithCamera(
```
lilike-adsk, line 195: As mentioned in slack, without enabling the camera prim, the test will fail as those attributes are set on camera prim instead of the free camera generated by Hydra on-the-fly.

lilike-adsk (approved):

PR #518, opened 2026-09-09
Simple percentage batch render progress message.

ppt-adsk:


lanierd-adsk:
I am fine with what you did for multiple frames. 
But I think we should also have completion per frame when you have a single frame.
There is a function to get the render stats : VtDictionary HdRenderDelegate::GetRenderStats()
hdPRMan and hdArnold support it and return in the key "percentDone" the completion percentage.
They both also support in common : "totalClockTime" and "renderProgressAnnotation" which I think is worth checking.

hdStorm supports it as well but returns only the GPU memory perf stats, no completion percentage which is understandable for a rasterizer.
If you don't add this per frame stats to this PR, please log it in JIRA.

lanierd-adsk (approved):
As I said it's up to you if you want to add per frame stats in this PR or in another.

ppt-adsk:
> As I said it's up to you if you want to add per frame stats in this PR or in another.

I think we should merge this first, as it is a step forward, and enter a ticket and prioritize single-frame intra-frame progress later.

PR #520, opened 2026-09-10
HYDRA-2520 - Add photographic exposure controls to the camera and pass them to hydra as linearExposureScale

vlasovi:
## What this changes

A Maya camera has no notion of photographic exposure, so a delegate rendering in physical
units has nothing to drive `HdCameraSchema::linearExposureScale` with. This adds opt-in
physical exposure attributes to the Maya camera shape and publishes the resulting linear
scale on the camera prim.

The scale follows the standard photographic relation:

    EV100        = log2(N² / t) − log2(S / 100)
    linear scale = 2^(−EV100 + evComp)

with N the f-stop, t the shutter time in seconds, and S the ISO. It is 1.0, meaning no
change, whenever physical exposure is switched off or any input is non-positive, so scenes
that do not opt in render exactly as before.

## The attributes

Eight attributes are added to the camera shape: an enable toggle, f-stop, shutter speed and
ISO (each a full-stop drop-down with a companion Custom float), and an EV compensation
float. Names, value tables, UI labels and the formula all live in a single header so every
scene index path translating a Maya camera reports the same exposure.

A preset index outside its value table means Custom, in which case the companion float is
read instead. Each enum's field order must match its value table, with Custom at index
== table size.

## Where the attributes come from

`EnsureExposureAttributes()` authors them from `MayaHydraCameraAdapter::Populate()` rather
than from a data source pull, so the DG is never modified mid-evaluation. It is idempotent.

Referenced and locked nodes cannot take dynamic attributes and are skipped, so a camera in
that state reports the pass-through scale of 1.0 and cannot use physical exposure.

## Why the attributes live on the camera shape

Unlike depth of field, where both Maya and Arnold already provide attributes to read,
a Maya camera carries nothing describing exposure, so the attributes have to be authored
somewhere. Putting them on the camera shape keeps exposure per-camera, keyable and
animatable, and saved with the scene alongside the camera it belongs to.
The cost is that `Populate()` authors them on every camera it translates, including the
default `persp`, `top`, `front` and `side` cameras, and because they are storable they
persist in the scene file and appear in the Channel Box even for users who never enable
physical exposure. The alternatives trade that away for something else: a separate
settings node gives up per-camera control and animation, non-storable attributes give up
persistence, and creating them only on opt-in needs a UI or command entry point to author
them from. If reviewers would rather pay one of those costs instead, the authoring is
confined to `EnsureExposureAttributes()` and the reader already falls back to a
pass-through scale of 1.0 when the attributes are absent, so switching strategies is a
contained change.

## Changes

- `mayaHydraCameraExposure.h`: attribute names, value tables, UI labels, the EV formula,
  and the idempotent attribute authoring helpers.
- `cameraAdapter.cpp`: `Populate()` ensures the attributes exist; `GetCameraParamValue()`
  answers `linearExposureScale` from the shape's own attributes, ahead of the `MFnCamera`
  lookup; the eight attributes join `kCameraParamAttributeNames` so editing any of them
  re-pulls the scale.
- `mayaHydraCameraDataSource.cpp`: advertises `linearExposureScale`, which is served
  through the existing `HdCameraSchema` float branch.

## Prerequisites and ordering

Depends on the camera prim binding PR (#NNN); `linearExposureScale` only reaches a delegate
that reads the camera prim.

Should land after the depth of field PR (#NNN). Both extend `kCameraParamAttributeNames` in
`cameraAdapter.cpp`, and taking depth of field first keeps that conflict to a one-line
resolution.


lib/mayaHydra/hydraExtensions/adapters/cameraAdapter.cpp
```diff
@@ -142,6 +149,9 @@ void MayaHydraCameraAdapter::Populate()
     if (_isPopulated) {
         return;
     }
+    // Authored at setup rather than during the data-source pull, so the DG is never
+    // modified mid-evaluation.
+    MayaHydraCameraExposure::EnsureExposureAttributes(GetDagPath().node());
```
lilike-adsk, line 154: Adding attribute to a Maya camera node? strange, we shouldn't change Maya node, should this be done in Maya?

vlasovi: Good question. Other plugins do it. We can do it too if needed. This will also depend on where we implement the UI to control those. If we implement it in MayaHydra then creation of attributes should also stay here. We can discuss more offline.

ppt-adsk: If these attributes are for Hydra render delegate consumption only, then I also feel they should stay in MayaHydra.  In that case we should consider converting them into extension attributes, which I feel are more consistent with the intent here, that is, having plugin-related attributes being automatically added to a node.  In such a case I believe the "ensure" part would not be needed, as these attributes would be present on camera nodes by virtue of the MayaHydra plugin being loaded.  We can discuss and check all this.

lilike-adsk: Our MayaHydra is a translator not renderer, still feel a bit weird to add rendering attributes at our side. We need to be careful as the new attribute will be saved in the scene, the database maintenance burden is not trivial based on my 3dsMax experience.

I don't know the history of dealing camera exposure in Maya, surprised it's not a general Maya camera attribute as most renderers can support it, maybe it's time to add it as built-in attribute at Maya side? A similar case like we need a built-in DomeLight data model in Maya.

vlasovi:
We decided to proceed differently with those changes. Closing this PR.

PR #523, opened 2026-09-14
HYDRA-2445 : Integrate support for UsdDefaultRenderSettings.currentRe…

lanierd-adsk:
Summary
HYDRA-2445: Support UsdDefaultRenderDescription.currentRenderer as the default attribute to get the current hydra Renderer from the scene.

hydraRender no longer defaults to Storm when -renderer is omitted. It now resolves the renderer from the explicit -renderer/-r flag, or from the USD UsdDefaultRenderDescription.currentRenderer attribute. If neither is set, it fails with a clear error. 

Test plan
- ctest -C RelWithDebInfo -R testCurrentRendererSelection (8 mayabatch tests, including missing/empty renderer cases)
- New Render.exe tests for renderCurrentFrame_UsdCurrentRendererStorm* (explicit flag, override, and no-flag resolution)


lanierd-adsk:
The maya build needs to be updated in ecg maya hydra as this PR requires a change in maya.

lanierd-adsk:
@copilot resolve the merge conflicts in this pull request

cmake/test.cmake
```diff
@@ -1051,7 +1069,7 @@ function(mayaHydra_add_cmd_line_render_test SCENE_FILE_LABELED)
     # -----------------
 
     cmake_parse_arguments(ARG
-        "COPY_SCENE"             # Boolean options.
+        "COPY_SCENE;NO_RENDERER_FLAG"             # Boolean options.
```
ppt-adsk: Maybe I'm missing something, but wouldn't it be simpler to not have this CMake parameter and just omit -renderer if no RENDERER parameter is passed?

cmake/test.cmake
```diff
@@ -1154,7 +1172,11 @@ function(mayaHydra_add_cmd_line_render_test SCENE_FILE_LABELED)
     # The command needs to be the name of an executable, without any
     # arguments, as CMake calls an executable with that string unparsed.
 
-    set(RENDER_ARGS "\"${RENDER_EXECUTABLE}\" -renderer \"${RENDERER}\" ${ARG_RENDERER_ARGS} \"${SCENE_PATH}\"")
+    if(ARG_NO_RENDERER_FLAG)
```
ppt-adsk: Here you can just fold in the -renderer with its argument into the ARG_RENDERER_ARGS, or similar.

cmake/test.cmake
```diff
@@ -1173,7 +1195,11 @@ function(mayaHydra_add_cmd_line_render_test SCENE_FILE_LABELED)
     if (WIN32)
         set(CMD PowerShell)
 		# Windows (PowerShell)
-		set(RENDER_ARGS "& \"${RENDER_EXECUTABLE}\" -renderer \"${RENDERER}\" ${ARG_RENDERER_ARGS} \"${SCENE_PATH}\"")
+		if(ARG_NO_RENDERER_FLAG)
```
ppt-adsk: Same as above.

doc/batchRenderSettings.md
```diff
@@ -88,6 +88,59 @@ So `-reg 0 W-1 0 H-1` covers the full image, `-reg p p p p` selects exactly
...
+
+### Version-contract requirement
+
+The `UsdDefaultRenderDescription` node and its `currentRenderer` attribute
+are looked up unconditionally, without feature-detecting whether they
+exist.  In every supported production configuration, MayaUSD's
+`UsdDefaultRenderDescription` singleton (and its `currentRenderer`
+attribute) is always present, so a missing node or attribute is treated as
+a coding error, not as "the feature is disabled".  Either that case or an
+unauthored (empty) attribute value is reported as "no renderer specified".
```
ppt-adsk: This is quite wordy for something that is quite simple.  Furthermore, the absence of a UsdDefaultRenderDescription node is a configuration error, not a coding error.  I would just say

The `UsdDefaultRenderDescription` node and its `currentRenderer` attribute are guaranteed to be present in every supported configuration.

doc/batchRenderSettings.md
```diff
@@ -88,6 +88,59 @@ So `-reg 0 W-1 0 H-1` covers the full image, `-reg p p p p` selects exactly
...
+  renderer specified..." before any render delegate is created.
+- **An unrecognized renderer name** (from either the flag or the
+  attribute): resolution itself does not validate the name against Maya's
+  registered renderers.  Instead, the name is looked up against the
+  registered Hydra render delegates when the batch renderer initializes;
+  if no matching render delegate plugin is found, it fails with "unknown
+  or unregistered renderer...".  Either way, the failure surfaces as a
+  Python `RuntimeError` from `cmds.hydraRender()` and does not crash or
+  otherwise disturb the Maya session — a subsequent `hydraRender` call
+  with a valid renderer succeeds normally.
```
ppt-adsk: Very wordy for little reason, in my opinion.  I would keep this to a single sentence and roll it into the previous section's paragraph:

There is no silent fallback to a default renderer.

lib/flowViewport/API/interfacesImp/fvpDataProducerSceneIndexInterfaceImp.cpp
ppt-adsk, line 1: Why did this start to appear in this build?  What does the current renderer have to do with scene index tree destruction?

lib/flowViewport/API/renderViewData/fvpRenderViewDataManager.cpp
ppt-adsk, line 1: I don't understand what changed to make this needed.

lib/mayaHydra/mayaPlugin/batchRendering/batchRenderer.cpp
ppt-adsk, line 1: A lot of complicated changes I don't quite understand.  If we can't identify the renderer, why do we even try to build the scene index tree?

lib/mayaHydra/mayaPlugin/batchRendering/hydraRenderCmd.cpp
```diff
@@ -161,9 +161,44 @@ HydraRenderCmd::~HydraRenderCmd()
 
 bool HydraRenderCmd::parseDatabase(const MArgDatabase& db)
 {
+    if (db.isFlagSet(_renderer)) {
```
ppt-adsk, line 164: Not sure why the move was done, no harm I suppose.

lib/mayaHydra/mayaPlugin/batchRendering/hydraRenderCmd.h
```diff
@@ -51,9 +53,28 @@ class HydraRenderCmd : public MPxCommand
...
     bool hydraRenderFromHydraV2RenderSettings();
 
+    /*! \brief Resolve the renderer to use for this batch render.
+     *
+     *  This function does not validate that the resolved renderer name
+     *  corresponds to a registered/known Hydra render delegate: that
+     *  validation is performed downstream by the renderer-plugin lookup
+     *  in BatchRenderer::_InitHydraResources(), which reports an error
+     *  and fails initialization if the name does not resolve to a
+     *  registered render delegate.
```
ppt-adsk, line 63: Why this architectural choice?

lib/mayaHydra/mayaPlugin/renderOverride.cpp
ppt-adsk, line 1: Why do we need this change?

lib/mayaHydra/mayaPlugin/renderSettingsUtils.cpp
```diff
@@ -80,18 +82,35 @@ bool IsPrmanRenderSettingsDriveRenderPassEnabled(const TfToken& rendererName)
     return TfGetenvBool("HD_PRMAN_RENDER_SETTINGS_DRIVE_RENDER_PASS", false);
 }
 
-UsdPrim _ReadActiveRenderDescriptionPrim(Ufe::Path& outPath)
+MPlug _GetUsdDefaultRenderDescriptionPlug(const char* attrName)
```
ppt-adsk, line 83: At this point may as well add a function to mayaUtils.h:
```
MStatus GetPlug(const MString& nodeName, const MString& attrName, MPlug& plug);
```

lib/mayaHydra/mayaPlugin/renderSettingsUtils.cpp
```diff
@@ -201,6 +220,87 @@ RenderSettingsType ReadRenderSettingsTypeFromRenderDelegate(const TfToken& rende
...
+    // not a coding error, so no TF_VERIFY/TF_WARN for this case.
+    MString rendererStr = plug.asString();
+    if (rendererStr.length() == 0) {
+        return TfToken();
+    }
+
+    return TfToken(rendererStr.asChar());
+}
+
+Ufe::SceneItemList GetAllMayaUsdProxyShapes()
```
ppt-adsk: We should not traverse the scene to find render settings.  I recently removed this code from MayaHydra.

lib/mayaHydra/mayaPlugin/renderSettingsUtils.cpp
```diff
@@ -201,6 +220,87 @@ RenderSettingsType ReadRenderSettingsTypeFromRenderDelegate(const TfToken& rende
...
+    std::copy(
+        gatewayItems.begin(),
+        gatewayItems.end(),
+        std::back_inserter(proxyShapes)
+    );
+    
+    return proxyShapes;
+}
+
+bool FindUsdRenderSettingsOnStage(
```
ppt-adsk: We should not do this.  This is unpredictable and potentially costly.

test/lib/cmdLineRender/CMakeLists.txt
```diff
@@ -583,6 +657,19 @@ foreach(script ${MAYABATCH_TEST_SCRIPT_FILES})
...
+    # file's Arnold cases are the first mayabatch tests converted; see
+    # testCurrentRendererSelection.py's use of an absolute render product path
+    # to work around Hydra V2 not resolving a relative USD productName.
+    if(test_filename STREQUAL "testCurrentRendererSelection.py")
+        list(APPEND test_env
+            "MAYA_HYDRA_HD_ARNOLD_HYDRA_V2_RENDER_SETTINGS=1"
+            "USDIMAGINGGL_ENGINE_ENABLE_SCENE_INDEX=1"
+            "TF_DEBUG=MAYAHYDRAPLUGIN_BATCHRENDER_RENDER_SETTINGS")
+    endif()
+
```
ppt-adsk: This should be removed.

test/lib/cmdLineRender/CMakeLists.txt
```diff
@@ -335,6 +335,79 @@ foreach(scene ${ARNOLD_MAYABATCH_RENDER_SCENES})
...
+# UsdDefaultRenderDescription.currentRenderer to "HdStormRendererPlugin"
+# directly in the scene file (rather than setting it at runtime, as the
+# Python test does), so pure Render.exe/CTest invocations can exercise it.
+# It also sets defaultRenderGlobals.preMel to "loadPlugin -quiet mayaHydra;"
+# so that mayaHydra -- and therefore its Hydra renderers -- is registered
+# with Maya's `renderer` command even when Render.exe is invoked without an
+# explicit -renderer flag (see the third test below): without an explicit
+# -renderer flag, no renderDesc/*.xml <melheader> runs, so nothing else would
+# load mayaHydra before Maya's currentRenderer()/mayaBatchRenderProcedure MEL
+# logic needs it registered.
```
ppt-adsk: I think we should find a different way to do this.  Using defaultRenderGlobals to load mayaHydra feels like a breach of separation of concerns: we explicitly moved away from defaultRenderGlobals.currentRenderer for this reason.  Now we would be bringing defaultRenderGlobals back into the requirements.

If UsdDefaultRenderDescription exists, then mayaUsdPlugin must be loaded to interpret it.  If the UsdDefaultRenderDescription can only be used with Hydra renderers, then mayaHydra must be loaded.  This needs a conversation at product management scope.

ppt-adsk (changes requested):
I think some things need discussion, namely:
- Changes to coverage build because of UsdDefaultRenderDescription.currentRenderer
- Requirement to use defaultRenderGlobals.preMel

cmake/test.cmake
lanierd-adsk: Good point — NO_RENDERER_FLAG is redundant with the implicit default. The function currently defaults RENDERER to HdStormRendererPlugin and always passes -renderer, so I needed a separate flag to exercise the no-flag path. I'll drop the default and only append -renderer when RENDERER is provided; the no-flag test will just omit RENDERER. The STORM_RENDER_SCENES loop is the only existing caller that relied on the implicit default, so I'll add RENDERER HdStormRendererPlugin there explicitly.

cmake/test.cmake
lanierd-adsk: Agreed — I'll prepend -renderer to ARG_RENDERER_ARGS when RENDERER is set, so both the POSIX and PowerShell RENDER_ARGS lines share a single form with no duplicated if/else.

cmake/test.cmake
lanierd-adsk: Same fix as above — -renderer is now folded into ARG_RENDERER_ARGS before the shell split, so the PowerShell path is also a single RENDER_ARGS line with no duplicated if/else.

doc/batchRenderSettings.md
lanierd-adsk: Updated, thanks

doc/batchRenderSettings.md
lanierd-adsk: Right, updated

lib/flowViewport/API/interfacesImp/fvpDataProducerSceneIndexInterfaceImp.cpp
lanierd-adsk: Nothing to do with current renderer directly — sorry this looked coupled.

This showed up because the coverage-fix commit re-enabled BatchRenderer::_ClearHydraResources() → RemoveRenderViewData() under CODE_COVERAGE_WORKAROUND (it used to be skipped). That tears down the dummy batch viewport, which runs RenderViewData::~RenderViewData() → removeAllDataProducerSceneIndicesFromView() → RemoveSceneIndex(). On Windows clang coverage builds those scene-index destructors crash (same long-standing CODE_COVERAGE_WORKAROUND issue elsewhere in flowViewport/batchRenderer).

The new currentRenderer batch tests exercise that teardown more often in coverage CI, which is why the crash surfaced here. The fix is to leak the ref instead of calling RemoveSceneIndex() under the workaround flag, matching the pattern already used in fvpRenderViewDataManager.cpp.

lib/flowViewport/API/renderViewData/fvpRenderViewDataManager.cpp
lanierd-adsk: Nothing in this file changed because of currentRenderer resolution — sorry this looks coupled.

What changed on this branch is that coverage CI started crashing during batch teardown. BatchRenderer::_ClearHydraResources() calls RemoveRenderViewData(kBatchRenderDummyPanelName). On dev, that path had no CODE_COVERAGE_WORKAROUND handling; only RemoveAllRenderViewData() already used the leakViewData() pattern.

We first worked around the crash by skipping RemoveRenderViewData() entirely under coverage (0767faf1b6), but that left batch teardown incomplete. In f61e5e8da8 we re-enabled it and added the same leak pattern to RemoveRenderViewData(). The new currentRenderer batch tests hit that path more often in coverage CI, which is why the crash surfaced here.

The #include <flowViewport/fvpUtils.h> was added in 47dfe62e97 because RemoveRenderViewData() now calls Fvp::leakSceneIndex() after RemoveSceneIndex(), and leaks a heap copy via leakViewData() so ~RenderViewData doesn't drop the last ref on scene indices whose destructors crash under Windows clang coverage builds. Same long-standing issue as elsewhere in flowViewport/batchRenderer — this file just needed the per-view variant of the workaround that RemoveAllRenderViewData() already had.

lib/mayaHydra/mayaPlugin/batchRendering/batchRenderer.cpp
lanierd-adsk: We don't build the scene index tree when renderer resolution fails.

There are two distinct cases:

1. Renderer not identified (missing/empty currentRenderer, empty -renderer)
hydraRenderCmd::GetRenderer() returns empty and doIt() fails immediately — no BatchRenderer is created, no Hydra init, no scene index tree.

2. Renderer name present but unusable (e.g. NotARealRendererPlugin)
A BatchRenderer is constructed with that token, but _InitHydraResources() bails out at GetRendererPlugin() / CreateRenderDelegate() / HdRenderIndex::New() — all before _mayaHydraSceneIndex, the merging proxy, or the filtering chain are created. Initialize() returns false and the render never runs.

The if (!_mayaHydraSceneIndex) return; guard is for teardown of that second case. _initializationAttempted is set at the top of _InitHydraResources(), so when the BatchRenderer is destroyed after a failed init, _ClearHydraResources() still runs. Without the guard, the coverage path could call _mayaHydraSceneIndex->_Destroy() on a null pointer. The non-coverage path already had a null check; this makes both paths consistent.

That's what test_InvalidCurrentRendererDoesNotCrashSession exercises: fail cleanly on a bogus renderer, then succeed on a valid one in the same session — no crash during partial-init teardown.

The rest of the diff (leakSceneIndex, re-enabling RemoveRenderViewData under coverage, etc.) is the existing Windows clang CODE_COVERAGE_WORKAROUND pattern. The new renderer-selection error-path tests hit batch teardown more often in coverage CI, which is why those workarounds expanded here — not because we're building scene indices without a valid renderer.

lib/mayaHydra/mayaPlugin/batchRendering/hydraRenderCmd.cpp
lanierd-adsk: The move isn't cosmetic — it's needed for the new renderer-resolution behaviour.

Previously doIt() always defaulted to Storm and only overrode if -renderer was set. Now we need three distinct cases:

No -renderer flag → fall back to UsdDefaultRenderDescription.currentRenderer
-renderer set to a non-empty name → use that
-renderer set to empty → hard error (distinct from case 1)
Case 1 vs 3 can't be expressed with a single TfToken — an empty token means both "no flag" and "empty flag" unless we also record whether the flag was present. So parseDatabase() captures _rendererFlagSet + _rendererFromFlag, and GetRenderer() implements the precedence/fallback logic.

parseDatabase() was already called from doIt() (it was just a stub before), so this follows the usual Maya command pattern of parsing args separately from acting on them. Happy to leave a one-line comment there if that would help.

lib/mayaHydra/mayaPlugin/batchRendering/hydraRenderCmd.h
lanierd-adsk: Two different questions, two layers:

GetRenderer() — configuration resolution: did the caller specify a renderer name, and from which source? It handles flag vs currentRenderer precedence and the "missing/empty name" cases (-renderer "", unauthored currentRenderer). It returns a raw token, same as GetCurrentRenderer() in renderSettingsUtils.cpp ("No validation is performed on the returned value").

BatchRenderer::_InitHydraResources() — runtime validation: given a name, can Hydra actually instantiate it? HdRendererPluginRegistry::GetRendererPlugin() is the single source of truth here; duplicating that lookup in GetRenderer() would mean two registry queries and two places to keep error messaging in sync.

This isn't new architecture for this PR — on dev, plugin validation already lived only in _InitHydraResources(). What changed is resolution (no Storm default; read currentRenderer when -renderer is omitted), not where registry validation happens.

The two layers also produce distinct, testable errors:

"no renderer specified" (test_MissingCurrentRendererFailsWithoutRendererFlag)
"unknown or unregistered renderer "NotARealRendererPlugin"" (test_InvalidCurrentRendererDoesNotCrashSession)
BatchRenderer is constructed with just the resolved token; init failure stays in the Hydra setup path where delegate/render-index creation already lives.

lib/mayaHydra/mayaPlugin/renderOverride.cpp
lanierd-adsk: Not required for the currentRenderer behavior itself — this is a coverage-build stability fix.

_ClearMayaHydraSceneIndex() already had a CODE_COVERAGE_WORKAROUND path that called _mayaHydraSceneIndex->_Destroy() and then _mayaHydraSceneIndex.Reset(). _Destroy() runs our Maya-side teardown (callbacks/adapters), but it does not stop the base HdRetainedSceneIndex C++ destructor from running when the RefPtr is reset. On Windows clang coverage builds that dtor crashes, which is why we use Fvp::leakSceneIndex() elsewhere.

While getting the new batch-renderer coverage tests green, the same incomplete pattern showed up in the new BatchRenderer::_ClearMayaHydraSceneIndex() path. I fixed both in the same commit (84bff4fd6a): call _Destroy() for Maya cleanup, then Fvp::leakSceneIndex() so the .Reset() below cannot drop the last ref and invoke the crashy dtor. renderOverride.cpp was updated for the same reason — coverage runs still hit viewport teardown (_ClearHydraResources() → _ClearMayaHydraSceneIndex()), and the old workaround there was insufficient.

Non-coverage builds are unchanged (RemoveSceneIndex() path).

lib/mayaHydra/mayaPlugin/renderSettingsUtils.cpp
lanierd-adsk: ok

lib/mayaHydra/mayaPlugin/renderSettingsUtils.cpp
lanierd-adsk: You're right — this was reintroduced accidentally during the merge with dev, not intentionally for currentRenderer.

test/lib/cmdLineRender/CMakeLists.txt
lanierd-adsk: Agreed — defaultRenderGlobals.preMel is a test harness workaround, not something we want as product behavior. We moved renderer selection off defaultRenderGlobals.currentRenderer onto UsdDefaultRenderDescription.currentRenderer; baking plugin loading into defaultRenderGlobals.preMel reintroduces a dependency on that node for a different reason, and I don't think that's the right long-term contract.

Context for why it landed: the third CTest (…NoRendererFlagUsesUsdCurrentRenderer) invokes Render.exe with no -renderer flag, so no renderDesc/*.xml <melheader> runs and nothing else loads mayaHydra before Maya's currentRenderer() / mayaBatchRenderProcedure MEL logic needs the Hydra renderer registered. The Python tests cover the same precedence without this hack — MayaHydraBaseTestCase loads mayaHydra at harness setup, and test_UsdCurrentRendererStormIsHonored sets the attribute at runtime.

I agree the production rule needs PM input:

if UsdDefaultRenderDescription is present → mayaUsdPlugin must be loaded to interpret it
if currentRenderer names a Hydra delegate → mayaHydra (and any renderer-specific plugin) must be loaded before resolution/render
For this PR I'll remove the preMel from the scene and drop (or quarantine) the no-flag Render.exe CTest, keeping the explicit-flag and override CTests plus the Python coverage. Happy to open a follow-up / PM thread for the Maya-side plugin-loading contract once we align on expected behavior for Render.exe with no -renderer flag.

test/lib/cmdLineRender/CMakeLists.txt
lanierd-adsk: Removed. MAYA_HYDRA_HD_ARNOLD_HYDRA_V2_RENDER_SETTINGS and USDIMAGINGGL_ENGINE_ENABLE_SCENE_INDEX are already the defaults, and TF_DEBUG shouldn't be in CI. Arnold V2 coverage stays in testCurrentRendererSelection.py itself (absolute render product path workaround).

lib/mayaHydra/mayaPlugin/batchRendering/hydraRenderCmd.cpp
ppt-adsk: But this turns two strictly local variables in doIt() into two new class data members, _rendererFlagSet and _rendererFromFlag, for no strong reason, since doIt() is the only method which will act on them.  This is a breach of encapsulation.  Not very convinced, I don't think adding data members as a means of passing data around inside an object for a single use is a good idea, but willing to let this go.

lib/mayaHydra/mayaPlugin/batchRendering/hydraRenderCmd.h
ppt-adsk: But runtime validation does not have to be deferred to _InitHydraResources().  Using the `renderer -namesOfAvailableRenderers` and filtering on Hydra capability we can immediately test if Hydra can instantiate it.

lib/mayaHydra/mayaPlugin/batchRendering/batchRenderer.cpp
ppt-adsk: We can test for (2) much earlier.  See comment in hydraRenderCmd.h

lib/mayaHydra/mayaPlugin/batchRendering/hydraRenderCmd.h
lanierd-adsk: Fair point — plugin lookup doesn’t have to wait until _InitHydraResources(). What I wanted to avoid is duplicating validation through a different mechanism than the one that actually creates the delegate.

namesOfAvailableRenderers + isHydra tells us the name is registered with Maya as a Hydra renderer; it doesn’t tell us Hydra can instantiate it. _InitHydraResources() still has to run GetRendererPlugin() → CreateRenderDelegate() → HdRenderIndex::New(), and any of those can fail independently (we already see delegate creation fail without a GPU context in MtohInitializeRenderPlugins()).

If we want an earlier fail, I’d prefer a small shared C++ helper on HdRendererPluginRegistry::GetRendererPlugin() (same check _InitHydraResources() uses today) rather than round-tripping through the Maya renderer command. That keeps one source of truth and one error message, while still letting GetRenderer() surface “unknown renderer” before constructing BatchRenderer.

Happy to add that helper in this PR if you’d like the earlier error path; I’d still keep full init validation in _InitHydraResources() for delegate/index creation failures.

lib/mayaHydra/mayaPlugin/batchRendering/hydraRenderCmd.cpp
lanierd-adsk: The three-way -renderer behaviour needs flag-presence tracking, but it doesn't need to live on the object. _gpuEnabled is a member because hydraPreRender() reads it later; _rendererFlagSet / _rendererFromFlag are only used to bridge parseDatabase() → GetRenderer() inside doIt(), so locals (or GetRenderer(const MArgDatabase& db)) would be cleaner.

Happy to refactor to locals + GetRenderer(bool flagSet, const TfToken& fromFlag) in a follow-up if you'd like; otherwise I'll leave as-is per your note.

lib/flowViewport/API/renderViewData/fvpRenderViewDataManager.cpp
ppt-adsk: Whitespace change only, should be removed from branch.

lib/mayaHydra/mayaPlugin/batchRendering/batchRenderer.cpp
```diff
@@ -592,11 +612,7 @@ void BatchRenderer::_ClearHydraResources()
 
     if (_renderIndex != nullptr) {
         GetMayaHydraLibInterface().UnregisterTerminalSceneIndex(_renderIndex->GetTerminalSceneIndex());
-#ifndef CODE_COVERAGE_WORKAROUND
```
ppt-adsk: Do we still need this removal?

lib/mayaHydra/mayaPlugin/batchRendering/batchRenderer.cpp
```diff
@@ -399,12 +399,19 @@ void BatchRenderer::_ExecuteHydraBatchRenderFrame()
 
 void BatchRenderer::_ClearMayaHydraSceneIndex()
 {
+    // _InitHydraResources() sets _initializationAttempted before it can bail out on an
```
ppt-adsk: Is this still true?  We seem to be using _mayaHydraSceneIndex as some kind of partially-constructed flag.  Where does _InitHydraResources() partially construct something without cleaning it up?

lib/mayaHydra/mayaPlugin/batchRendering/hydraRenderCmd.h
ppt-adsk: I really feel that changes to this file introduce gratuitous complexity, and are not needed.  All the command needs to do is to obtain the renderer from the scene if it is not passed in through a flag, then validate the renderer name.

